PR #199, opened 2026-06-18
GH-C22: launch artifact observable handle for terminal-bench (narrow slice)

liangsalt:
Implements GH-C22 ("launch artifact observable handles") as a narrow slice that
  EXTENDS the existing terminal-bench launch/poll path instead of adding a
  standalone module.

  What it adds
  - `build_terminal_bench_launch_artifact_handle()`: a compact, board-recordable
    handle bundling launch identity (run/job basenames + pid-file basename),
    compact artifact refs, the exact `allowed_poll_command` + argv shape, and
    read-boundary flags. Public-safe (host paths reduced to basenames, no raw
    bodies) so a fresh/forgetful heartbeat can re-poll WITHOUT chat memory.
  - `observe_terminal_bench_launch_artifact_handle()`: re-probes the handle under a
    run_root and returns a compact, read-boundary-respecting observation. Reuses
    the existing detached pid-file probe (no duplicate probe logic).
  - `_pid_is_running()`: cross-platform, read-only pid liveness extracted from
    `_detached_process_state_from_pid_file`. POSIX outcomes unchanged
    byte-for-byte; adds a Windows branch (OpenProcess/GetExitCodeProcess) instead
    of a bare `os.kill(pid, 0)`, which on Windows routes through TerminateProcess.
  - `poll-worker-materialization-probe` additively carries `launch_artifact_handle`
    (new key; no existing key changed).

  Validation
  - `python examples/launch-artifact-handle-smoke.py` (the task's "Focused fake
    launch artifact smoke") — passes.
  - `python -m py_compile goal_harness/*.py goal_harness/benchmark_adapters/*.py`.
  - `goal-harness check` public/private boundary scan clean on changed files.

### DIFF
--- a/examples/launch-artifact-handle-smoke.py
+++ b/examples/launch-artifact-handle-smoke.py
@@ -1,0 +1,149 @@
+"""Focused fake launch artifact smoke for GH-C22 (no real worker is launched).
+
+Validates the ``terminal_bench_launch_artifact_handle_v0`` OBSERVE contract with
+a fake handle + a temp run-root holding a fake pid file:
+- the handle carries an ``allowed_poll_command`` + argv shape (a fresh/forgetful
+  heartbeat can re-poll WITHOUT chat memory)
+- a pid file pointing at our own pid    -> process_state == running
+- a pid file pointing at a dead pid     -> process_state == ended
+- a missing pid file                     -> process_state == unknown
+- read_boundary gates process_state / artifact_refs off when declared
+- the observation is compact-only        -> no raw bodies / logs / host paths
+- handle building is public-safe         -> host paths reduce to basenames
+
+Cross-platform: the underlying probe uses OpenProcess/GetExitCodeProcess on
+Windows (so observing our own pid does NOT terminate this process) and
+os.kill(pid, 0) on POSIX.
+
+Run: python examples/launch-artifact-handle-smoke.py
+"""
+
+import os
+import sys
+import tempfile
+from pathlib import Path
+
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+from goal_harness.benchmark_adapters.terminal_bench import (  # noqa: E402
+    TERMINAL_BENCH_LAUNCH_ARTIFACT_HANDLE_SCHEMA,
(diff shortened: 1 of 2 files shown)


huangruiteng:
Thanks for the clean GH-C22 slice. I checked it against current `main` after the recent benchmark-route changes:

- `git apply --check` on current `main`: passes.
- `python examples/launch-artifact-handle-smoke.py`: passes.
- `python3 -m py_compile goal_harness/*.py goal_harness/benchmark_adapters/*.py`: passes.

I do think the PR has real incremental value: the “observable launch handle” idea is exactly the right direction for long benchmark runs, because a later heartbeat should be able to re-observe a launched job from compact state rather than from chat memory.

That said, I would not merge this exact shape yet. Since this PR was opened, the benchmark path has moved toward a cloud-host default route plus a runner-agnostic launch/observe/ingest lifecycle. To make this fit the current direction, I’d suggest a small revision:

1. Move the handle contract toward a generic benchmark control-plane surface, e.g. `benchmark_launch_artifact_handle_v0` / `RunHandle` + `Observation` under `goal_harness/benchmark_core`, with Terminal-Bench as the first adapter implementation. Otherwise we will likely duplicate this again for SkillsBench and ALE.
2. Emit/persist the handle at launch time, not only from the poll path. If memory is lost after launch but before the first poll, the current implementation still has no durable handle to recover from.
3. Make the handle truly re-observable without hidden chat memory. Today the handle stores basenames and `allowed_poll_command`, but the poll command still needs private `run_root` / `jobs_dir` values from elsewhere. Either add a public-safe locator/handle id that the controller can resolve to private paths, or expose a CLI observe command that reads a controlled handle record and performs the private lookup without recording raw paths.
4. Keep the compact/read-boundary discipline, but centralize validation so the adapter does not merely trust arbitrary `read_boundary` flags from a handle payload.
5. The `_pid_is_running()` extraction looks like a useful independent fix and could be split out if you want a smaller first landing slice.

So my recommendation is: keep the concept, revise the shape to be generic and launch-persisted, then we can merge it as part of the shared benchmark lifecycle rather than a Terminal-Bench-only side contract.

PR #532, opened 2026-06-22
[codex] add Lark Kanban control-plane adapter

ZaynJarvis:
## Summary
- add a `loopx lark-kanban` CLI surface for schema, board creation planning/execution, sample task seeding, feasibility case seeding, and heartbeat worker processing
- model LoopX todos, claims, user gates, handoff, evidence, and run history in a Feishu/Lark Base Kanban board
- define the recommended lightweight operator Kanban card profile while preserving full record context in the table
- document the adapter contract and add deterministic smoke coverage for schema, command planning, claim, worker execution, writeback, UX cases, and CLI fixture behavior

## Validation
- `python3 -m py_compile loopx/*.py loopx/cli_commands/*.py`
- `python3 examples/lark-kanban-control-plane-smoke.py`
- `python3 -m loopx.cli --format json lark-kanban plan-create --base-name "LoopX Kanban POC" --base-token base_fixture --user-open-id ou_fixture`
- `python3 -m loopx.cli --format json lark-kanban seed-cases --base-token base_fixture --table-id tbl_fixture`
- `python3 -m loopx.cli check --scan-root .`

## Notes
- Direct Base-to-local-agent triggering is left as a user gate because this environment has no reachable cloud-to-edge daemon/callback bridge. The implemented path is heartbeat claim plus explicit worker execution allowlisting.
- Lark card field visibility is currently configured in the Lark UI; the available shortcut CLI only exposes Kanban cover configuration. The compact operator field set is documented and exported in schema output.
- The live Feishu Base prototype was validated separately and is intentionally not recorded in the public PR body.


huangruiteng:
Thanks for the prototype. I did another live product/usability pass after the latest `setup` workflow landed, using the target question: "Can a user take this PR version plus `lark-cli` and quickly create a LoopX Kanban board they can actually use?"

Current read: much closer. The PR is now useful enough to create a real private Lark Base/Table and seed a working scenario board, but I would still keep it draft until the setup/doctor fallback below is fixed.

What I verified locally on the PR head:

- `lark-cli 1.0.56` is available and the user identity is authenticated.
- `python3 -m loopx.cli --format json lark-kanban schema` returns the expected LoopX operator fields.
- `python3 -m loopx.cli --format json lark-kanban setup --base-name ... --table-name ... --as user --execute` successfully created a real Base plus the initial table/schema.
- I then used the created table to write 11 real scenario-anchor records and create views for priority review / human gates / evidence review.
- `lark-kanban use --base-token ... --table-id ... --view-id ... --as user` successfully wrote a local board config.

The remaining blocker I hit is important for self-serve setup:

`setup` creates the Base/table successfully, but then fails at:

```bash
lark-cli base +base-block-list --as user --base-token ... --type table
```

because the current user token is missing `base:block:read`. The error is good and explicit, but the workflow should not have to fail after creation in this case. `+base-create --table-name --fields` already returns the created table id, so `setup` can write the local config directly from that response and treat `base-block-list` as an optional discovery/enrichment step. Same for `doctor`: if config already has `base_token/table_id/view_id` and `base-get` + `view-list` pass, missing `base:block:read` should probably be a warning with an auth hint, not a hard failure.

Suggested next patch:

1. In `setup`, persist config immediately from the `base-create` return payload when it contains the new table id.
2. Use `base-block-list` only when attaching to an existing Base without a table id, or when enrichment is available.
3. In `doctor`, distinguish required checks from optional discovery checks:
   - required: `lark-cli`, auth, `base-get`, table/view access when ids are configured;
   - optional: `base-block-list` discovery, visible-fields refinement.
4. Keep printing the auth command for `base:block:read`, but do not discard the usable board state.

This PR is now a strong foundation for the LoopX management-surface story: it can turn LoopX goals/todos/gates/evidence into a Lark Kanban-style review surface. The next level of polish is making the happy path robust when a normal user has enough Base/table/view/record permissions but not every discovery scope.

huangruiteng:
Review Findings

[P1] `setup` 仍会在后置视图配置失败时丢掉已创建的可用 board 状态。

在 `loopx/lark_kanban.py:1407` 已经从 `base-create` 响应提取到了 `base_token/table_id`，但本地 config 要到 `loopx/lark_kanban.py:1562` 之后才写入；中间任何视图配置失败都会在 `loopx/lark_kanban.py:1557` 直接返回失败。

我用 fake live runner 复现了：`base-create` 返回 `base_live_fixture/tbl_live_fixture`，`view-list` 通过，最后 `+view-set-visible-fields` 因权限失败；结果 payload 里有 `base_token/table_id`，但 `config_exists=false`。这仍然没有满足 PR 评论里“不要 discard usable board state”的要求。建议在拿到 `base_token/table_id` 后立即保存最低可用 config，视图过滤/分组/visible-fields 作为 enrichment warning 或 partial setup 结果。

Open Questions

`base-block-list` 本身看起来已经被降级了：只有 `base-create` 没抽到 table id 时才会走 `loopx/lark_kanban.py:1425`。但后续 view 配置权限失败仍然会造成同类“外部资源已创建、本地状态没留下”的事故，所以我会把它视作同一个 blocker 的未完全修复。

Validation

已跑过：

```bash
python3 -m py_compile loopx/*.py loopx/cli_commands/*.py
python3 examples/lark-kanban-control-plane-smoke.py
python3 -m loopx.cli --format json lark-kanban schema
git diff --check origin/main...HEAD
```

现有 smoke 都过，但它没有覆盖这个 partial live setup 回滚/保全路径。

huangruiteng:
Review Findings after the latest update (`5b2d768`)

[P1] `setup` partial-success blocker is still present.

The latest update did not change `loopx/lark_kanban.py`, so the previous setup concern still reproduces: after `base-create` returns a usable `base_token/table_id`, the local config is still written only after the later view configuration steps. In the current PR head, `base-create` extracts ids around `loopx/lark_kanban.py:1407-1411`, but config persistence happens only around `loopx/lark_kanban.py:1562-1575`; failures from `view-create` / visible-fields setup still return before config is saved (`loopx/lark_kanban.py:1536-1537`, `1557-1560`).

I reran the fake live runner path: `base-create` returned `base_live_fixture/tbl_live_fixture`, `view-list` passed, then `+view-set-visible-fields` failed for permissions. The result still had `base_token/table_id` in the payload but `config_exists=false` / `stored_exists=false`. This still discards a usable external board state after creating the external resource.

[P1] The new active-state todo fallback is too narrow for connected read-only goals with passive latest runs.

The new `active_state_todo_attention_item()` path only runs when `goal_attention(goal)` returns no item (`loopx/status.py:7122-7126`). But for common connected goals, `adapter_status in CONNECTED_ADAPTER_STATUSES` (`connected`, `connected-read-only`, `pre-tick-runnable`) already creates a latest-run attention item (`loopx/status.py:247-250`, `7059-7065`). That means a connected-read-only goal with a passive latest run and an open user todo can still surface `waiting_on=codex`, `source=latest_run`, and stale `recommended_action` instead of projecting the active-state user gate.

Repro shape I used: `connected-read-only` adapter, latest run classification `state_refreshed`, plus one open user todo in active state. `quota should-run` correctly sees `interaction_contract.user_channel.action_required=true`, but `status` still shows the attention item as `waiting_on=codex` / `project_asset_gate=none` because the fallback never gets a chance to replace or enrich the existing latest-run item.

Suggested direction:

1. Preserve the minimum usable Lark Kanban config immediately after a usable `base_token/table_id` is known; treat view/filter/visible-field setup as enrichment warnings or partial setup repair items.
2. For active-state todo projection, let open user todos override or enrich connected/latest-run attention items, not only the “no attention item exists” case.
3. Add focused regression coverage for both shapes: partial live setup config preservation after enrichment failure, and connected-read-only passive latest run plus open user todo.

Validation rerun on the latest PR head:

```bash
python3 -m py_compile loopx/*.py loopx/cli_commands/*.py
python3 examples/lark-kanban-control-plane-smoke.py
python3 examples/active-state-todo-attention-fallback-smoke.py
python3 -m loopx.cli --format json lark-kanban schema
git diff --check origin/main...HEAD
```

All existing checks passed, but they do not cover the two blocker paths above.

huangruiteng:
Re-reviewed latest update `caba9c5`.

The two previous blockers look fixed:

- Lark Kanban setup now persists the minimum usable local config as soon as `base_token/table_id` are known, and view/filter/visible-field failures return `ok=true`, `partial=true` with enrichment warnings instead of discarding the created board state.
- Active-state user todos now override the connected/latest-run attention item path, so connected-read-only passive runs can still project the concrete user gate.

Validation passed locally:

```bash
python3 -m py_compile loopx/*.py loopx/cli_commands/*.py
python3 examples/lark-kanban-control-plane-smoke.py
python3 examples/active-state-todo-attention-fallback-smoke.py
python3 -m loopx.cli --format json lark-kanban schema
git diff --check origin/main...HEAD
loopx check --scan-path loopx/lark_kanban.py --scan-path loopx/status.py --scan-path examples/lark-kanban-control-plane-smoke.py --scan-path examples/active-state-todo-attention-fallback-smoke.py --scan-path docs/lark-kanban-control-plane-adapter.md
```

`loopx check` only reported the existing duplicate index-row warning, unrelated to this PR. I think this is mergeable now.

PR #604, opened 2026-06-23
Add loopx mode for Claude Code CLI

liangsalt:
## Summary

Adds **loopx goal-mode for Claude Code** — a constructed goal-mode that lets loopx
(the deterministic control plane) drive Claude Code on a long-running goal. It is
the Claude analogue of loopx's Codex integration: Codex has a native goal mode +
OS sandbox; Claude Code does not, so goal-mode is built from Claude Code's
extension points (slash command + hooks + MCP + statusline).

Everything lives under `loopx/claude_goal_mode/` and is wired into `~/.claude` by
the existing `scripts/install-local.sh`. No changes to the loopx kernel.

## What it looks like

```
❯ /loopx write RTL code

  I'll make the one-time safety note, then begin the loop.

  Safety note: this goal runs largely unattended, gated only by the loopx hook
  (read-only allowed; writes confined to the goal scope). Enabling auto mode
  (Shift+Tab) adds a classifier that catches exfil/escalation the hook can't —
  do you want it on? I'll begin now; answer whenever.

  Called loopx (should_run) → Quota OK (1 slot, eligible). Claimed the open todo,
  inspected the project, checked the sim toolchain (iverilog/vvp/yosys)…

  [loopx cc-test0621 · ▶ 0✓/1▸ · next: write RTL code]      ← statusline
```

One `/loopx <task>` sets up the goal, surfaces a one-time auto-mode safety choice,
then works **in-session** — the Stop hook keeps it advancing across turns until
loopx says stop.

## Usage

| Command | Effect |
|---|---|
| `/loopx <task>` | bootstrap/ensure a goal, add the task, arm goal-mode, start working in-session |
| `/loopx` (no args) | arm goal-mode for this project's existing goal |
| `/loopx background` | also install the **opt-in** headless launchd timer (cross-session autonomy) |
| `/loopx status` | full goal detail: objective, state, next action, open todos |
| `/loopx off` | disarm goal-mode (and stop the background timer if installed) |

Install: `scripts/install-local.sh` wires the PreToolUse hook + Stop hook +
statusline + MCP + `/loopx` into `~/.claude` (skipped when `claude` is not on PATH,
so a Codex-only machine gets no stray config).

## How it works

- **PreToolUse hook is the whole gate** (`hooks/goal_policy.py`): `should_run`
  (fail-closed) + Edit/Write `write_scope` + a destructive-bash denylist. Decisions
  are deterministic (no LLM); loopx decides what runs.
- **Stop hook** (`hooks/goal_stop.py`): deterministic in-session continuation —
  after each turn, if `should_run` is true and todos remain, keep going. loopx
  decides, **not** Claude Code's `/goal` (which judges completion from the chat
  transcript).
- **MCP server** (`mcp/loopx_mcp.py`): `should_run` / `list_todos` / `claim_task` /
  `complete_task`.
- **Statusline** (`statusline/goal_status.py`): running / paused+reason / ⚠ needs-you,
  plus next action and todo progress.
- **Background timer is opt-in** (`/loopx background`): default `/loopx` relies on
  the in-session Stop hook, so there is **no unattended headless loop unless you
  ask for one**. Armed state is a marker (`goal_state.py`), decoupled from the timer.
- Installer resolves the real python interpreter (`python3` → `python` → current),
  so wiring works on machines that register `python` but not `python3`.

## Isolation & safety

loopx drives Claude Code **directly — this PR ships without an OS sandbox**.
Containment is the hook (should_run + write_scope + destructive-bash denylist),
which is loopx's original posture: fine for trusted local work, but Bash is gated
only by a denylist. So:

- **Unattended runs:** `/loopx` makes a one-time offer to enable Claude Code
  **auto mode** (a classifier that catches exfil/escalation the hook can't).
- **Untrusted code / high-stakes unattended:** run Claude Code inside a **dev
  container / VM** (environment-level isolation).

> We explored wiring loopx into Claude Code's native **`/sandbox`** (the sandboxed
> Bash tool — Seatbelt on macOS / bubblewrap on Linux) to match Codex's
> `workspace-write`. The per-command `/sandbox` boundary ran into real friction
> with loopx's external control plane (the launchd heartbeat, the global registry
> under `~/.codex`, and headless network for builds). It was an exploratory attempt
> that hit problems, so **this PR intentionally ships without the `/sandbox`
> integration for now** and recommends environment-level isolation (container/VM)
> instead. Combining with `/sandbox` can be revisited as follow-up.

## Discussion (for maintainer): global runtime root is `~/.codex/loopx`

loopx's global runtime root defaults to `~/.codex/loopx` (`loopx/paths.py:
DEFAULT_RUNTIME_ROOT`) and each project registry pins `common_runtime_root`. It is
shared across agents (Codex/Claude). For the Claude adapter this is a Codex-branded
path. De-Codex-ifying it (e.g. `~/.loopx`) is a **kernel** change that affects Codex
too and needs migration of existing `~/.codex/loopx/` state, so it is out of scope
here. Follow-up options: (1) keep the shared root; (2) add a `LOOPX_RUNTIME_ROOT`
env override (low-risk); (3) rename the kernel default + migrate via the existing
`state_migration.py`. Flagging for direction.

## Scope

- New: `loopx/claude_goal_mode/**`, `loopx/claude_goal_baseline.py`.
- Changed: `scripts/install-local.sh` (wire the Claude installer; skip when `claude`
  absent; naming).
- No loopx kernel changes.


huangruiteng:
Thanks for pushing this. I think the direction is valuable: Claude Code is an important LoopX runtime target, and this PR makes the intended bridge much more concrete.

Before merge, I think we need to tighten the installation/control-plane boundary. The current implementation wires LoopX into user-level Claude Code config from the normal LoopX installer:

- `scripts/install-local.sh` defaults `LOOPX_INSTALL_CLAUDE` to enabled when `claude` is on PATH.
- `install.py` writes `~/.claude/settings.json`, registers user-scope MCP, installs a user-level `/loopx` command, and adds global `PreToolUse` / `Stop` hooks.
- Those hooks are matcher-wide, so even if they often no-op outside armed LoopX projects, a hook bug, latency, goal-id collision, or config side effect can affect unrelated Claude Code projects.
- Please also avoid removing existing user permission deny rules such as `Read(~/.ssh/**)` / `Read(~/.aws/**)`. Even if those came from an older LoopX install, deleting global permission rules should require explicit user approval.

My requested shape:

1. Make Claude integration explicit opt-in, not a default side effect of `scripts/install-local.sh`. For example, default `LOOPX_INSTALL_CLAUDE=0`, or move this behind a separate `loopx claude install` / `python .../install.py` command with a clear warning.
2. Prefer project-level `.claude/settings.json` for hooks, or expose a deliberate `--scope user` / `--scope project` choice. User-scope hooks should be visibly acknowledged.
3. Do not mutate unrelated existing Claude permission rules. If legacy cleanup is needed, print a suggested manual cleanup command instead.
4. Add a focused smoke or dry-run test that proves the normal LoopX install does not write `~/.claude` unless the Claude adapter is explicitly requested.

There is also an optional third route worth considering. Claude Code now has native `/loop`, `/goal`, and `/background` commands. That is a different design path from this PR's constructed hook/heartbeat approach: use Claude's native `/loop` as the scheduler/executor, and let LoopX provide the control-plane protocol (`should_run -> claim todo -> bounded work -> verify -> complete_task/spend`). In that route, LoopX would mainly ship MCP tools plus a lightweight `/loopx` helper that generates a native `/loop ...` prompt, while hooks remain optional hardening rather than the default runtime.

I do not think that third route has to replace this PR, but the PR should explicitly choose between the two models: constructed LoopX-owned Claude goal mode vs Claude-native `/loop` + LoopX control-plane adapter. Either can be valid, but the global install side effect needs to be fixed first.

liangsalt:
Thanks — this was exactly the right boundary to tighten. I've pushed updates (`d3c51d3`, `4b36cca`) that fix the install/control-plane boundary and make the design model explicit.

**Design model:** I chose the **Claude-native `/loop` + LoopX control-plane adapter** (your second option). Native `/loop` is the runtime — it executes a project `.claude/loop.md` that `/loopx` writes; LoopX provides the control-plane protocol via MCP (`should_run` → `claim_task` → bounded, verified work → `complete_task`/spend). Hooks are optional hardening, **not** the default runtime. The constructed runtime (launchd heartbeat, `Stop` hook, headless drivers) has been removed.

On your requested shape:

1. **Explicit opt-in.** `scripts/install-local.sh` now defaults `LOOPX_INSTALL_CLAUDE=0`; the normal install never writes `~/.claude`. The adapter installs only when explicitly requested (`LOOPX_INSTALL_CLAUDE=1`, or by running `install.py` directly).
2. **Deliberate scope; no global hooks by default.** `install.py` now requires `--scope user|project` (no default). Project scope is preferred and confined to `<project>/.claude`; user scope prints a banner acknowledging it affects all projects. The default install wires only the MCP server + the `/loopx` command — **no hooks**. The PreToolUse `should_run` gate is opt-in via `--harden` and is project-scoped. The `Stop` hook is gone entirely.
3. **No mutation of existing permission rules.** Install never deletes permission rules. A legacy LoopX creds-deny now prints a suggested manual-cleanup command instead of auto-removing anything.
4. **Focused smoke/dry-run test.** `examples/claude-install-optin-smoke.py` proves: the normal install doesn't touch `~/.claude`; an explicit `--scope` is required; a project-scope install is isolated and writes no hooks by default; `--harden` is opt-in and project-scoped; and existing permission rules are preserved. It includes a dry-run-writes-nothing assertion.

I also added `loopx/claude_goal_mode/README.md` documenting the opt-in install commands, usage, and uninstall. Happy to adjust scope/naming further.

huangruiteng (changes requested):
Thanks for the update. The revised direction is much safer than the first version: default install stays opt-in, `install.py` requires explicit scope, project-scope install is isolated, and the smoke covers the install boundary. I also verified:

- `python3 examples/claude-install-optin-smoke.py`
- `python3 -m py_compile ...` for the new Claude adapter modules
- a local `/loopx <task>` empty-project smoke that bootstraps/registers/seeds/writes `.claude/loop.md`

I still see two blocking lifecycle issues before this should merge:

1. **The active goal resolver can pick the wrong goal in multi-goal projects.** `goal_state.goal_context()` selects the first registry goal with `coordination.primary_agent`, falling back to `goals[0]`. That is not the same thing as the Claude project’s armed `/loopx` goal. In a real LoopX repo with multiple goals, the hook/statusline/MCP can run `quota should-run`, claim, and complete against the wrong goal. Please persist or derive the armed goal id from the `.claude/loop.md` setup state, or call a canonical LoopX resolver instead of guessing from registry order.

2. **`complete_task(next_agent_todo=...)` hard-codes `--next-claimed-by cc-controller`, but `/loopx` only registers `cc` by default.** I reproduced this with a temp default `cc` goal: `todo complete --claimed-by cc --next-agent-todo ... --next-claimed-by cc-controller` fails with `next_claimed_by='cc-controller' is not registered ... registered_agents=cc`. This breaks the important complex-task path where an iteration completes one todo and creates the next. Please either use the active primary/current agent for next claim ownership, or delegate to the existing LoopX primary/side-agent completion semantics rather than hard-coding a new controller id.

One doc-level follow-up: the README currently says `--harden` denies out-of-scope commands and denies everything when `should_run` is false, but the implementation allows read-only tools before the gate, gates Bash by a denylist, and defers unknown tools. That limitation is okay if intentional, but the README should match the actual security boundary so users do not treat it as strong isolation.

liangsalt:
Both lifecycle issues and the doc follow-up are fixed in `fe36268`, with a focused test (`examples/claude-goalmode-lifecycle-smoke.py`).

**1. Active-goal resolver no longer guesses by registry order.** `/loopx` now persists the armed goal (`goal_id` + `agent_id`) in a `loopx:armed` marker at the top of `.claude/loop.md`, and `goal_state.goal_context()` resolves *that* goal first — used uniformly by the hook, statusline, and MCP. It falls back to the previous primary-agent/first-goal heuristic only when the project isn't armed, or when the marker points at a goal no longer in the registry. The test covers a two-goal registry (resolves the armed goal, not registry order) and the stale-marker fallback.

**2. `complete_task(next_agent_todo=...)` no longer hard-codes a controller id.** I dropped `--next-claimed-by cc-controller`. With `--next-agent-todo` and no `--next-claimed-by`, LoopX assigns the next todo to the completing agent (and to `primary_agent` for side-agent review) via its own completion semantics, so it always lands on a registered agent. Verified against your repro on a default `cc` goal:
- old: `todo complete … --next-agent-todo X --next-claimed-by cc-controller` → `ok: False`, `next_claimed_by='cc-controller' is not registered … registered_agents=cc`
- new: `todo complete … --next-agent-todo X` → `ok: True`, next todo created and owned by `cc`.

**3. README now matches the actual `--harden` boundary.** It documents that read-only tools are allowed before the gate, `should_run=false` is fail-closed for non-read-only tools, `Edit`/`Write` are confined to `write_scope`, `Bash` is gated by a destructive-command denylist (not confined to scope), and unknown tools defer to the normal permission flow — and states explicitly that this is a deterministic policy layer, **not strong isolation** (use a container/VM for untrusted work).

Anyone can confirm both fixes with `python3 examples/claude-goalmode-lifecycle-smoke.py`.

huangruiteng (changes requested):
Thanks for the latest fixes. I re-ran the focused validation and the previous blockers are materially improved: normal LoopX install is opt-in for Claude, the design is now Claude-native `/loop` + LoopX control-plane MCP, the armed-goal resolver no longer guesses by registry order, and the next-todo owner path no longer hard-codes `cc-controller`.

Validation I ran:

- `python3 examples/claude-install-optin-smoke.py` ✅
- `python3 examples/claude-goalmode-lifecycle-smoke.py` ✅
- `python3 -m py_compile ...` for the new Claude adapter modules ✅
- `git diff --check origin/main...HEAD` ✅
- public/private boundary scan over the PR files ✅

I still think this needs changes before merge because this is a runtime/permission-sensitive adapter:

1. **[P1] Do not auto-mutate the user's system Python as a fallback.** In `loopx/claude_goal_mode/scripts/install.py`, `provision_mcp_python()` falls back to `pip install --break-system-packages mcp` when the dedicated venv path fails (`install.py` around lines 108-110). Even with explicit Claude opt-in, a project-scoped adapter install should not silently modify the user's active/system Python environment. Please remove this automatic fallback and instead fail with a manual instruction, or put it behind a very explicit flag such as `--allow-system-pip`.

2. **[P1] Do not delete the legacy `goal-harness` MCP entry during install.** `install_mcp()` removes both `loopx` and `goal-harness` MCP entries before adding the new one (`install.py` around lines 129-130). Installing the new `loopx` MCP does not require deleting a user-owned or legacy MCP entry, and user scope would affect all Claude Code projects. Please only replace the `loopx` entry by default; for `goal-harness`, print a manual cleanup/migration note or gate removal behind an explicit migration flag.

3. **[P1] `Task` should not be treated as read-only/safe by the hardening hook.** `goal_policy.py` includes `Task` in `READONLY_TOOLS`, so an armed project allows `Task` before consulting `should_run`. In Claude Code, `Task` can launch an agent that performs writes, so this can bypass the deterministic gate when `should_run=false`. Please remove `Task` from the unconditional allow-list and add a focused assertion that `should_run=false` denies `Task`, while `should_run=true` can defer/allow normal Claude permission flow.

Open question / non-blocking if you prefer to keep it best-effort: `connect.py` currently ignores non-zero exit codes from both the `loopx connect` step and the delegated project install (`connect.py` around lines 49 and 82), then still prints the normal next steps. For a setup helper, I would prefer fail-closed behavior when registry connection or install fails, especially because global route collision protection now intentionally returns non-zero.

**Merge decision:** hold / request changes for the three P1 items above.

PR #715, opened 2026-06-26
Define event-sourced LoopX state contract

huangruiteng:
## Summary
- add event_sourced_state_contract_v0 for moving canonical todo/history truth to append-only events while keeping ACTIVE_GOAL_STATE.md as the human/agent workbench
- link the contract from protocol docs and the state interaction model
- add a smoke covering deterministic replay, same-priority ordering, idempotent append, mutation rejection, and public/private boundary terms

## Validation
- python3 examples/event-sourced-state-contract-smoke.py
- python3 examples/task-graph-projection-fixture-smoke.py
- git diff --check
- git diff --cached --check
- loopx check --scan-path docs/reference/protocols/event-sourced-state-contract-v0.md --scan-path docs/reference/protocols/README.md --scan-path docs/state-interaction-model.md --scan-path examples/event-sourced-state-contract-smoke.py

Note: docs-governance-smoke currently fails on the pre-existing docs root Markdown count threshold; this PR only adds a protocol subdirectory doc and one example smoke.

### DIFF
--- a/docs/reference/protocols/README.md
+++ b/docs/reference/protocols/README.md
@@ -15,6 +15,7 @@ Current contracts:
 - [task_graph_projection_v0](task-graph-projection-v0.md)
 - [long_horizon_agent_state_protocol_v0](long-horizon-agent-state-protocol-v0.md)
 - [global_manager_command_v0](global-manager-command-v0.md)
+- [event_sourced_state_contract_v0](event-sourced-state-contract-v0.md)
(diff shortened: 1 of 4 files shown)


huangruiteng:
Self-review result: no blocking findings. Revalidated in the full PR batch: event-sourced state contract smoke, git diff --check, py_compile for the smoke set, and loopx boundary check passed. The change is public protocol/documentation plus focused smoke coverage; no private state, credentials, local paths, or raw logs are included.

PR #744, opened 2026-06-27
Raise smoke review threshold to 500 lines

huangruiteng:
## Summary
- update AGENTS smoke retention guidance to re-check smoke structure after roughly 500 lines instead of 300
- preserve the existing guidance to split reusable logic and keep smokes thin

## Validation
- git diff --check
- rg confirmed only the 500-line threshold remains


huangruiteng:
Self-review: single-line AGENTS.md contributor guidance update from roughly 300 to 500 lines for smoke-size re-check. Scope is contributor metadata only; no runtime, benchmark runner, scoring, public/private policy, or generated artifact changes. Validation: git diff --check; rg confirmed the old 300-line threshold was replaced.

PR #755, opened 2026-06-27
Handle SkillsBench host-local ACP return shape

huangruiteng:
## Summary
- make the SkillsBench host-local ACP override match BenchFlow versions that return either a 3-tuple or legacy 4-tuple from connect_acp
- avoid importing benchflow.agents.protocol unless the installed BenchFlow actually expects a session adapter
- expose compact public prerequisite fields for the detected return arity/session adapter path
- add route smoke coverage for 3-tuple, 4-tuple, and unannotated connect_acp signatures

## Validation
- python3 -m py_compile scripts/skillsbench_automation_loop.py examples/skillsbench-goal-start-product-mode-route-smoke.py
- git diff --check
- patch-only boundary scan for private paths/raw evidence/secrets: no new matches

## Scope
Benchmark helper/runtime compatibility only. Does not change scoring, task semantics, leaderboard/submission behavior, permission boundaries, or launch jobs.

### DIFF
--- a/examples/skillsbench-goal-start-product-mode-route-smoke.py
+++ b/examples/skillsbench-goal-start-product-mode-route-smoke.py
@@ -231,6 +231,26 @@ def _assert_control_score_surface() -> None:
     assert compacted_prerequisites["goal_start_planned_todo_count_expected"] == 3
 
 
+def _assert_host_local_acp_return_arity_compat() -> None:
+    sys.path.insert(0, str(REPO_ROOT))
+    from scripts.skillsbench_automation_loop import (
+        _benchflow_connect_acp_return_arity,
+    )
+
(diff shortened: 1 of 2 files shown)


huangruiteng:
Self-review: this is a narrow benchmark helper/runtime compatibility patch. The host-local ACP override now introspects BenchFlow connect_acp return shape and only imports ACPSessionAdapter for 4-tuple runtimes, which fixes current SkillsBench BenchFlow 3-tuple installs without changing scoring, task semantics, submission/leaderboard behavior, permission boundaries, or job launch behavior. Validation passed locally: py_compile, goal-start route smoke, git diff --check, and patch-only public/private boundary scan.

PR #784, opened 2026-06-27
Guard SkillsBench stdin writes with timeout watchdogs

huangruiteng:
## Summary
- feed reverse-channel Codex stdin on a daemon writer thread so watchdog deadlines start even if the child does not drain stdin
- apply the same guard to the SkillsBench local ACP relay
- add a focused smoke for a child process that never reads a large stdin prompt

## Validation
- python3 -m py_compile scripts/skillsbench_reverse_channel_bridge.py loopx/benchmark_adapters/skillsbench_acp_relay.py examples/skillsbench-benchmark-run-smoke.py
- targeted reverse-channel timeout smokes passed

## Scope
Benchmark helper/runtime only. Does not change scoring, task semantics, leaderboard/submission behavior, or launch benchmark jobs.

### DIFF
--- a/examples/skillsbench-benchmark-run-smoke.py
+++ b/examples/skillsbench-benchmark-run-smoke.py
@@ -162,6 +162,25 @@ def test_reverse_channel_first_action_timeout_stops_codex_process() -> None:
     assert response["credential_values_recorded"] is False
 
 
+def test_reverse_channel_timeout_survives_blocked_stdin_write() -> None:
+    start = time.monotonic()
+    response = _run_codex_payload(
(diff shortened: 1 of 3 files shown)


huangruiteng:
Self-review: benchmark helper/runtime-only fix. The failure mode is a blocked stdin pipe before timeout polling starts; this moves stdin writes off the watchdog thread and adds a focused smoke that would hang on the old path. Local validation passed: py_compile plus targeted reverse-channel timeout smokes. No scoring/task semantics/leaderboard/submission changes.

PR #889, opened 2026-06-29
Fix SkillsBench verifier uv bootstrap staging

huangruiteng:
## Summary
- patch staged SkillsBench verifier wrappers to provide uv installer tarball downloads through the public releases.astral.sh mirror when tasks use astral.sh/uv install.sh
- propagate public-safe uv bootstrap patch metadata through compact status and benchmark ledger reducers
- add SkillsBench smoke coverage for staging, preflight, compact, and reduce-only recovery

## Validation
- python3 -m py_compile scripts/skillsbench_automation_loop.py examples/skillsbench-benchmark-run-smoke.py loopx/status.py loopx/benchmark_ledger.py
- git diff --check
- python3 examples/skillsbench-benchmark-run-smoke.py

No scoring, task semantics, leaderboard behavior, private material, raw trajectories, verifier output, or local paths are included.

### DIFF
--- a/examples/skillsbench-benchmark-run-smoke.py
+++ b/examples/skillsbench-benchmark-run-smoke.py
@@ -56,6 +56,7 @@ from scripts.skillsbench_automation_loop import (  # noqa: E402
     DEFAULT_MAX_ROUNDS,
     DEFAULT_PRODUCT_MODE_SOFT_VERIFY_POLICY,
     DEFAULT_SOFT_VERIFIER_TIMEOUT_SEC,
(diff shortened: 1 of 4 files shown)


huangruiteng:
Self-review note: narrow SkillsBench runtime/helper fix. Verified public-safe metadata path and focused smokes locally: py_compile, git diff --check, examples/skillsbench-benchmark-run-smoke.py. No scoring/task semantics, leaderboard behavior, private evidence, raw trajectories, verifier output, or local paths are included.

PR #890, opened 2026-06-29
Fix verifier uv mirror URL base

huangruiteng:
## Summary
- fix the SkillsBench staged uv bootstrap mirror patch to set INSTALLER_DOWNLOAD_URL to the release directory base expected by uv's installer
- update smoke fixtures so we do not pre-compose the uv target tarball path

## Validation
- git diff --check
- python3 examples/skillsbench-benchmark-run-smoke.py

Follow-up to #889 after live citation-check showed uv appends the target tarball name itself.


huangruiteng:
Self-review note: tiny follow-up to #889. Live verifier run showed uv appends the target tarball name to INSTALLER_DOWNLOAD_URL, so the patch must provide the release directory base. Local validation passed: py_compile, git diff --check, examples/skillsbench-benchmark-run-smoke.py.

PR #911, opened 2026-06-29
Require task progress for SkillsBench app goal turns

huangruiteng:
## Summary
- Require SkillsBench task-stage app-server/ACP turns to produce meaningful task-facing bridge progress within the first-action window.
- Treat no-bridge and status-only bridge turns as recoverable first-action failures instead of allowing long idle runs.
- Merge app-server bridge operation compact traces into the common controller/reducer counters so compact results and the ledger can observe task-facing bridge activity.

## Root Cause
The previous app-server route only knew that the native worker was alive or that some bridge file activity happened. That let internal app-server/TUI activity, or lifecycle-only commands such as status probes, masquerade as benchmark progress. The runner could then wait for a long timeout even when the worker had not actually operated on the task sandbox.

## Validation
- python3 -m py_compile loopx/benchmark_adapters/skillsbench_acp_relay.py scripts/skillsbench_automation_loop.py examples/skillsbench-app-server-goal-worker-smoke.py examples/skillsbench-benchmark-run-smoke.py
- python3 examples/skillsbench-app-server-goal-worker-smoke.py
- focused: test_skillsbench_controller_trace_counts_are_compacted

No raw task text, trajectories, verifier output, credentials, or local private paths are added.

### DIFF
--- a/examples/skillsbench-app-server-goal-worker-smoke.py
+++ b/examples/skillsbench-app-server-goal-worker-smoke.py
@@ -1059,6 +1059,248 @@ sys.exit(0)
         proc.wait(timeout=2)
 
 
+def test_acp_relay_fails_fast_when_app_server_worker_never_uses_bridge() -> None:
+    fake_worker = """#!/usr/bin/env python3
+import time
+
+time.sleep(30)
+"""
+    with tempfile.TemporaryDirectory(prefix="skillsbench-app-goal-bridge-first-action-") as tmp:
+        root = Path(tmp)
+        worker = root / "fake_worker.py"
+        work = root / "work"
+        trace_dir = root / "worker-traces"
+        worker.write_text(fake_worker, encoding="utf-8")
+        worker.chmod(0o755)
+        work.mkdir()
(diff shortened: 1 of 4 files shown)


huangruiteng:
Self-review: no blocking issues found.

Why this is a root-cause fix rather than trace-only:
- The app-server/ACP relay now requires meaningful task-facing bridge progress for task-stage prompts. Internal app-server activity alone is not enough.
- A worker that never uses the bridge is killed at first-action timeout.
- A worker that only calls lifecycle/status bridge commands is also killed as no meaningful task progress.
- The reducer now ingests app-server bridge operation traces, so compact results and the common case ledger can see the same task-facing evidence that the runner uses.

Validation:
- py_compile on changed runtime/reducer/smoke files
- skillsbench-app-server-goal-worker smoke ok
- skillsbench-goal-start-product-mode-route smoke ok
- focused controller trace compact smoke ok

Boundary: no raw task text, raw trajectories, verifier output, credentials, local private paths, or generated benchmark logs are included.

PR #912, opened 2026-06-29
Isolate parallel SkillsBench batch cases

huangruiteng:
## Summary
- Run parallel SkillsBench batch cases in per-case subprocesses instead of in the same Python process.
- Preserve the existing in-process path for serial batch/single-case execution.
- Add a focused smoke for child argv generation: no inherited task-ids, per-case task-id, parallel-cases=1, suffixed job/rollout names, and app-server/bridge flags preserved.

## Root Cause
A 30-case r8 run showed an adaptive-cruise-control worker launching with a reverse-channel bridge command for the 3d-scan-calc Docker project. The actual issue is not trace loss: parallel batch execution runs multiple BenchFlow cases in one Python process while run_benchflow_case patches module-global ACP hooks. Concurrent cases can therefore use another case's connect/bridge closure.

This PR fixes the isolation boundary so parallel cases cannot share those global BenchFlow hooks.

## Validation
- focused: test_skillsbench_parallel_batch_isolates_case_process_argv
- plan-only 2-case parallel batch exercised the subprocess path successfully
- python3 examples/skillsbench-app-server-goal-worker-smoke.py

Boundary: no raw benchmark logs/tasks/trajectories, credentials, or local private paths added to tracked files.

### DIFF
--- a/examples/skillsbench-benchmark-run-smoke.py
+++ b/examples/skillsbench-benchmark-run-smoke.py
@@ -9338,6 +9338,55 @@ def test_skillsbench_runner_plan_supports_controller_trace_path() -> None:
         assert plan["include_task_skills"] is False, plan
 
 
+def test_skillsbench_parallel_batch_isolates_case_process_argv() -> None:
+    args = parse_args(
+        [
+            "--task-ids",
+            "3d-scan-calc,adaptive-cruise-control",
+            "--parallel-cases",
+            "2",
+            "--route",
+            "codex-app-server-goal-baseline",
+            "--run-group-id",
+            "skillsbench-goal-baseline-30case-test",
+            "--job-name",
+            "skillsbench-goal-baseline-30case-test",
+            "--rollout-name",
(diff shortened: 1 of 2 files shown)


huangruiteng:
Self-review: no blocking issues found.

Root-cause check:
- r8 exposed a real bridge cross-talk bug: one case's host ACP bridge closure was used by another case in the same batch process.
- The fix moves parallel batch cases into subprocesses, isolating BenchFlow module-global ACP monkeypatches per case.
- Serial/single-case behavior is unchanged.
- This is not a trace-only change; it changes the execution isolation boundary that caused the wrong sandbox bridge.

Validation:
- py_compile for changed script/smoke
- focused parallel batch argv isolation smoke
- plan-only two-case parallel batch through real subprocess path
- app-server goal worker smoke
- goal-start product-mode route smoke

Boundary: tracked changes contain no raw benchmark evidence, local paths, credentials, or generated logs.

PR #972, opened 2026-06-30
Record install source revision in doctor

huangruiteng:
## Summary

- record source checkout git metadata in release manifests when `install-local.sh` builds a local release snapshot
- surface the manifest source commit/ref/dirty state through `loopx doctor` JSON and markdown
- extend `examples/install-local-smoke.py` so the canary fails if local checkout provenance regresses back to `local_checkout @ n/a`

Motivation: a just-merged source fix can be hidden by the default installed release snapshot. Before this change, doctor could only show `local_checkout @ n/a`, so operators had to manually cross-check source and wrapper behavior.

Solved or not: the change adds a reusable manifest provenance contract and validates the local-install read path end to end. Archive installs keep working without git metadata.

User/operator impact: `loopx doctor` can now explain which source revision produced the installed snapshot, and the install smoke catches the exact n/a regression that made the previous stale-wrapper issue hard to diagnose.

Main risk: installed-user compatibility. The manifest fields are additive, archive installs tolerate missing git metadata, and existing update reads continue to pass.

Design judgment: this is a general release provenance surface, not a one-off special case. A later refinement could compare the manifest commit against a live checkout commit when both are available, but the first step is making the installed snapshot self-describing.

## Validation

- `python3 -m py_compile loopx/release_manifest.py loopx/doctor.py`
- `python3 examples/install-local-smoke.py`
- `python3 examples/codex-cli-packaged-install-smoke.py`
- `python3 examples/loopx-update-smoke.py`
- `python3 -m loopx.cli check --scan-path loopx/release_manifest.py --scan-path loopx/doctor.py --scan-path scripts/install-local.sh --scan-path examples/install-local-smoke.py --limit 200`
- `git diff --check`

Boundary scan: `loopx check` public/private scan was clean for the changed files; it reported only the pre-existing duplicate index rows warning for `loopx-meta`.

### DIFF
--- a/examples/install-local-smoke.py
+++ b/examples/install-local-smoke.py
@@ -21,6 +21,15 @@ if str(REPO_ROOT) not in sys.path:
 from loopx.doctor import add_promotion_readiness_freshness, build_install_freshness  # noqa: E402
 
 
+def git_output(args: list[str]) -> str:
+    return subprocess.run(
+        ["git", "-C", str(REPO_ROOT), *args],
+        check=True,
+        capture_output=True,
+        text=True,
+    ).stdout.strip()
+
+
 def run_install(env: dict[str, str], release_id: str) -> subprocess.CompletedProcess[str]:
     return subprocess.run(
         [str(INSTALL_SCRIPT)],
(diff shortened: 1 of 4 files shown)


huangruiteng:
Self-review: no blocking findings. This is a narrow additive install/doctor provenance change: manifests now record local source git commit/ref/dirty state, doctor projects it, and install-local-smoke fails if local checkout provenance regresses to local_checkout @ n/a. Validated locally with py_compile, install-local-smoke, codex-cli-packaged-install-smoke, loopx-update-smoke, loopx check, and git diff --check. GitHub does not allow approving my own PR, so merging under explicit owner self-merge authorization.

PR #990, opened 2026-06-30
Fix SkillsBench batch child payload parsing

huangruiteng:
## Summary
- skip parse-derived internal fail-fast provenance fields when serializing SkillsBench batch child CLI args
- preserve defaulted fail-fast semantics by letting child parse_args rederive formal route defaults
- recover structured child JSON payloads from stdout/stderr with harmless mixed output without recording raw output
- cover the batch child argv and payload recovery contracts in the SkillsBench benchmark smoke

## Validation
- focused batch subprocess smoke for argv isolation and mixed stderr payload recovery
- plan-only --task-ids ... --parallel-cases 2 subprocess batch e2e
- git diff --check
- loopx check --scan-path scripts/skillsbench_automation_loop.py --scan-path examples/skillsbench-benchmark-run-smoke.py --limit 200

### DIFF
--- a/examples/skillsbench-benchmark-run-smoke.py
+++ b/examples/skillsbench-benchmark-run-smoke.py
@@ -10038,6 +10038,11 @@ def test_skillsbench_parallel_batch_isolates_case_process_argv() -> None:
     )
     child_argv = skillsbench_loop._batch_case_args_to_cli(case_args)
     assert "--task-ids" not in child_argv, child_argv
+    assert "--apt-risk-fail-fast-defaulted" not in child_argv, child_argv
+    assert "--bootstrap-light-fail-fast-defaulted" not in child_argv, child_argv
+    assert "--verifier-bootstrap-fail-fast-defaulted" not in child_argv, child_argv
+    assert "--fail-fast-on-apt-risk" not in child_argv, child_argv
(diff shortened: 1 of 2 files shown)


huangruiteng:
Findings: none.

Product/architecture judgment: this is a narrow SkillsBench runner contract fix. The bug was that batch case subprocess isolation serialized parse-derived in-memory fields back into child CLI args, which could make child argparse fail before the structured setup-preflight closeout reached the parent. The patch keeps those internal/default provenance fields out of the child CLI, preserves formal route default derivation in the child, and makes the parent recover structured child JSON from stdout/stderr when harmless mixed output is present, without recording raw output.

Validation performed locally:
- `python3 -m py_compile scripts/skillsbench_automation_loop.py examples/skillsbench-benchmark-run-smoke.py`
- focused batch subprocess smoke for argv isolation and mixed stderr payload recovery
- plan-only `--task-ids ... --parallel-cases 2` subprocess batch e2e
- `git diff --check`
- `loopx check --scan-path scripts/skillsbench_automation_loop.py --scan-path examples/skillsbench-benchmark-run-smoke.py --limit 200` (public boundary clean; only existing duplicate-index warning)

Merge decision: self-merge under the benchmark helper/runtime authorization; no scoring/task semantics, permission boundary, raw evidence, or leaderboard behavior changes.

PR #1018, opened 2026-07-01
Document handoff gate successor replan state

huangruiteng:
## Summary
- add `successor_replan_required` to the core control-plane derived state definitions
- route cleared handoff gates without successor/no-follow-up through `SuccessorReplan` in the state machine
- point the core IP-029 catalog lens at `examples/quota-cleared-blocker-successor-gate-smoke.py`

## Validation
- `python3 examples/quota-cleared-blocker-successor-gate-smoke.py`
- `python3 -m loopx.cli check --scan-path docs/product/core-control-plane/state-definitions.md --scan-path docs/product/core-control-plane/state-machine.md --scan-path docs/product/core-control-plane/interaction-catalog.md`

## Notes
- `loopx check` passed with the known duplicate index warning only; public boundary scan was clean for the three touched docs.


huangruiteng:
Self-review notes:

- Scope is docs-only and single-purpose: connect the existing IP-029 handoff-gate smoke pattern back to the core control-plane state definitions and state machine.
- The change does not introduce a new IP or a second source of truth; it makes the existing `successor_replan_required` runtime behavior visible in the core graph.
- Boundary scan is clean for the three touched docs; no private evidence, local paths, raw benchmark material, credentials, or promotion claims are added.
- Validation passed: `python3 examples/quota-cleared-blocker-successor-gate-smoke.py`; `python3 -m loopx.cli check --scan-path ...` with only the known duplicate index warning.

No blockers found.

PR #1048, opened 2026-07-01
Align scheduler ack status lookback

huangruiteng:
## Summary
- Make `quota scheduler-ack` collect status with the same autonomous replan/history lookback used by `quota should-run`.
- Add smoke coverage that the CLI handler does not fall back to the default short `--limit=5` view when acknowledging scheduler RRULE state.

## Motivation
A heartbeat observed `quota should-run --agent-id codex-product-capability` recommending `FREQ=MINUTELY;INTERVAL=10` for `agent_scope_wait`, but `quota scheduler-ack --applied-rrule FREQ=MINUTELY;INTERVAL=10` rejected it because the ack path recomputed the scheduler hint from a shorter status window and expected a different RRULE/reset token. This made host cadence update and LoopX scheduler ack disagree.

## Validation
- `python3 examples/quota-scheduler-state-ack-smoke.py`
- `python3 -m compileall -q loopx/cli_commands/quota.py examples/quota-scheduler-state-ack-smoke.py`
- `git diff --check`
- public/private boundary scan over changed paths
- source-run against `loopx-meta`: fixed scheduler-ack accepts the `should-run` RRULE without requiring a manual `--limit 30` workaround

### DIFF
--- a/examples/quota-scheduler-state-ack-smoke.py
+++ b/examples/quota-scheduler-state-ack-smoke.py
@@ -19,6 +19,7 @@ if str(REPO_ROOT) not in sys.path:
 from loopx.policies.scheduler_hint import build_scheduler_hint  # noqa: E402
 from loopx.quota import AgentScopeFrontierAction  # noqa: E402
 from loopx.scheduler_state import SCHEDULER_STATE_SCHEMA_VERSION  # noqa: E402
+from loopx.status import AUTONOMOUS_REPLAN_PERIODIC_LOOKBACK  # noqa: E402
 
 
 AGENT_SCOPE_ACTIONS = [action.value for action in AgentScopeFrontierAction]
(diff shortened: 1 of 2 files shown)


huangruiteng:
Findings: none.

- Motivation: Codex App heartbeats apply `quota should-run.scheduler_hint.codex_app.recommended_rrule`, then call `quota scheduler-ack`. The live failure was that `should-run` used the larger autonomous replan lookback, while `scheduler-ack` recomputed the current scheduler hint from the default short status window and could expect a different RRULE/reset token.
- Solved: this PR makes `scheduler-ack` collect status with the same autonomous replan lookback as `should-run`, so ack validates the RRULE it just asked the host to apply. The smoke pins that CLI-handler contract.
- User/operator impact: host cadence changes no longer get stuck in an applied-but-unacknowledged state for side-agent wait/monitor lanes.
- Main risk: low but runtime-adjacent. The policy being acknowledged is unchanged; the PR only aligns the read window used to recompute it.
- Design judgment: this is the right layer for the fix because the drift was in the CLI command handler, not in the scheduler policy itself.

Validation:
- `python3 examples/quota-scheduler-state-ack-smoke.py`
- `python3 -m compileall -q loopx/cli_commands/quota.py examples/quota-scheduler-state-ack-smoke.py`
- public/private boundary scan over changed paths; only benign `reset_token` field-name hits.
- source-run against `loopx-meta` confirmed scheduler payload still renders and the ack path no longer requires a manual `--limit 30` workaround.

Merge decision: self-merged by explicit owner request; proceeding with admin-bypass merge.

PR #1169, opened 2026-07-03
Expose deferred successors in goal frontier

huangruiteng:
## Summary
- Add a compact `deferred_successors` projection to `goal_frontier_projection`, including ready/blocked counts and top ready todo ids.
- Add an explicit `acceptance_gaps` projection field so status/quota/diagnose consumers can rely on a stable shape even when no gap is present.
- Render deferred-ready and acceptance-gap counts in quota markdown and cover both monitor-only and ready-deferred-successor cases with focused smokes.

## Validation
- python3 -m py_compile loopx/policies/goal_frontier.py loopx/quota.py
- python3 examples/quota-replan-decision-plane-smoke.py
- python3 examples/work-lane-contract-smoke.py
- python3 examples/status-markdown-smoke.py
- python3 examples/next-action-projection-contract-smoke.py
- live quota shape check for loopx-meta/codex-side-bypass with PYTHONPATH=. confirmed no decision change and new projection fields present
- git diff --check
- public/private boundary scan over diff

This keeps the status sufficiency work in the goal-frontier projection seam. Quota only renders the compact facts; it does not absorb more per-agent vision or replan state-machine logic. The user/operator win is that agents can see ready deferred successors and explicit no-gap state from the same status/quota/diagnose payload instead of estimating from chat history.

### DIFF
--- a/examples/quota-replan-decision-plane-smoke.py
+++ b/examples/quota-replan-decision-plane-smoke.py
@@ -141,12 +141,20 @@ def assert_replan_beats_monitor_quiet_skip() -> None:
     assert guard["interaction_contract"]["agent_channel"]["must_attempt"] is True, guard
     assert guard["goal_frontier_projection"]["replan_required"] is True, guard
     assert guard["goal_frontier_projection"]["monitor_only_lanes"]["present"] is True, guard
+    assert guard["goal_frontier_projection"]["deferred_successors"] == {
+        "ready_count": 0,
+        "blocked_count": 0,
+        "current_agent_ready_count": 0,
+        "ready_todo_ids": [],
+    }, guard
+    assert guard["goal_frontier_projection"]["acceptance_gaps"] == [], guard
     assert guard["autonomous_replan_decision"]["decision_plane"] == (
         "goal_frontier_before_lane_quiet_or_agent_scope_wait"
     ), guard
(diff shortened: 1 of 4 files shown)


huangruiteng:
Self-review before admin-bypass merge.

Findings: none.

Product/architecture judgment: this is a narrow status/quota sufficiency improvement. It adds stable frontier fields for deferred successors and acceptance gaps inside the goal-frontier projection seam, while keeping quota.py as a consumer/renderer rather than a home for per-agent vision or replan policy. The main user/operator benefit is less chat-memory inference when deciding whether a lane has runnable successor work.

Validation performed locally:
- python3 -m py_compile loopx/policies/goal_frontier.py loopx/quota.py
- python3 examples/quota-replan-decision-plane-smoke.py
- python3 examples/work-lane-contract-smoke.py
- python3 examples/status-markdown-smoke.py
- python3 examples/next-action-projection-contract-smoke.py
- live quota shape check for loopx-meta/codex-side-bypass with PYTHONPATH=. confirmed no decision change and new projection fields present
- git diff --check
- public/private boundary scan over diff

Merge decision: self-merge. The PR is small, single-purpose, validated, public-safe, and does not touch Lark Kanban, first-screen presentation, permission boundaries, benchmark scoring, or runner semantics.

PR #1252, opened 2026-07-03
Allow SkillsBench egress auto direct fallback

huangruiteng:
## Summary
- add a direct TCP egress probe and let SkillsBench benchmark-egress proxy mode auto fall back to direct egress when a configured proxy is unavailable but direct access is verified
- prevent stale proxy env injection after direct fallback and project fallback state into public runner prerequisites/config
- document auto/require/off semantics and add a focused public-safe smoke

## Validation
- targeted benchmark-egress smokes: env forwarding, require-mode block, auto direct fallback
- loopx check --scan-path scripts/skillsbench_automation_loop.py --scan-path examples/skillsbench-benchmark-run-smoke.py --scan-path docs/benchmark-developer-workflow.md

## Boundary
- no real proxy endpoint, local path, raw trajectory, verifier output, or private run artifact is committed
- require mode remains fail-closed; only auto mode can direct-fallback after a successful public test-host probe

### DIFF
--- a/docs/benchmark-developer-workflow.md
+++ b/docs/benchmark-developer-workflow.md
@@ -1382,8 +1382,11 @@ standard proxy variables only to the private BenchFlow subprocess and agent
 environment, runs a bounded HTTP CONNECT preflight, and records only
 source/scheme/endpoint-kind/port plus `proxy_url_recorded=false` in public
 artifacts. Use `--benchmark-egress-proxy-mode require` when the run must fail
-fast without that private proxy, and `--benchmark-egress-proxy-mode off` only
-for explicit no-proxy debugging.
+fast without that private proxy. The default `auto` mode validates a configured
+proxy first, but if that proxy is unavailable and direct egress to the public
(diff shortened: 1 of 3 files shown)


huangruiteng:
Self-review before admin-bypass merge:

Findings: none.

Product/architecture judgment: this is a narrow SkillsBench benchmark-egress helper/runtime repair. It preserves require-mode fail-closed behavior, adds only auto-mode direct fallback after a public test-host probe succeeds, and records public-safe fallback state so future app-server xhigh countability debugging does not depend on private run logs. It does not change scoring, task semantics, leaderboard/submission behavior, or permission boundaries.

Validation performed:
- targeted benchmark-egress smokes for env forwarding, require-mode block, and auto direct fallback
- loopx check on the changed runner, smoke, and docs paths

Boundary: no real proxy endpoint, local path, raw trajectory, verifier output, or private artifact is committed. Merge decision: self-merge with admin bypass under the benchmark helper/runtime PR policy.

PR #1253, opened 2026-07-03
Filter internal aggregate flag from SkillsBench batch children

huangruiteng:
## Summary
- exclude the internal current-aggregate update boolean from SkillsBench batch child argv
- add a focused smoke asserting child argv does not emit the unrecognized aggregate flag and still forces child parallelism to 1

## Validation
- targeted batch smokes: plan-only parallel case contract and child argv filter
- reproduced the 5-case parallel plan-only command; it now returns ok=true instead of child payload errors
- loopx check on changed runner and smoke paths

## Boundary
- no real proxy endpoint, raw trajectory, verifier output, local/private run artifact, or task text is committed

### DIFF
--- a/examples/skillsbench-benchmark-run-smoke.py
+++ b/examples/skillsbench-benchmark-run-smoke.py
@@ -773,6 +773,34 @@ def test_skillsbench_plan_only_batch_parallel_case_contract() -> None:
         ] is True, result
 
 
+def test_skillsbench_batch_case_cli_filters_internal_aggregate_flag() -> None:
+    args = parse_args(
+        [
+            "--task-ids",
+            "citation-check,3d-scan-calc",
+            "--parallel-cases",
+            "2",
+            "--route",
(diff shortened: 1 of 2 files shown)


huangruiteng:
Self-review before admin-bypass merge:

Findings: none.

Product/architecture judgment: this is a narrow SkillsBench batch runner argv hygiene fix. It prevents an internal projection flag from being serialized into isolated child CLI invocations, which was causing parallel app-server xhigh reruns to fail before any case work. The change does not alter scoring, task semantics, benchmark prompts, egress policy, permission boundaries, or leaderboard behavior.

Validation performed:
- targeted batch smokes for plan-only parallel contract and child argv filtering
- 5-case parallel plan-only reproduction now returns ok=true
- loopx check on changed paths

Boundary: no private run artifacts, real proxy values, raw trajectories, verifier logs, or task text committed. Merge decision: self-merge with admin bypass under the benchmark helper/runtime PR policy.

PR #1357, opened 2026-07-04
Fix app-server Goal completed turn status fallback

huangruiteng:
## Summary
- Treat `turn/completed` events without an explicit status as completed in the app-server Goal driver.
- Update the SkillsBench app-server Goal worker smoke to exercise that event shape and assert the turn remains countable-clean.

## Validation
- `python3 examples/skillsbench-app-server-goal-worker-smoke.py`
- `python3 examples/codex-app-server-goal-driver-smoke.py`

## Boundary
- Public-safe compact smoke only; no raw task text, raw trajectories, verifier output, credentials, or local benchmark logs are added.

### DIFF
--- a/examples/skillsbench-app-server-goal-worker-smoke.py
+++ b/examples/skillsbench-app-server-goal-worker-smoke.py
@@ -112,7 +112,7 @@ for line in sys.stdin:
             "method": "turn/completed",
             "params": {
                 "threadId": "thread-skillsbench",
-                "turn": {"id": "turn-event-skillsbench", "status": "completed"},
+                "turn": {"id": "turn-event-skillsbench"},
             },
         }), flush=True)
         continue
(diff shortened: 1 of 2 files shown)


huangruiteng:
Findings: none.

- Motivation: app-server Goal traces can emit `turn/completed` without an explicit status, while the driver previously kept the earlier `running`/`inProgress` status and made the SkillsBench worker finish guard classify the run as unclean/missing.
- Solved or not: this PR fixes the driver at the event-normalization layer by treating a missing status on `turn/completed` as `completed`, matching the existing `event_msg:turn_completed` behavior. The smoke now exercises that exact event shape.
- User/operator impact: completed app-server Goal turns should no longer become uncountable solely because the completion event omitted the status field.
- Main risk: low and localized to app-server Goal event normalization; it does not change scoring, task text handling, permissions, or benchmark submission behavior.
- Design judgment: this is the reusable driver-level seam, not a reducer-side exception.

Validation:
- `python3 examples/skillsbench-app-server-goal-worker-smoke.py`
- `python3 examples/codex-app-server-goal-driver-smoke.py`

Boundary: public-safe smoke only; no raw task text, trajectories, verifier output, credentials, or local benchmark logs added.

Merge decision: self-merge with owner-authorized benchmark runtime closeout bypass.

PR #1580, opened 2026-07-06
Route summarized evidence back into the auto-research ring

huangruiteng:
## Summary\n- add the next-hypothesis successor to the initial evaluator summarize_evidence path\n- make the worker-turn smoke cover the real seeded summarize_evidence action, not only later evaluation-summary actions\n\n## Why\nThe visible KNN run now completes one real dev+holdout improvement, but the initial evaluator seed action can end in promotion review instead of returning to the first role for the next research round. This keeps the ring lightweight: last role -> curator review -> hypothesis proposer, without adding automatic replan or heavier auto-research code.\n\n## Validation\n- python3 examples/auto-research-worker-turn-smoke.py\n- python3 examples/auto-research-knn-continuation-smoke.py\n- python3 examples/auto-research-knn-preset-start-smoke.py\n- loopx canary premerge --from-git-diff\n\n## Boundary\n- role successor metadata and smoke only\n- no fake metrics, no pre-tick, no heartbeat prompt changes\n

### DIFF
--- a/examples/auto-research-worker-turn-smoke.py
+++ b/examples/auto-research-worker-turn-smoke.py
@@ -300,11 +300,11 @@ def main() -> int:
             goal_id=GOAL_ID,
             role="agent",
             text=(
-                "[P0-auto-research-live] Summarize held-out validation and "
+                "[P0-auto-research-live] Summarize visible held-out evidence and "
                 "open the next research round."
             ),
             task_class="advancement_task",
-            action_kind="write_evaluation_summary",
+            action_kind="summarize_evidence",
             claimed_by=EVALUATOR_AGENT_ID,
             dry_run=False,
         )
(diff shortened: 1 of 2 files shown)


huangruiteng:
Self-merge review note:\n\nChanged surfaces:\n- auto-research role successor metadata\n- worker-turn smoke coverage for the seeded evaluator summarize_evidence path\n\nWhy this is enough:\n- The observed visible run completed one real dev+holdout improvement but the initial evaluator path could stop at promotion review instead of routing back to the first role for a second research round.\n- The fix adds a role-declared successor for summarize_evidence when there is a validated candidate and holdout_improvement_count < 2. That preserves the lightweight ring model: evaluator -> curator -> hypothesis, without introducing automatic replan, fake ticks, or heavier user/auto-research logic.\n\nValidation run locally:\n- python3 examples/auto-research-worker-turn-smoke.py\n- python3 examples/auto-research-knn-continuation-smoke.py\n- python3 examples/auto-research-knn-preset-start-smoke.py\n- loopx canary premerge --from-git-diff: passed, merge_gate_passed=true, self_merge_allowed=true\n\nBoundary scan:\n- no raw logs, private paths, credentials, local artifacts, benchmark scoring, or task semantics changed\n

PR #1750, opened 2026-07-09
Keep benchmark standard seeds lane-local

huangruiteng:
## Summary

- treat the local run-ledger seed as an already accepted target-lane ledger
- remove redundant historical run-group filtering from standard aggregate refresh
- retain the target-lane label and add a dry-run regression assertion

## Why

A dedicated accepted-lane seed is the stable source of truth for a long-running benchmark. Reapplying broad historical run-group filters can exclude accepted rows whose legacy group labels differ, recreating count drift after a correct seed has already been selected. Future batches append only their compact closeouts to this lane-local ledger.

## Validation

- SkillsBench egress/launcher, reverse-tunnel closeout, and current-ledger aggregate smokes pass
- risk-based premerge canary: 7/7 checks pass, public boundary clean, 0 failures

This is a launcher cleanup only. It does not change scoring, task semantics, reducer behavior, permissions, submissions, or launch jobs.

### DIFF
--- a/examples/skillsbench-benchmark-egress-policy-smoke.py
+++ b/examples/skillsbench-benchmark-egress-policy-smoke.py
@@ -172,6 +172,7 @@ def test_public_launcher_batches_three_cases_with_closeout_sync() -> None:
             "SKILLSBENCH_EXPECTED_LOOPX_GIT_HEAD": "abc1234",
             "SKILLSBENCH_DOCKER_PROXY_HOST": "host.docker.internal",
             "SKILLSBENCH_RUN_STAMP": "20260710T000000CST",
+            "SKILLSBENCH_CANONICAL_CASE_IDS_FILE": "/opaque/canonical-ids.txt",
         }
     )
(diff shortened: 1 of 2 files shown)


huangruiteng:
## Maintainer pre-merge review

**Changed surfaces:** SkillsBench public launcher seed contract and one focused dry-run smoke.

**Checks:** launcher/egress smoke, reverse-tunnel closeout smoke, current-ledger aggregate smoke, diff checks, changed Python compile, 6 catalog canaries, and public boundary scan all passed. Failures/skips: none.

**Manual hold:** expected `benchmark_sensitive` hold. Owner authorization covers benchmark self-merges. This patch removes redundant filtering and does not change scoring, task semantics, reducers, permissions, submissions, or launch jobs.

**Decision:** authorized admin self-merge. The dedicated lane seed is already audited; future compact rows are appended to that ledger, so historical group-name filtering is both redundant and capable of excluding accepted legacy rows.

PR #1753, opened 2026-07-09
Keep issue-fix domain receipts truthful

huangruiteng:
## Summary
- make issue-fix domain-state writers persist `write_performed=true` in the decision row when the atomic upsert succeeds
- keep the local `write_result` receipt in the CLI response only, not in the compact JSONL decision row
- centralize the rule for feasibility and PR-lifecycle writers and remove post-write CLI mutation

## Root cause
The CLI persisted the packet while `domain_state_projection.write_performed` was still `false`, then changed only the in-memory response to `true` after the upsert. The ledger and command response therefore contradicted each other.

## Validation
- `python3 examples/issue-fix-feasibility-smoke.py`
- `loopx canary premerge --from-git-diff`: 6/6 selected checks passed; merge gate and self-merge eligibility true
- changed-path public boundary scan: 0 errors, 0 warnings

## Architecture
The domain-pack writer now owns receipt consistency for every caller. It prepares the truthful projection before the atomic write, restores the in-memory flag if the write raises, and attaches the operational receipt only after success. No schema version, raw path, or external-write behavior changes.

### DIFF
--- a/examples/issue-fix-feasibility-smoke.py
+++ b/examples/issue-fix-feasibility-smoke.py
@@ -154,6 +154,10 @@ def main() -> int:
         assert ledger.exists(), ledger
         rows = ledger.read_text(encoding="utf-8").splitlines()
         assert len(rows) == 1, rows
+        first_row = json.loads(rows[0])
+        first_projection = first_row["domain_state_projection"]
+        assert first_projection["write_performed"] is True, first_row
+        assert "write_result" not in first_projection, first_row
 
         second_command = command.copy()
         second_command[second_command.index("confirmed")] = "missing"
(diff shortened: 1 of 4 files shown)


huangruiteng:
Self-review complete: no actionable findings.

**Product/architecture judgment**
- Motivation: eliminate a contradictory write receipt that could mislead replanning and ledger consumers.
- Solved: persisted decision rows and CLI responses now agree that the local upsert occurred, while only the response carries the operational receipt.
- User/operator impact: agents can trust recent issue-fix domain evidence without inspecting implementation order or guessing which surface is authoritative.
- Main risk: changing both feasibility and PR-lifecycle persistence semantics. Both end-to-end smokes now assert response and persisted-row behavior.
- Design: receipt consistency belongs to the domain-pack writer, not duplicated CLI branches.

**Validation**
- focused feasibility and PR-lifecycle smokes passed
- workflow-contract, intake, line-budget, compile, diff, and boundary checks passed
- risk premerge gate passed 6/6 with no holds and `self_merge_allowed=true`

**Merge decision**
Small, single-purpose, validated control-plane repair; eligible for side-agent self-merge under the active contract.

PR #1769, opened 2026-07-10
Canonicalize agent vision lifecycle states

huangruiteng:
## Summary
- centralize agent-vision lifecycle token normalization and closed-state detection in a cycle-free goals module
- canonicalize closure aliases such as `vision_satisfied` and `closed_no_followup` while preserving custom states as open
- reject malformed prose states at the refresh-state write boundary and document the lifecycle contract
- cover both new writes and legacy quota/status reads

## Motivation
A bounded closeout written as `vision_satisfied` passed refresh-state validation but quota treated it as an open acceptance gap, causing another autonomous replan. Closure semantics lived only as projection-side magic strings.

## Architecture
This does not add a protocol field or a strict enum. Existing domain-specific states remain extensible and open; only explicit closure aliases normalize to canonical states. Write and read paths share one lightweight predicate.

## Validation
- focused goal-vision refresh-state and quota decision-plane smokes
- goal-vision protocol, autonomous-replan, compaction, state-projection, side-agent state-machine, and integrated control-plane smokes
- `loopx canary premerge --from-git-diff`: 17/17 passed, zero failures, warnings, or manual holds
- public/private boundary scan: 7 changed files clean

## Risk
Low compatibility risk: legacy closure aliases become correctly terminal; unknown custom states intentionally remain open. No benchmark behavior, permissions, credentials, private state, or raw evidence is changed.

### DIFF
--- a/docs/reference/protocols/goal-vision-replan-contract-v0.md
+++ b/docs/reference/protocols/goal-vision-replan-contract-v0.md
@@ -77,6 +77,14 @@ the same `agent_id` as the refresh run. This keeps `research-executor`,
 `evaluator-promoter`, and other roles from overwriting or satisfying each
 other's active vision.
 
+`state` is a lower `snake_case` lifecycle token. Domain-specific states remain
+extensible and are treated as open. The write path canonicalizes closure aliases
+such as `closed`, `satisfied`, and `vision_satisfied` to `vision_closed`, and
+`closed_no_followup` to `no_followup`. Quota/status use the same centralized
+closure predicate when reading older persisted packets, so a legacy alias cannot
+silently reopen a satisfied vision. Prose or malformed state values fail at the
+write boundary with an actionable error.
+
 When a valid packet includes `replan_trigger_summary`, status/quota projects it
 as `goal_frontier_projection.acceptance_gaps[]`. If no runnable advancement
 frontier remains, that gap is evaluated before monitor quiet skip and can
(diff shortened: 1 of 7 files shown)


huangruiteng:
## Self-review

### Findings
No blocking findings. The diff keeps custom/domain vision states open, canonicalizes only explicit closure aliases, and uses the same predicate for new writes and legacy reads.

### Assumptions
- `completed_current_slice` and other unrecognized states must remain open because slice completion is not whole-vision acceptance evidence.
- Existing closure aliases may safely normalize to canonical stored states; no consumer is documented to require the alias spelling itself.

- **Motivation:** prevent an apparently satisfied agent vision from silently reopening as an autonomous replan gap.
- **Solved:** write-path canonicalization plus read-path legacy handling covers both future and persisted packets.
- **Operator impact:** agents no longer need to memorize projection-side magic strings to close a vision. Malformed prose fails at the write boundary.
- **Main risk:** accidentally closing custom intermediate states. The implementation avoids that by preserving unknown tokens as open, with a regression for `completed_current_slice`.
- **Design:** a 51-line pure goals module is the smallest cycle-free shared dependency; no schema field, strict global enum, or new state-machine branch was added.

### Validation
- focused refresh-state write-path and quota projection smokes passed after rebase
- protocol, autonomous-replan, compaction, state-projection, side-agent, and integrated control-plane smokes passed
- premerge gate: 17/17 passed; 0 failures, 0 warnings, 0 manual holds
- public/private boundary: 7 changed files clean
- no benchmark behavior, permission boundary, private state, raw evidence, or credentials changed

### Merge decision
Self-merge approved under the owner standing authorization. Coverage is sufficient because it exercises both protocol boundaries changed here, the composed control-plane path, the canary-selected risk profiles, and the public boundary.

PR #1772, opened 2026-07-10
Make benchmark ledger catch-up logically idempotent

huangruiteng:
## Summary
- expose the stable ledger signature already used by merge/drift checks
- treat either matching run id or matching logical signature as an existing catch-up row
- cover equivalent results from different run groups in the generic closeout smoke

## Root cause
A ledger logical-backfill merge can intentionally preserve the existing run id. Run-id-only catch-up therefore retried one equivalent compact on every closeout even though the ledger already represented it.

## Boundaries
- no scoring, case semantics, runner launch, or aggregate-selection changes
- no benchmark jobs launched
- compact/public evidence only

## Validation
- focused closeout, ledger, and supervisor smokes passed
- real public-root canary against a temporary copy of the live ledger: `40/40` historical candidates recognized as already present, `catchup_upserted_count=0`; aggregate preserved `87/87`, sum `28.525`, mean `0.327874`
- standard premerge canary: 7/7 executed checks passed, 0 failures
- public boundary scan: 0 errors, 0 warnings

### DIFF
--- a/examples/benchmark-remote-closeout-catchup-smoke.py
+++ b/examples/benchmark-remote-closeout-catchup-smoke.py
@@ -16,6 +16,7 @@ if str(REPO_ROOT) not in sys.path:
     sys.path.insert(0, str(REPO_ROOT))
 
 from loopx.benchmark_core.remote_closeout import closeout_remote_benchmark_batch
+from loopx.benchmark_ledger import update_benchmark_run_ledger
 
 
 def _compact(case_id: str, score: float) -> dict[str, object]:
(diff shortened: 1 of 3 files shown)


huangruiteng:
## Maintainer review

**Findings:** no blocking issues found.

**Root cause verified:** a compact can be logically merged into an existing ledger row while preserving the existing row's run id. The first implementation compared only run ids, so one equivalent `fix-visual-stability` compact remained a perpetual catch-up candidate. The stable signature already used by ledger merge/drift is the correct shared identity contract.

**Product / architecture judgment:** this is the smallest durable repair. It reuses one ledger-owned identity function in closeout instead of inventing a second equivalence rule, and the smoke now covers different run groups with the same logical result. No scoring, selection, runner, or task behavior changes.

**Validation:** focused smokes passed; standard premerge canary executed 7 checks with 0 failures; public boundary scan clean. A real public-root canary against a temporary live-ledger copy recognized all 40 historical candidates as present with `catchup_upserted_count=0`, preserving `87/87`, sum `28.525`, mean `0.327874`.

**Manual hold:** benchmark-sensitive policy requires maintainer review; this comment supplies it under the owner's standing benchmark self-merge authorization.

**Merge decision:** approved for self-merge.

PR #1776, opened 2026-07-10
Isolate release manifest imports

huangruiteng:
## Summary
- generate `release.json` from inside the copied release snapshot with an isolated `PYTHONPATH`
- extend the existing no-clone release gate to install from a deliberately stale LoopX checkout
- require manifest version/tag to match the installed archive runtime

## Root cause
`install-local.sh` prepended the new release snapshot to `PYTHONPATH`, but `python -m` still searched the caller current working directory first. Running `loopx update --execute` from an older LoopX checkout could therefore write that checkout's version into the new archive install manifest. Runtime code was current, so `loopx doctor` correctly reported a manifest/runtime mismatch.

## Validation
- no-clone release verification, including stale-checkout regression: passed
- local installer smoke: passed
- packaged install smoke: passed
- fresh-clone quickstart: passed
- release version contract: passed
- update smoke: passed
- bash syntax and Python compile checks: passed
- standard premerge canary: 17 selected + 4 direct checks, 0 failures/warnings/holds, self-merge allowed
- public/private scan and `git diff --check`: clean

## Scope
No installer flags, release-manifest schema, update semantics, permissions, or rollback behavior change. The patch only makes the manifest generator load the snapshot it is describing.

### DIFF
--- a/examples/release/codex-cli-no-clone-release-verification-smoke.py
+++ b/examples/release/codex-cli-no-clone-release-verification-smoke.py
@@ -5,6 +5,8 @@ from __future__ import annotations
 
 import json
 import os
+import re
+import shutil
 import subprocess
 import tarfile
 import tempfile
@@ -26,6 +28,17 @@ FIRST_RUN_COMMANDS = (
 
 def normalize(text: str) -> str:
     return " ".join(text.split())
+
+
+def package_version(root: Path) -> str:
+    match = re.search(
+        r'^__version__\s*=\s*"([^"]+)"$',
+        (root / "loopx" / "__init__.py").read_text(encoding="utf-8"),
+        re.MULTILINE,
+    )
+    if not match:
+        raise AssertionError(f"missing LoopX version in {root}")
+    return match.group(1)
 
 
 def run(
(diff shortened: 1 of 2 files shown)


huangruiteng:
Self-review / premerge evidence

- Changed surfaces: release snapshot manifest generation and the existing no-clone release verifier. No CLI option, schema, permission, scheduler, benchmark, or rollback contract changes.
- Regression shape: verifier runs the archive installer from a fake stale checkout whose package advertises `9.9.9`; installed manifest and runtime must still match the archive version. This fails on the v0.1.15 installer and passes with this patch.
- Focused checks: no-clone release verification, install-local, packaged install, fresh-clone quickstart, release version contract, update smoke, bash syntax, Python compile, and diff check passed.
- Risk-based canary: public-entry/install/release profile selected 17 checks plus 4 direct checks; 0 failures, warnings, or manual holds; `self_merge_allowed=true`.
- Boundary hygiene: no private state, local paths, credentials, raw logs, or generated evidence in the diff.

Coverage is sufficient because the fix changes exactly one Python invocation boundary and the regression proves the previously failing caller-cwd condition through the supported no-clone installer.

PR #1789, opened 2026-07-10
Fix refresh-state validation atomicity

huangruiteng:
## Summary
- validate bounded vision unchanged reasons before any active-state mutation
- preserve the existing vision checkpoint normalization contract
- add a real non-dry-run regression proving rejected refreshes leave state unchanged

## Validation
- goal-vision-refresh-state-budget-smoke.py
- next-action-projection-contract-smoke.py
- refresh-state-agent-lane-scope-smoke.py
- python compileall
- loopx check
- loopx canary premerge --from-git-diff (17/17 passed, no manual holds)

### DIFF
--- a/examples/project/goal-vision-refresh-state-budget-smoke.py
+++ b/examples/project/goal-vision-refresh-state-budget-smoke.py
@@ -153,7 +153,7 @@ def write_json(path: Path, value: dict) -> None:
 def main() -> int:
     with tempfile.TemporaryDirectory(prefix="loopx-goal-vision-budget-") as tmp:
         root = Path(tmp)
-        registry_path, runtime, _project = write_fixture(root)
+        registry_path, runtime, project = write_fixture(root)
         valid_path = root / "valid-vision.json"
         invalid_path = root / "invalid-vision.json"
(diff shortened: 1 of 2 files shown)


huangruiteng:
No blocking findings.

Assumption: the standing owner authorization for benchmark-related self-merge covers this benchmark-triggered control-plane repair. The change does not alter benchmark scoring, task semantics, permissions, publication, or submission behavior.

Product/architecture judgment: refresh-state is a state-write boundary, so rejecting an input after mutating ACTIVE_GOAL_STATE breaks LoopX source-of-truth guarantees. The fix keeps the existing vision normalization contract and moves its validation ahead of the write lock; the non-dry-run regression proves a rejected command leaves the state byte-for-byte unchanged. This is the smallest reusable repair for the observed bad case.

Validation: focused goal-vision budget/atomicity smoke; next-action projection smoke; agent-lane refresh smoke; compileall; diff checks; loopx check; premerge canary 17/17 passed with no manual holds; public/private boundary clean.

Decision: approved for owner-authorized admin squash self-merge.

PR #1795, opened 2026-07-10
Make deferred capacity resumes actionable

### LINKED ISSUE #1792: Make deferred todos actionable with typed capacity resume conditions
## Summary

LoopX has read/projection support for deferred todos and resume conditions, but the CLI cannot currently create an actionable deferred agent todo. At the same time, resource-capacity resume predicates are accepted syntactically and only become visibly unsupported during projection.

This leaves resource-constrained schedulers without a safe way to record work that should remain non-runnable until capacity becomes available.

## Observable behavior

On LoopX `v0.1.16` at public commit `f3a67155430a`:

```bash
loopx todo add \
  --goal-id demo-capacity \
  --role agent \
  --task-class advancement_task \
  --text "Run candidate when a short worker is available" \
  --status deferred \
  --resume-when capacity_available:short_pool

loopx todo list --goal-id demo-capacity
```

Actual behavior:

- `--status deferred` is accepted by the parser but the added todo is persisted as `open`.
- The resume condition is persisted, then projected with `unsupported=true` and `satisfied=false`.
- Trying `todo update --status deferred` for an agent todo is rejected by the terminal-completion guard, while `todo complete` can only write `done`.

The result is that the read model understands deferred todos, but the supported CLI write path cannot create one. A caller may also believe work is deferred when it is actually open and eligible for normal todo/quota routing.

## Expected behavior

1. `todo add --status deferred` either persists a deferred todo end to end or fails immediately with an actionable error. It must not silently create an open todo.
2. There is a supported transition for an existing agent todo to deferred state, either through `todo update --status deferred` or a dedicated `todo defer` command.
3. Unsupported `resume_when` kinds fail closed at write time instead of producing an inert condition later in projection.
4. LoopX provides a typed, non-project-specific capacity condition such as `resource_available:<resource_key>` or `capacity_available:<resource_key>` through a registered resolver/provider contract.

## Suggested acceptance contract

- Add -> defer -> unsatisfied resume condition keeps the todo out of executable and quota-consuming frontiers.
- When the registered capacity resolver reports availability, the deferred todo becomes resume-ready exactly once and re-enters normal claim/routing.
- Unknown condition kinds are rejected before state mutation.
- Repeated polls and resume evaluation are idempotent.
- Capacity conditions never evaluate arbitrary user strings, shell commands, or unregistered callbacks.
- Add an end-to-end smoke covering create, defer, unsatisfied projection, capacity transition, resume, and quota eligibility.

## Suspected product surfaces

This is a hypothesis based on the public source layout:

- `loopx/cli_commands/todo.py`: `--status` is exposed globally, but the add branch does not pass it to `add_goal_todo`.
- `loopx/todos.py`: new todo metadata is initialized with `status=open`; the agent terminal-transition guard also treats deferred like completed work.
- `loopx/control_plane/todos/contract.py`: the supported resume kinds are currently limited to `todo_done` and `pr_merged`.
- `loopx/control_plane/todos/todo_summary.py`: other kinds are marked unsupported during projection.

## Why this is general

The same lifecycle occurs with GPU pools, benchmark workers, CI runners, API rate limits, and other bounded external resources. A typed capacity contract would let domain adapters supply availability without embedding one project's scheduler semantics in LoopX core.

huangruiteng:
## Summary
- persist `todo add --status deferred` and allow `todo update --status deferred`
- reject unsupported resume condition kinds before state mutation
- resolve `capacity_available:<key>` from quota runtime capability declarations
- add a focused idempotent lifecycle and routing smoke

## Validation
- `python3 examples/control_plane/todo-deferred-capacity-cli-smoke.py`
- `python3 examples/control_plane/todo-contract-smoke.py`
- `python3 examples/control_plane/todo-cli-smoke.py`
- `python3 examples/control_plane/todo-lifecycle-cli-smoke.py`
- `python3 examples/control_plane/todo-deferred-resume-lanes-smoke.py`
- `python3 examples/control_plane/quota-replan-decision-plane-smoke.py`
- standard `canary premerge --from-git-diff` (all selected checks passed)

Closes #1792

### DIFF
--- a/examples/control_plane/todo-deferred-capacity-cli-smoke.py
+++ b/examples/control_plane/todo-deferred-capacity-cli-smoke.py
@@ -1,0 +1,210 @@
+#!/usr/bin/env python3
+"""Smoke-test deferred todo writes and runtime capacity resume routing."""
+
+from __future__ import annotations
+
+import sys
+import tempfile
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+SMOKE_DIR = Path(__file__).resolve().parent
+if str(SMOKE_DIR) not in sys.path:
+    sys.path.insert(0, str(SMOKE_DIR))
+if str(REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(REPO_ROOT))
+
+from todo_lifecycle_fixtures import (  # noqa: E402
+    GOAL_ID,
+    parsed_agent_summary,
+    parsed_items,
+    run_cli,
+    run_cli_error,
+    write_fixture,
+)
+
+
+CAPACITY_RESUME = "capacity_available:short_pool"
+
+
+def remove_legacy_placeholder(state_file: Path) -> None:
+    state_file.write_text(
+        state_file.read_text(encoding="utf-8").replace(
+            "- [ ] Legacy monitor-only placeholder.\n",
+            "",
+        ),
+        encoding="utf-8",
+    )
+
+
+def assert_deferred_add_and_capacity_resume() -> None:
+    with tempfile.TemporaryDirectory(prefix="loopx-deferred-capacity-add-") as tmp:
+        registry_path, state_file = write_fixture(Path(tmp))
+        remove_legacy_placeholder(state_file)
+        before = state_file.read_text(encoding="utf-8")
+
+        unsupported = run_cli_error(
+            registry_path,
+            "todo",
+            "add",
+            "--goal-id",
+            GOAL_ID,
+            "--role",
+            "agent",
+            "--text",
+            "Wait for an unregistered external predicate.",
+            "--status",
+            "deferred",
+            "--resume-when",
+            "external_signal:short_pool",
+        )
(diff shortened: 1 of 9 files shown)


huangruiteng:
Maintainer self-review

**Findings:** no blocking issues found. The write path now distinguishes `deferred` from completed work, rejects unknown resume kinds before mutation, and keeps capacity evaluation projection-only and idempotent.

**Product / architecture judgment:** this closes the mismatch where the read model understood deferred todos but the CLI could only create runnable open work. Reusing quota runtime capability declarations as the capacity provider keeps the contract generic and avoids a second callback registry or project-specific scheduler API. Main risk is hot-path todo/quota routing; the focused smoke proves absent/present capacity, repeated polls, and no durable-state mutation.

**Validation:** deferred-capacity CLI smoke; todo contract, CLI, lifecycle, deferred-lane, and quota-replan smokes; Python compile; public/private boundary scan; standard `canary premerge --from-git-diff` with all selected checks passing. GitHub reports no remote checks for this branch.

**Merge decision:** self-merge authorized by the owner; single-purpose runtime fix with focused public regression coverage.

PR #1798, opened 2026-07-10
Route benchmark proxy through rootless host interface

huangruiteng:
## Summary
- detect rootless Docker after daemon API negotiation
- keep docker0 for normal daemons, but use the host interface reachable through rootless NAT for benchmark build proxy egress
- restrict the rootless forwarder to the NAT source address observed from local containers, avoiding a LAN-wide open proxy
- preserve explicit proxy-host overrides

## Root cause
The benchmark proxy and Dockerfile env patch were both enabled, but the remote daemon is rootless. Build containers cannot reach host loopback, docker0, magic hostnames, or host-gateway; they can reach the host external interface and arrive NATed as that same local interface address. The previous host-only preflight therefore produced a false positive while both builds timed out in APT.

## Validation
- container-to-host probes for loopback/docker0/magic/rootless/explicit gateway all failed as expected
- host external interface probe succeeded and confirmed the NAT peer category is the local host interface
- launcher remote dry-run selected `rootless_host_interface` and daemon API `1.43`
- launcher focused smoke and `bash -n` passed
- six selected benchmark/core canaries and public-boundary scan passed

## Holds and boundaries
- generic `benchmark_sensitive` manual hold remains; owner explicitly authorized benchmark self-merge
- no task, scoring, verifier, submission, leaderboard, or permission semantics change
- no private addresses, credentials, task content, raw logs, or generated artifacts are committed

### DIFF
--- a/examples/skillsbench-benchmark-egress-policy-smoke.py
+++ b/examples/skillsbench-benchmark-egress-policy-smoke.py
@@ -229,10 +229,14 @@ def test_public_launcher_applies_ssh_options_to_remote_discovery() -> None:
         fake_ssh.write_text(
             "#!/usr/bin/env bash\n"
             "printf '%s\\n' \"$*\" >> \"$SKILLSBENCH_TEST_SSH_LOG\"\n"
-            "if [[ \"$*\" == *'ip -4 addr show docker0'* ]]; then\n"
-            "  printf 'docker-bridge.example.invalid\\n'\n"
-            "else\n"
+            "if [[ \"$*\" == *'/version'* ]]; then\n"
             "  printf '1.43\\n'\n"
+            "elif [[ \"$*\" == *'docker info'* ]]; then\n"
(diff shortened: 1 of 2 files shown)


huangruiteng:
Maintainer self-review packet:

Changed surfaces: public SkillsBench launcher endpoint discovery/forwarder admission and its focused dry-run smoke. Checks: `bash -n`, focused launcher smoke, remote rootless-mode dry-run, five negative container reachability probes, one successful external-interface probe with NAT peer classification, premerge diff/compile checks, six benchmark/core canaries, and public-boundary scan. All executed checks passed; no checks were skipped.

Manual hold: premerge reports the generic `benchmark_sensitive` hold. Owner authorization permits benchmark self-merge. The change repairs proxy transport only and keeps explicit overrides; it does not alter task/scoring/verifier/submission semantics. Peer filtering is necessary because the forwarder now binds an interface visible on the host LAN, and the validated rootless container source is the host interface itself.

PR #1803, opened 2026-07-10
Fail fast when the Codex sandbox cannot execute

huangruiteng:
## Summary
- exercise the selected host-local Codex sandbox during runner preflight, not only `codex --version`
- emit public-safe readiness and precise setup attribution when the sandbox runtime cannot execute
- preserve an explicit no-probe path for `danger-full-access`, where Codex sandboxing is intentionally disabled

## Validation
- `uv run --extra test pytest -q tests/test_skillsbench_remote_codex_preflight.py` (5 passed)
- `python examples/skillsbench-host-local-launch-plan-smoke.py`
- `python examples/skillsbench-benchmark-run-smoke.py`
- `python examples/skillsbench-runner-core-contract-smoke.py`
- `loopx canary premerge --from-git-diff` (all 6 selected catalog canaries passed; expected benchmark-sensitive maintainer hold)

## Boundary
- no scoring, task-semantic, submission, permission, or benchmark launch behavior changes
- no raw task text, trajectories, verifier output, credentials, private paths, or generated run artifacts included

### DIFF
--- a/loopx/benchmark_adapters/skillsbench_codex_runtime.py
+++ b/loopx/benchmark_adapters/skillsbench_codex_runtime.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import shutil
 import subprocess
 import tempfile
 from collections.abc import Callable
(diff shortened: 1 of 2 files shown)


huangruiteng:
No blocking findings.

**Assumptions**
- `danger-full-access` is an explicit operator choice; the preflight records that the Codex sandbox probe is not required rather than claiming it ran.
- non-bypass modes must prove the exact selected `sandbox_mode` can execute a fixed no-op before any benchmark setup.

- Motivation: a partial Codex install or unsupported Linux sandbox can pass `--version` and TUI readiness, then fail every first tool call after a case has already started.
- Solved: the reusable runtime preflight now exercises the selected sandbox profile and attributes failure as runner setup, preventing false solution failures and wasted case time.
- Operator impact: unsupported hosts fail in seconds with public-safe diagnostics; explicit bypass remains visible.
- Main risk: compatibility with Codex versions that do not expose the `sandbox` subcommand. That is intentionally treated as unavailable runtime capability for non-bypass modes.
- Design: this belongs in the bounded `skillsbench_codex_runtime` preflight and reuses existing prerequisite reduction; it does not add benchmark-specific scoring or task logic.

**Validation**
- focused pytest: 5 passed
- SkillsBench host-local launch-plan, benchmark-run, and runner-core smokes passed
- exact remote runtime check: `workspace-write` fails fast with `codex_cli_sandbox_probe_failed`; explicit `danger-full-access` reports ready without a false sandbox claim
- `loopx canary premerge --from-git-diff`: diff checks, Python compile, line-budget, adapter readiness, benchmark core, artifact boundary, run-ledger, and case-analysis all passed (6/6 catalog canaries)
- public/private scan clean; no task text, trajectories, verifier output, credentials, local paths, or generated artifacts
- expected manual hold: `benchmark_sensitive`; satisfied by explicit owner authorization for benchmark helper/runtime self-merge

**Merge decision**: self-merge approved. Scope is runner prerequisite validation only; no scoring, task semantics, submission, permission-boundary, or benchmark-launch changes.

PR #1804, opened 2026-07-10
Add focused Explore graph exports

huangruiteng:
## Summary

- add read-only Explore graph filtering by repeated node status and exact tag
- keep matching nodes' ancestors by default so focused graphs retain topology context
- expose filter metadata and graph-local counts for Mermaid and JSON exports
- document executive-view usage and extend the durable result-layer smoke

## Validation

- source-checkout smoke against a live public-safe result projection
- `loopx canary premerge --from-git-diff` (17 selected checks passed)
- `git diff --check`
- public/private boundary scan on all four changed paths

## Boundary

This is a read-only projection change. It does not mutate Explore state, Lark tables, todo planning, quota, spawn policy, or execution authority.

### DIFF
--- a/docs/capabilities/explore/README.md
+++ b/docs/capabilities/explore/README.md
@@ -65,6 +65,27 @@ parent/`subtopic_of` topology tree, and Mermaid flowchart source.
 `loopx explore graph --graph-format mermaid|json [--out <file>]` exports the
 topology for a Feishu doc, whiteboard, or any Mermaid renderer.
 
+For an executive view, keep the append-only result log as the single source
+and export a focused graph instead of maintaining a second hand-written
+Mermaid file:
+
+```bash
+loopx explore graph \
+  --goal-id <id> \
+  --status exploring \
+  --status blocked \
+  --tag executive \
+  --graph-format mermaid \
+  --out explore-executive.mmd
+```
+
+Repeated statuses match any requested status, repeated tags match any exact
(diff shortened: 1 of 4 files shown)


huangruiteng:
## Maintainer review

**Findings:** No blocking issues found. The no-filter path preserves the existing full graph output, filters are read-only, and ancestor expansion is cycle-bounded.

**Open assumptions:** Tags are exact public-safe projection tags; substring or expression matching is intentionally out of scope. `counts` remains the full result-projection count for compatibility, while `graph_counts` describes the filtered export.

**Product/architecture judgment:** The operator problem is real: hand-maintained executive Mermaid files can drift from the append-only Explore topology. A pure graph-view helper plus CLI-only filters solves that problem without creating a second state source or expanding planner/runtime authority. Status/tag selection with optional ancestor context is a reusable projection contract rather than a project-specific renderer. Main risk is CLI compatibility; the default no-filter behavior and additive metadata keep that risk low.

**Validation:** `python3 examples/explore-result-layer-smoke.py`; source-checkout filter smoke against a live public-safe projection; `loopx canary premerge --from-git-diff` with all 17 selected checks passing; `git diff --check`; candidate-path and rendered-PR public/private boundary scans.

**Merge decision:** Self-merge approved under explicit owner authorization. The PR is narrow, single-purpose, public-safe, and does not change execution, permission, quota, spawn, or Lark mutation behavior.

PR #1808, opened 2026-07-10
fix(lark): run idempotent issue-fix field migrations before views

huangruiteng:
## Motivation

Live validation of the issue-fix Kanban link and stage migration exposed a product-path gap. Existing-board setup configured views before issue-fix fields. A platform rejection from an already-configured view caused partial return before field migration, while unconditional same-state field updates could themselves fail as no-op writes.

## Implementation

- compare the existing field-list read model with desired Issue, Pull Request, and Stage definitions;
- emit field updates only for real type/style/option-order drift;
- address fields by their resolved field ids;
- run migrations immediately after field discovery and before any view enrichment;
- report updated_fields separately from created_fields;
- cover drift, idempotent no-op, resolved ids, and migration-before-view ordering in the generic Lark smoke.

## Validation

- Python compile for the touched sink modules: passed
- issue-fix-outcome-projection-smoke: passed
- lark-kanban-control-plane-smoke: passed
- premerge canary: direct diff checks, Python compile, 5 selected catalog canaries, and public-boundary scan passed
- git diff check: passed

## Risk and gaps

This changes setup ordering for existing boards only. New boards already receive the desired field definitions at table creation. Field mutations remain same-type updates and are skipped when the live field list already matches. The view-enrichment platform limitation remains isolated as a partial setup warning and no longer prevents field migration.

### DIFF
--- a/examples/lark-kanban-control-plane-smoke.py
+++ b/examples/lark-kanban-control-plane-smoke.py
@@ -173,7 +173,46 @@ def existing_setup_runner(args: list[str], cwd: Path | None, timeout: float | No
     if args[-1:] == ["--help"]:
         return {"returncode": 0, "stdout": "help\n", "stderr": "", "timed_out": False}
     if "+field-list" in args:
-        return {"returncode": 0, "stdout": json.dumps({"ok": True, "data": {"fields": [{"name": "Task"}, {"name": "Status"}, {"name": "Claim"}, {"name": "Priority"}]}}), "stderr": "", "timed_out": False}
+        return {
+            "returncode": 0,
+            "stdout": json.dumps(
+                {
+                    "ok": True,
+                    "data": {
+                        "fields": [
+                            {"name": "Task"},
+                            {"name": "Status"},
+                            {"name": "Claim"},
+                            {"name": "Priority"},
+                            {
+                                "id": "fld_issue_existing",
+                                "name": "Issue",
+                                "type": "text",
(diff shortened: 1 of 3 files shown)


huangruiteng:
## Self-review

### Findings

No blocking findings.

The live bad case showed that a correct desired schema is insufficient when migration sits behind best-effort view enrichment. This patch moves drift detection and field mutation to the existing-board discovery seam, uses resolved field ids, and skips no-op writes from the actual field-list read model. That is the lowest reusable layer and avoids repository-specific setup branches. The main compatibility risk is mutating existing select option order; the comparison preserves all desired option names and only updates when the ordered name contract differs.

### Validation

- Python compile for LoopX package, CLI modules, and touched sink modules: passed.
- issue-fix-outcome-projection-smoke: passed.
- lark-kanban-control-plane-smoke: passed, including drift, idempotent no-op, resolved field ids, and migration-before-view ordering.
- premerge canary: direct diff checks, Python compile, 5 selected catalog canaries, and public-boundary scan passed.
- git diff check: passed.
- GitHub reported no CI runs for this PR.

### Boundary / hold review

No private state, credentials, absolute local paths, raw logs, permission changes, benchmark behavior, or project-specific logic are included. The patch is a narrow follow-up to the owner-authorized issue-fix Kanban capability and has live bad-case evidence.

### Decision

Self-merge with admin bypass, then refresh the installed release and rerun setup/readback against the real Base.

PR #1818, opened 2026-07-10
Fix verifier completion polling on silent Docker exec

huangruiteng:
## Summary
- read benchmark completion markers through `docker compose cp` when the sandbox exposes the Compose transport
- keep the existing exec-stdout reader as the generic fallback
- record the selected public-safe completion reader and cover both paths

## Why
Some remote Docker transports return from `compose exec` without preserving stdout. The verifier can finish and write an exit-status marker, while an stdout-based `cat` poll never observes it and waits until the outer timeout. Copying the marker to the runner host makes completion observation independent of exec output capture.

## Validation
- 8 focused pytest tests passed
- SkillsBench benchmark-run, runner-core, host-local launch, egress, UV-cache, benchmark-core, and Python line-budget smokes passed
- `loopx canary premerge --from-git-diff`: all 6 selected catalog canaries passed; benchmark-sensitive manual hold reviewed under owner authorization

## Scope
No scoring, task semantics, permission, submission, or benchmark launch behavior changes.

### DIFF
--- a/loopx/benchmark_core/__init__.py
+++ b/loopx/benchmark_core/__init__.py
@@ -31,6 +31,7 @@ from .artifacts import (
 )
 from .container_exec import (
     parse_container_exit_status,
+    read_container_file_via_compose_copy,
     run_container_command_with_exit_status,
     wrap_container_command_with_exit_status,
 )
@@ -191,6 +192,7 @@ __all__ = [
     "PreflightResult",
     "RoundReward",
     "RunHandle",
+    "read_container_file_via_compose_copy",
     "run_container_command_with_exit_status",
     "RUN_PERMISSION_POLICY_SCHEMA_VERSION",
     "RUN_PERMISSION_QUOTA_PROJECTION_SCHEMA_VERSION",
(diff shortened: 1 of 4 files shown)


huangruiteng:
Maintainer pre-merge review

Changed surfaces:
- shared benchmark container completion helper
- SkillsBench final-verifier completion polling
- focused regression tests for Compose-copy and exec-stdout readers

Validation:
- 8 focused pytest tests passed
- benchmark-core adapter, SkillsBench runner-core, host-local launch, egress policy, UV-cache, full benchmark-run, and Python line-budget smokes passed
- `loopx canary premerge --from-git-diff`: 6/6 selected catalog canaries passed; compile and diff checks passed
- staged public/private boundary scan found no credentials, private paths, raw logs/trajectories, verifier output, or internal links

Failures/skips:
- no validation failure
- canary reports the expected benchmark-sensitive manual hold; owner has explicitly authorized self-merge for benchmark helper/runtime changes
- no live benchmark is launched by this PR

Coverage is sufficient because the regression is the transport contract itself: the primary test reproduces empty exec stdout while a copied exit marker remains readable, and a second test preserves the non-Compose fallback. The diff does not alter scoring, task semantics, permissions, submissions, or solver behavior.

PR #1837, opened 2026-07-10
Fix live issue-fix packet timestamps

huangruiteng:
## Motivation

Live issue-fix CLI commands inherited dated fixture timestamps when callers omitted --generated-at. That made fresh PR lifecycle observations look stale and could corrupt ordering in compact domain state.

## Implementation

- Resolve one current UTC timestamp at the issue-fix CLI invocation boundary.
- Apply it consistently to workflow, feasibility, lifecycle, outcome, reviewer, fixture, and repository-memory commands.
- Preserve explicit --generated-at values for deterministic fixtures and tests.
- Add a lifecycle CLI regression covering both the live default and explicit override.

## Validation

- 11 focused issue-fix smoke scripts passed.
- Ruff check passed on both changed files.
- Python compile passed.
- Real read-only PR lifecycle fetch produced the current UTC invocation time and performed no write.
- loopx canary premerge --from-git-diff passed: 6 selected checks, 0 failures, 0 manual holds; public boundary clean.

## Risk and exclusions

The change is limited to CLI defaults; direct packet-builder defaults remain deterministic for library fixtures. No credentials, local paths, runtime state, raw logs, or private evidence are included.

### DIFF
--- a/examples/issue-fix-pr-lifecycle-smoke.py
+++ b/examples/issue-fix-pr-lifecycle-smoke.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+from datetime import datetime, timezone
 import json
 import re
 import subprocess
@@ -235,6 +236,7 @@ def main() -> int:
             ),
             encoding="utf-8",
         )
+        invoked_after = datetime.now(timezone.utc).replace(microsecond=0)
         cli_result = run_cli(
             [
                 "issue-fix",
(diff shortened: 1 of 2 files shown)


huangruiteng:
Findings: none.

Open questions or assumptions: none. The direct packet-builder defaults intentionally remain deterministic; only live CLI invocations receive the current UTC timestamp when no override is supplied.

Product and architecture judgment: this fixes a control-plane evidence-freshness bug at the correct boundary. A single invocation timestamp is resolved once in the issue-fix CLI and reused across all subcommands, avoiding command-specific clock drift while preserving explicit fixture overrides. The change is reusable rather than repository-specific. Main risk is compatibility for callers that accidentally relied on dated defaults; the explicit flag remains available and focused smokes cover both paths.

Validation: 11 focused issue-fix smokes passed; Ruff check and Python compile passed; a real read-only PR lifecycle fetch returned the current UTC timestamp with no write; premerge canary passed 6 selected checks with 0 failures and 0 manual holds; public/private boundary clean.

Merge decision: approved for owner-authorized self-merge.

PR #1840, opened 2026-07-10
Automate goal-default reviewer notifications

huangruiteng:
## Motivation

Secondary reviewer notification was implemented but still required manual `--notification-sinks-json` assembly on every call. Its receipts lived only in the caller's private input, so a restart could lose idempotency. Long-running goals also risked sending through whichever Lark profile happened to be active.

## Implementation

- let a connected goal register one repo-relative local config pointer under `.loopx/config/`
- expose only `enabled/config_pointer_registered` in the goal boundary, never the pointer value
- make `reviewer-request --goal-id --project` discover the sink automatically
- require explicit reader/user and sender/bot profile bindings for goal-default Lark sinks
- verify reader group access, dedicated bot identity, and reviewer membership in the sender app's `open_id` namespace before send
- persist only verified `sha256:` receipts in the existing PR lifecycle row
- preserve those receipts across later lifecycle monitor upserts
- make restart/retry return `already_notified` without another provider call
- document the contract in English and Chinese

No notification ledger or repository-specific branch was added.

## Validation

- `python3 examples/project/configure-goal-smoke.py`
- `python3 examples/issue-fix-reviewer-notification-sink-smoke.py`
- `python3 examples/issue-fix-reviewer-request-smoke.py`
- Ruff on all changed Python files
- `loopx canary premerge --from-git-diff --format json`: 17/17 selected checks passed; public boundary clean; 0 failures, warnings, or manual holds; self-merge allowed

## Risk and exclusions

- Goal-default execution fails closed if its local config or PR lifecycle row is missing.
- The explicit/manual legacy `bot_profile` input remains compatible; the stronger dual-profile contract applies to goal defaults.
- Secondary failure does not erase a verified canonical GitHub request.
- No credentials, profile values, destination/member IDs, local paths, raw provider responses, or runtime state are committed or emitted publicly.


huangruiteng:
No remaining findings after self-review.

During review I found and corrected one precision bug before merge: the Lark reviewer membership check compared the expected `open_id` against raw JSON text, so a prefix could theoretically match a different member. Commit `a1f13484` now parses the response and compares exact member ids, with a regression fixture for the prefix case.

Open questions / assumptions:

- The registered config remains local-private under `.loopx/config/`; only a boolean capability summary reaches public boundary projections.
- Cross-application Lark `open_id` values are intentionally verified in the sender bot application's namespace. The user profile is only used to prove destination access.
- A goal-default execute requires an existing PR lifecycle row. Verified `sha256:` receipts are stored in that row and preserved across later monitor upserts.

- Motivation: reviewer routing was already computed, but each pilot still needed bespoke sink flags and could duplicate notifications after restart.
- Solved: a connected goal can now register one repo-relative private config pointer; `reviewer-request --goal-id ... --project ...` discovers it, enforces reader/sender identity bindings, verifies the external write, and persists only an opaque receipt.
- Operator impact: routine reviewer notification becomes a default issue-fix capability while permission failures stay explicit and retry-safe.
- Main risk: external messaging permissions and provider identity namespaces. The implementation fails closed before send, keeps raw provider data/private ids out of output, and verifies the sent marker afterward.
- Design: this is a generic issue-fix contract with a real CLI call site and domain-state lifecycle, not repository-specific branching.

Validation:

- `python3 examples/issue-fix-reviewer-notification-sink-smoke.py`
- `python3 examples/issue-fix-reviewer-request-smoke.py`
- `uvx ruff check loopx/capabilities/issue_fix/reviewer_notification.py examples/issue-fix-reviewer-notification-sink-smoke.py`
- `loopx canary premerge --from-git-diff --format json`: 17/17 selected checks passed, public-boundary scan clean, zero failures/warnings/manual holds, `self_merge_allowed=true`.

Merge decision: self-merge approved under the standing owner authorization after the review correction above. The first real OpenViking notification remains separately time-gated and will validate the installed path before the capability todo is closed.

PR #1842, opened 2026-07-11
Add gated validated outcome memory writeback

huangruiteng:
## Motivation

Issue-fix can already read checkout-verified public repository context, but it has no controlled way to retain a proven delivery outcome for later advisory retrieval. OpenViking's experimental one-shot session command does not expose an idempotency key, so using it directly would make automatic retries unsafe.

## Implementation

- add an explicit `loopx issue-fix outcome --write-repository-memory` call site
- require independent local provider opt-in, completed delivery, passed validation, stable `recorded_at`, and a matching revision-scoped public resource namespace
- write only a distilled public fact with revision, provenance, freshness, public outputs, risk notes, workspace/peer scopes, and stable supersession/idempotency hashes
- reuse deterministic provider resource sync so identical retries perform no second write and conflicts stop instead of overwriting
- reject transcript, tool-result/log, expert-answer, credential, private-material, raw-content, and local-path capture
- keep conversation/session capture out of scope; `ov add-memory` is not used
- register the contract and focused smoke in the capability catalog and canary planner
- document the default-off contract in English and Chinese

## Validation

- `python3 examples/issue-fix-outcome-projection-smoke.py`
- `python3 examples/canary/catalog-planner-smoke.py`
- `loopx check --scan-path docs/capabilities/issue-fix/ --scan-path examples/issue-fix-validated-memory-writeback-smoke.py`
- `loopx canary premerge --from-git-diff`: 17/17 selected checks passed; 0 failures, 0 warnings, 0 manual holds

## Risk and limits

The hook is double-gated and remains disabled by default. It writes only deterministic public resources; it does not enable OpenViking session memory, automatic lifecycle capture, transcript capture, or private namespaces. A live OpenViking dogfood write still requires a separate project boundary decision and local configuration.

### DIFF
--- a/docs/capabilities/issue-fix/README.md
+++ b/docs/capabilities/issue-fix/README.md
@@ -460,7 +460,8 @@ the current repository revision remains authoritative.
 - multi-repository issue portfolios with bounded concurrency;
 - maintainer preference learning from public accepted/rejected outcomes;
 - reviewer load balancing and bus-factor awareness;
-- richer repository memory with explicit workspace/peer isolation;
+- evaluate validated-outcome memory utility across repeated issues before any
+  default enablement or session-memory expansion;
 - Open Knowledge Format interoperability after the repository-context contract
   stabilizes;
 - project-level metrics for accepted fixes, cycle time, human attention,
@@ -543,8 +544,8 @@ endings and one terminal newline are normalised), and
 persists only the compact hook projection in the existing repository context.
 Unverified hits contribute counts only; their summaries are not persisted.
 Provider unavailability, empty retrieval, or a missing checkout is fail-open;
-raw memory bodies, automatic transcript capture, memory writeback, private
-namespaces, credentials, and provider config paths are never retained.
+raw memory bodies, automatic transcript capture, private namespaces,
+credentials, and provider config paths are never retained.
 
 Minimal local provider config (the revision must also appear in `scope_ref`):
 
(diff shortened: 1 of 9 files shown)


huangruiteng:
Self-review complete.

- Changed surfaces: issue-fix outcome CLI, repository-memory provider contract, stable delivery timestamp projection, capability catalog/canary routing, focused smoke, English/Chinese capability docs.
- Contract review: writeback needs both explicit CLI execution and local `writeback_enabled`; it also requires completed delivery, passed validation, stable `recorded_at`, and matching revision scope. Session/conversation capture remains disabled.
- Idempotency review: the distilled fact has a stable content hash and immutable resource target; an identical retry produces zero provider writes, while provider content conflict stops rather than overwrites.
- Boundary review: raw transcript/tool logs/results/expert answers/credentials/private material and true raw/local capture flags are rejected; public packets expose only compact hashes and opaque provider refs.
- Direct checks: diff hygiene and changed Python compile passed.
- Focused checks: repository-memory, validated-writeback, outcome-projection, capability-guide, and catalog-planner smokes passed.
- Canary: 17/17 selected checks passed, 0 failures, 0 warnings, 0 manual holds; public boundary scan passed for all 9 changed files.
- Failures/skips: none. Live OpenViking writeback intentionally not enabled in project state; that remains a separate owner boundary decision.
- Merge decision: approved for self-merge under the standing owner authorization and the passed LoopX premerge gate.

PR #1844, opened 2026-07-11
Guard validated memory revision ancestry

huangruiteng:
## Motivation

Real OpenViking restart dogfood exposed a provenance bug in validated-outcome memory writeback. The persisted fact for public issue #3102 recorded repository revision `cbdab573...` and delivery commit `a071b948...`, but git proves that delivery commit is not an ancestor of the pinned revision. After restart the fact remained retrievable even though the pinned checkout still contained the pre-fix payload shape.

The previous gate checked that provider scope matched the outcome repository context, but did not prove that the recorded delivery commit was contained in that revision.

## Implementation

- require `--repo-path` for explicit validated-outcome memory writeback
- resolve the pinned revision and delivery `commit_ref` in the approved checkout
- require `git merge-base --is-ancestor <commit_ref> <repository_revision>` before any provider write
- return a compact public-safe blocker with zero external writes for missing, unresolved, or divergent commits
- keep checkout paths and raw git output out of packets
- tighten `commit_ref` so option-like refs cannot reach git
- document that squash flows must record the final merge/squash commit

## Validation

- focused writeback smoke covers ancestor success, identical retry idempotency, divergent history, zero provider writes on rejection, disabled/unsafe/failed-validation paths, and the CLI path
- real public pilot replay rejects the #3102 mismatch with `delivery_commit_not_in_repository_revision`, `commit_is_ancestor=false`, and `external_writes_performed=false`
- issue-fix outcome, workflow-contract, and bilingual capability-guide smokes passed
- `loopx canary premerge --from-git-diff`: 4 direct + 17 selected checks passed; 0 failures, warnings, or manual holds; public boundary clean

## Risk and omissions

- This intentionally tightens the writeback CLI: callers must provide an approved checkout.
- A squash workflow must record the final merge/squash commit rather than the superseded feature-branch commit.
- The previously written #3102 fact is not deleted automatically. Current retrieval classifies it as unverified with zero patch influence; cleanup or supersession needs a separately verified post-merge revision rather than destructive implicit repair.

### DIFF
--- a/docs/capabilities/issue-fix/README.md
+++ b/docs/capabilities/issue-fix/README.md
@@ -596,8 +596,12 @@ Validated-outcome writeback is a separate, default-off hook. It runs only when
 the caller explicitly adds `--write-repository-memory`, the local provider
 config independently sets `writeback_enabled: true`, delivery evidence says
 `completed`, validation says `passed`, the delivery evidence has a stable
-`recorded_at`, and the outcome revision matches the configured public resource
-scope. LoopX writes one distilled fact containing
+`recorded_at`, the outcome revision matches the configured public resource
+scope, and `--repo-path` proves with git that delivery `commit_ref` is an
+ancestor of that pinned revision. Divergent, missing, or unresolved commits
+block before the provider is called. Squash flows should record the final
+merge/squash commit, not a superseded feature-branch commit. The checkout path
+and raw git output are never retained. LoopX writes one distilled fact containing
 revision, provenance, freshness, public outputs, risks, a stable supersession
 key, and explicit workspace/peer scopes. A content hash selects the immutable
 target, so an identical retry reads and accepts the existing fact without a
@@ -785,6 +789,7 @@ loopx issue-fix outcome \
   --write-delivery-evidence \
   --repository-memory-provider-json provider.json \
   --write-repository-memory \
+  --repo-path /path/to/approved/repo \
   --agent-id codex-issue-fix \
   --format json
 ```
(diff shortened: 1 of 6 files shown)


huangruiteng:
## Self-review

### Findings

No blocking findings.

- Squash merge callers must provide the final merge/squash commit as `commit_ref`; a superseded feature-branch commit is correctly rejected because ancestry cannot prove that its content is present.
- The existing stale #3102 fact remains stored. This PR prevents new facts with the same provenance error; it does not perform destructive cleanup without a verified superseding revision.

- **Motivation:** restart dogfood proved persistence but refuted revision-pinned correctness: a retrievable outcome fact referenced a delivery commit absent from its repository revision.
- **Solved:** writeback now requires a caller-approved checkout and proves both commits resolve plus `commit_ref` is an ancestor of `repository_revision` before constructing/calling the provider.
- **Operator impact:** misleading outcome facts fail closed with a compact reason code and zero provider writes; checkout paths and raw git output remain local.
- **Main risk:** intentional CLI tightening for existing writeback users. The path is explicit/default-off, and docs plus error text describe the new requirement.
- **Design:** the invariant lives in the generic provider writeback boundary and has a real OpenViking replay; there is no repository-name special case or second ledger.

### Validation

- focused validated-memory writeback smoke: passed, including ancestor success, idempotent retry, divergent history rejection, and zero-write proof
- real #3102 replay: `delivery_commit_not_in_repository_revision`, `commit_is_ancestor=false`, `external_writes_performed=false`
- issue-fix outcome, workflow-contract, and bilingual guide smokes: passed
- `loopx canary premerge --from-git-diff`: 4 direct + 17 selected checks passed; 0 failures, warnings, or manual holds
- public/private boundary: 6 changed files scanned, clean
- GitHub remote checks: none configured for this branch

### Merge decision

Self-merge approved under the owner's standing authorization. Canary reports `self_merge_allowed=true`; admin bypass is used only for the repository review rule.

PR #1845, opened 2026-07-11
Fix legacy reviewer notification receipt persistence

huangruiteng:
## Motivation

Verified secondary reviewer notifications for OpenViking PRs #3133 and #3139 could not persist their hashed receipts because their existing PR lifecycle rows predated the `maintainer_correction_body_captured` boundary flag. The send succeeded, but the receipt-only upsert failed current validation, leaving cross-process retries able to notify again.

## Implementation

- Normalize only missing capture-boundary flags to explicit `false` in the receipt-only migration path.
- Preserve the existing public-boundary guard: any truthy capture flag still fails.
- Extend the provider-neutral reviewer-request smoke with a historical lifecycle row, restart-safe receipt persistence, and the truthy-flag rejection case.

## Validation

- `python3 examples/issue-fix-reviewer-request-smoke.py`
- Real historical PR #3133 row reproduced the failure before the patch and persisted successfully in an isolated ledger after the patch.
- `loopx canary premerge --from-git-diff`: 4 direct checks and 9 selected checks passed; 0 failures, warnings, or holds; public-boundary scan clean.

## Risk and non-goals

The change is limited to receipt metadata migration for rows already accepted into issue-fix domain state. It does not weaken truthy capture guards, change notification selection, send content, authority, or external provider behavior.

### DIFF
--- a/examples/issue-fix-reviewer-request-smoke.py
+++ b/examples/issue-fix-reviewer-request-smoke.py
@@ -743,6 +743,14 @@ def main() -> int:
             },
         )
         upsert_issue_fix_pr_lifecycle_ledger_jsonl(lifecycle_path, lifecycle)
+        legacy_lifecycle = json.loads(
+            lifecycle_path.read_text(encoding="utf-8").splitlines()[0]
+        )
+        legacy_lifecycle.pop("maintainer_correction_body_captured")
+        lifecycle_path.write_text(
+            json.dumps(legacy_lifecycle, sort_keys=True) + "\n",
+            encoding="utf-8",
+        )
         goal_execute_runner = FakeCombinedRunner(
             FakeGitHubRunner(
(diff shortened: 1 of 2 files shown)


huangruiteng:
Self-review complete.

- Product judgment: this is a backward-compatible receipt-only migration for rows already accepted into issue-fix domain state; it does not broaden notification authority or alter send behavior.
- Safety: only absent capture flags become explicit `false`; existing truthy flags remain blocked and are covered by the smoke.
- Regression proof: a historical lifecycle row missing `maintainer_correction_body_captured` fails before the patch and persists a hashed receipt after it.
- Validation: focused reviewer-request smoke passed; premerge canary passed 4 direct + 9 selected checks with 0 failures/warnings/holds; public-boundary scan clean.

No actionable review findings.

PR #1846, opened 2026-07-11
Fix auxiliary due monitor authorization after quota compaction

huangruiteng:
## Motivation

A quota decision can report multiple claimed due monitors while intentionally projecting only one due-monitor item in its compact payload. The monitor-poll gate treated that display projection as the complete execution allowlist, so an omitted but genuinely due monitor was rejected and could remain permanently overdue.

## Implementation

- preserve the compact quota payload contract;
- when an advancement lane carries `due_monitor_context`, resolve an explicitly requested monitor from the registry and authorize it only when it is open, unexpired, actually due, and claimed by the quota decision's current agent;
- keep the selected-monitor obligation strict, so target keys cannot redirect a poll to a different due monitor;
- add an end-to-end regression with two due monitors, one compactly projected, and an independent higher-priority advancement todo.

## Validation

- `python3 examples/control_plane/monitor-poll-policy-smoke.py`
- `python3 -m compileall -q loopx/control_plane/quota/monitor_poll.py examples/control_plane/monitor-poll-writeback-smoke.py`
- `loopx canary premerge --from-git-diff --format json`
  - standard tier, 2 changed files
  - 3/3 direct checks passed
  - Python compile passed
  - 7/7 catalog canaries passed
  - 8/8 risk-profile smokes passed
  - 1/1 public/private boundary check passed
  - no failures, skips, warnings, or manual holds

## Risk and non-goals

The fallback cannot authorize arbitrary or other-agent work: it requires the existing due-monitor context plus exact registry identity, due state, and claimant equality. It does not widen the selected-monitor lane and does not change quota payload size or projection ordering.

### DIFF
--- a/examples/control_plane/monitor-poll-writeback-smoke.py
+++ b/examples/control_plane/monitor-poll-writeback-smoke.py
@@ -64,6 +64,8 @@ def write_fixture(
     selected_target_key: str | None = TARGET_KEY,
     include_other_monitor: bool = False,
     include_user_gate: bool = False,
+    include_advancement: bool = False,
+    primary_monitor_priority: str = "P0",
 ) -> tuple[Path, Path]:
     project = root / "project"
     runtime = root / "runtime"
@@ -105,6 +107,17 @@ def write_fixture(
         if include_user_gate
         else ""
     )
+    advancement = (
+        "- [ ] [P0] Advance the independent delivery task.\n"
+        "  <!-- loopx:todo "
+        "todo_id=todo_monitorpolladvance "
+        "status=open "
+        "task_class=advancement_task "
+        f"claimed_by={AGENT_ID} "
+        "-->\n"
+        if include_advancement
+        else ""
+    )
     user_gate = (
         "\n## User Todo\n\n"
         "- [ ] [P0-user] Review the unrelated publication gate.\n"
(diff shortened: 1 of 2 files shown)


huangruiteng:
Self-review complete: **approved for owner-authorized self-merge**.

- **Changed surfaces:** quota monitor-poll authorization and its end-to-end writeback smoke.
- **Contract reviewed:** compact `monitor_due_items` remains presentation-only; the fallback resolves the exact registry todo and requires `due_monitor_context`, an advancement obligation, an open/unexpired/due monitor, and `claimed_by == decision agent`.
- **Negative boundary retained:** the selected due-monitor obligation still rejects redirection to a different monitor; no arbitrary, unclaimed, other-agent, not-due, or expired todo becomes executable.
- **Validation:** focused policy/writeback smokes and compile passed. `loopx canary premerge --from-git-diff` passed 3/3 direct checks, Python compile, 7/7 catalog canaries, 8/8 risk-profile smokes, and 1/1 public/private boundary check.
- **Failures/skips/manual holds:** none. GitHub reports no repository checks for this head.
- **Scope/privacy:** two public product/test files only; no credentials, local paths, raw logs, runtime state, or project-specific logic.

Coverage is sufficient because the regression exercises the real CLI read/write path with two due monitors, a deliberately compacted due-item projection, and a higher-priority advancement lane, while existing negative coverage proves selected-monitor target-key isolation.

PR #1847, opened 2026-07-11
Guard default CLI promotion from experimental worktrees

huangruiteng:
## Motivation

`install-local.sh` always replaced both the default `loopx` release snapshot and the live `loopx-canary` link. A dirty or feature worktree could therefore silently replace the CLI and skills used by unrelated active goals, while `doctor` compared two identities that had drifted together.

## Implementation

- automatically promote only a clean checkout whose HEAD matches the approved `origin/main` identity;
- route dirty, non-main, or otherwise unverified checkouts to `loopx-canary` only, leaving the default CLI, skills, and manual unchanged;
- retain an explicit `LOOPX_PROMOTE_DEFAULT=1` override after validation;
- mark no-clone GitHub archive installs as trusted archive promotions;
- persist `promotion_mode` in the release manifest and surface it through `loopx doctor` JSON/Markdown;
- update contributor/agent repair guidance and register a durable deep canary for the boundary.

## Product-path evidence

An isolated-HOME reproduction against the prior installer returned success while replacing an existing default executable from a feature worktree. The new smoke executes both real installer paths: the same untrusted checkout preserves the default and only updates canary; explicit promotion replaces the default and records `explicit_override` in both the manifest and doctor output.

## Validation

- `python3 examples/release/local-install-promotion-boundary-smoke.py`
- `python3 examples/install-local-overwrite-smoke.py`
- `python3 examples/release/release-version-contract-smoke.py`
- `python3 examples/install-local-smoke.py`
- `python3 examples/codex-cli-packaged-install-smoke.py`
- `bash -n scripts/install-local.sh scripts/install-from-github.sh`
- final-commit `loopx canary premerge --from-git-diff`
  - 4/4 direct checks passed
  - 8/8 catalog checks passed
  - 8/8 risk-profile smokes passed
  - 1/1 public/private boundary check passed
  - 17/17 selected checks passed, no failures, warnings, skips, or manual holds

## Risk and compatibility

Clean `origin/main` and `loopx update`/GitHub archive flows continue promoting automatically. The behavioral change is intentionally limited to dirty or unapproved local checkout installs. They now print the canary-only decision and the exact explicit-promotion command instead of silently mutating the default environment.


huangruiteng:
Self-review result: **no blocking code findings; approved after an independent reviewer**.

### Findings

No P0/P1/P2 correctness finding in the reviewed diff. The exact bad case is covered through the real installer path: an unapproved checkout preserves an existing default executable and updates only `loopx-canary`; an explicit promotion replaces default and records its mode in both `release.json` and `doctor`.

### Open assumptions

- `origin/main` is the default locally approved source identity; environments that intentionally use another clean reference must opt in explicitly.
- Existing release manifests without `promotion_mode` remain valid and project `null` through doctor.
- GitHub archive updates remain intentional default promotions and are labeled `trusted_github_archive`.

- **Motivation:** prevent one experimental worktree from silently changing the CLI and skills used by unrelated long-running goals.
- **Solved:** the mutation guard now lives at the installer call site, not in project-specific heartbeat logic; source identity, explicit override, manifest provenance, doctor visibility, agent repair guidance, and a durable canary all agree.
- **Operator impact:** normal clean-main and `loopx update` flows keep working; feature/dirty checkout installs become visibly canary-only.
- **Main risk:** contributor compatibility for workflows that previously relied on feature branches replacing default implicitly. The printed explicit-promotion command and updated guidance make that transition actionable.
- **Design:** this is a reusable default-vs-canary promotion contract. It avoids goal-specific fields and uses existing installer, manifest, doctor, catalog, and skill call sites.

### Validation

- focused real-path promotion smoke, local overwrite, release manifest, local install, and packaged install smokes passed;
- final commit `loopx canary premerge --from-git-diff` passed 4/4 direct checks, 8/8 catalog checks, 8/8 risk-profile checks, and 1/1 boundary check;
- 17/17 selected checks passed with zero failures, warnings, skips, or manual holds;
- public/private scan clean across all 17 changed files.

### Merge decision

`approved after reviewer`. Although the diff is focused and fully validated, it changes installed-user runtime promotion behavior, so I am retaining an independent review gate rather than admin self-merging it.

PR #1853, opened 2026-07-11
Fail closed on missing agent vision baselines

huangruiteng:
## Motivation

A required vision checkpoint could accept an unchanged reason even when the agent had never persisted a vision. Long-running campaigns could therefore lose their advancement acceptance gap and fall into monitor-only quiet.

## Implementation

- Reuse the existing run-history vision reducer to resolve the current per-agent baseline during refresh-state.
- Reject unchanged-with-reason as a satisfying decision when no baseline exists; project missing_required plus missing_baseline instead.
- Preserve missing-baseline context through compact history and quota acceptance-gap guidance.
- Document the first-baseline rule and cover the real refresh-state to quota path with a future-monitor fixture.

## Validation

- goal-vision-refresh-state-budget-smoke
- goal-vision-replan-contract-smoke
- quota-replan-decision-plane-smoke
- run-history-agent-context-retention-smoke
- run-compaction-readmodel-smoke
- heartbeat-quota-flow-smoke
- LoopX premerge canary: 4 direct checks and 17 selected checks passed; 0 failures, warnings, skips, or manual holds
- Public/private boundary scan clean across all 5 changed files
- Real OpenViking compatibility preview accepts unchanged-with-reason after a persisted campaign vision exists

## Risk

The behavior change is limited to an invalid first-checkpoint case. Existing agents with a persisted active vision keep unchanged-with-reason behavior; explicit retire/supersede paths remain available for bounded goals.

### DIFF
--- a/docs/reference/protocols/goal-vision-replan-contract-v0.md
+++ b/docs/reference/protocols/goal-vision-replan-contract-v0.md
@@ -129,8 +129,10 @@ Valid checkpoint decisions are:
 
 - `patched`: the refresh wrote a bounded `agent_vision` packet for the same
   `agent_id`;
-- `unchanged_with_reason`: the current per-agent vision still applies, with a
-  compact public-safe reason;
+- `unchanged_with_reason`: an already persisted current per-agent vision still
+  applies, with a compact public-safe reason. A reason cannot create the first
+  vision baseline; without one, the checkpoint remains `missing_required` and
+  requires `write_vision_patch`;
 - `retired_or_superseded`: the frontier was explicitly closed, superseded, or
   given a no-follow-up rationale;
 - `missing_required`: the turn was material but did not make a per-agent vision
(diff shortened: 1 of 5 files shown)


huangruiteng:
No blocking findings.

Open assumptions: the compact run index remains the canonical local source for the latest per-agent vision. The new lookup runs only when an unchanged reason is supplied, so ordinary refresh-state calls do not gain a history-read cost.

Product and architecture judgment: this closes a real campaign-control failure where reason text could masquerade as a missing vision baseline, erase acceptance gaps, and permit monitor-only quiet. The implementation is reusable rather than project-specific: it reuses the existing latest-agent-vision reducer, preserves missing-baseline context through compact history, and lets the existing goal-frontier/quota replan path do the scheduling work. Existing agents with a persisted vision retain unchanged-with-reason behavior; bounded goals can still use explicit retire or supersede closeout.

Validation: focused refresh-state and quota regression passed; goal-vision protocol, quota decision-plane, run-history retention, run-compaction, and heartbeat flow smokes passed. Premerge canary passed 4 direct checks plus 17 selected checks with 0 failures, warnings, skips, or manual holds. Public/private boundary scan was clean for all 5 changed files. A real OpenViking dry-run confirmed compatibility once the campaign vision exists.

Merge decision: self-merge approved under the owners standing authorization and the clean premerge gate.

PR #1855, opened 2026-07-11
Fix quota spend after material monitor transitions

huangruiteng:
## Summary

- allow a material `quota_monitor_poll` with an accountable delivery outcome to be consumed by the next quota spend
- keep unchanged monitor polls non-accountable
- keep duplicate accounting blocked after a `quota_slot_spent` event

## Motivation

The heartbeat contract says a material monitor transition may spend once after validated writeback. The monitor command already records `material_change=true` and `delivery_outcome=outcome_progress`, but slot accounting rejected every monitor classification before reading those fields. This stranded the writeback-to-spend lifecycle.

## Validation

- `python3 examples/control_plane/quota-material-monitor-spend-smoke.py`
- `python3 examples/control_plane/quota-spend-agent-identity-smoke.py`
- `python3 -m py_compile loopx/control_plane/quota/slot_accounting.py`
- `python3 -m loopx.cli --format json canary premerge --from-git-diff`
  - selected 14, failures 0, warnings 0, manual holds 0
- public/private boundary scan: clean, 2 files scanned

## Risk

The allowance is narrow: both `material_change=true` and an accountable `delivery_outcome` are required. Ordinary monitor no-change events remain non-spendable, and an existing spend event still prevents duplicate accounting.

## Merge decision

Hold for independent review because this changes quota hot-path runtime behavior, despite the narrow diff and passing premerge gate.

### DIFF
--- a/examples/control_plane/quota-material-monitor-spend-smoke.py
+++ b/examples/control_plane/quota-material-monitor-spend-smoke.py
@@ -1,0 +1,111 @@
+#!/usr/bin/env python3
+"""Smoke-test quota accounting after a material monitor transition."""
+
+from __future__ import annotations
+
+import json
+import sys
+import tempfile
+from pathlib import Path
+from typing import Any
+
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+if str(REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(REPO_ROOT))
+
+from loopx.control_plane.quota.slot_accounting import (  # noqa: E402
+    build_quota_slot_preview_for_decision,
+)
+
+
+GOAL_ID = "material-monitor-spend"
+
+
+def write_run_index(runtime: Path, records: list[dict[str, Any]]) -> None:
+    index_path = runtime / "goals" / GOAL_ID / "runs" / "index.jsonl"
+    index_path.parent.mkdir(parents=True, exist_ok=True)
+    index_path.write_text(
+        "".join(json.dumps(record) + "\n" for record in records),
+        encoding="utf-8",
+    )
+
+
+def preview(runtime: Path) -> dict[str, Any]:
+    quota = {
+        "compute": 1.0,
+        "window_hours": 24,
+        "slot_minutes": 1,
+        "spent_slots": 0,
+        "allowed_slots": 1440,
+    }
+    before = {
+        "ok": True,
+        "goal_id": GOAL_ID,
+        "should_run": False,
+        "effective_action": "monitor_quiet_skip",
+        "state": "eligible",
+        "safe_bypass_allowed": False,
+        "quota": quota,
+    }
+    status = {
+        "runtime_root": str(runtime),
+        "attention_queue": {"items": [{"goal_id": GOAL_ID}]},
+        "run_history": {"goals": [{"id": GOAL_ID, "quota": quota}]},
+    }
+
+    return build_quota_slot_preview_for_decision(
+        status,
+        goal_id=GOAL_ID,
+        before=before,
+        after_decision=lambda _: {"state": "eligible", "should_run": False},
+        quota_status_builder=lambda goal, **_: goal["quota"],
+        self_repair_spend_actions=frozenset(),
+    )
+
+
+def main() -> int:
+    unchanged_poll = {
+        "generated_at": "2026-01-01T00:00:00+00:00",
+        "goal_id": GOAL_ID,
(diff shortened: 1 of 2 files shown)


huangruiteng:
## Findings

**[P1] Bind delivery-completion spend to the delivery run agent identity** — `loopx/control_plane/quota/slot_accounting.py:108`

The new material-monitor allowance makes a `quota_monitor_poll` accountable, but `_latest_unspent_accountable_delivery_run()` selects only by goal and `build_quota_slot_preview_for_decision()` never checks the selected run's `agent_id` against the requested `--agent-id`. In a peer goal, `agent-a` can therefore consume `agent-b`'s material monitor transition; the resulting `quota_slot_spent` event is then attributed to `agent-a`, and the global spent marker prevents B from accounting for its own delivery.

Minimal reproduction on commit `d0aec11d`: write one material `quota_monitor_poll` with `agent_id=agent-b`, request the slot preview with `agent_id=agent-a`; the preview returns `ok=true` and `delivery_completion_spend=true`.

Please make agent-scoped completion accounting fail closed when both identities are present and differ, while preserving an explicit compatibility path for genuinely unscoped legacy runs/calls. Add matrix coverage for same-agent allowed, cross-agent rejected, unchanged poll rejected, and duplicate spend rejected. The current `quota-spend-agent-identity-smoke.py` proves propagation of the requested id, not ownership matching.

The product motivation is sound: a validated material monitor transition should be spendable exactly once, while unchanged polls remain free. The four-line allowance is appropriately narrow, but peer runtime makes identity part of the accounting key; goal-only selection is not sufficient once monitor transitions become accountable. Fixing this in the shared run-selection/accounting seam keeps the rule generic rather than monitor-specific.

## Validation

Passed on PR head:
- `quota-material-monitor-spend-smoke.py`
- `quota-spend-agent-identity-smoke.py`
- `monitor-poll-writeback-smoke.py`
- `py_compile` and `git diff --check`

The additional cross-agent probe fails as described.

## Merge Decision

**Request changes / hold.** Re-review after the ownership-match regression is added.

huangruiteng:
Addressed the P1 ownership finding in commit `64626d85`.

### Changed surfaces

- Delivery-completion run selection now receives the requested agent identity.
- When both the requested identity and delivery-run identity exist, mismatches fail closed.
- Missing identity on either side remains the explicit legacy/unscoped compatibility path.
- Successful previews and spend events expose `delivery_run_agent_id` for auditability.

### Regression matrix

- same-agent material monitor spend: allowed
- cross-agent material monitor spend: rejected
- unscoped caller with scoped legacy run: allowed
- scoped caller with unscoped legacy run: allowed
- unchanged monitor poll: rejected
- duplicate spend after `quota_slot_spent`: rejected

### Validation

- focused material-monitor spend smoke: passed
- quota spend agent-identity smoke: passed
- monitor-poll writeback smoke: passed
- changed Python compile and diff checks: passed
- premerge canary: selected 14, failures 0, warnings 0, manual holds 0

Requesting fresh independent re-review. The PR remains held until that review clears the runtime accounting change.

huangruiteng:
## Findings

No remaining findings on commit `64626d85`. The prior P1 ownership bug is fixed.

## Assumptions / Residual Risk

- Agent-scoped callers and runs now match when both identities are present. Missing identity on either side remains an explicit legacy/unscoped compatibility path; this is acceptable because current peer heartbeat paths supply `--agent-id`, while old/manual callers remain operable.
- Goal-wide run ordering remains the existing completion-accounting model. This PR does not redesign concurrent delivery correlation, and no new issue in that pre-existing seam was found within this bounded fix.

The implementation stays right-sized and generic: ownership matching lives in the shared accountable-run selector, not in monitor-specific CLI code. Material monitor transitions can now be charged exactly once without letting one identified peer consume another peer's delivery. Exposing `delivery_run_agent_id` in preview/event output makes the decision auditable.

## Independent Validation

Passed on PR head:
- `quota-material-monitor-spend-smoke.py` (same-agent allowed, cross-agent rejected, scoped/unscoped compatibility, unchanged rejected, duplicate rejected, audit field)
- `quota-spend-agent-identity-smoke.py`
- `monitor-poll-writeback-smoke.py`
- changed Python compile and diff checks
- `canary premerge --from-git-diff`: 14 selected, 14 passed, 0 warnings, 0 manual holds

## Merge Decision

**Approved after independent re-review.** The runtime hold from the previous review is cleared; implementation owner may merge under the normal PR boundary.

PR #1856, opened 2026-07-11
Fix runtime capability projection in thin heartbeats

huangruiteng:
## Motivation

A generated thin heartbeat without explicit host capability arguments tells the worker to run a bare quota guard. If the live runtime actually has network or polling capability, that observed capability can be omitted and an executable todo can be misclassified as a user gate.

## Implementation

- Teach the thin dispatcher to pass capabilities actually observed at runtime with `--available-capability` and never turn them into user gates.
- Keep explicit generator declarations as the deterministic path for hosts whose capabilities are known at install time.
- Extend the thin interface budget by 70 characters for this safety invariant.

## Validation

- `PYTHONPATH=. python3 examples/control_plane/heartbeat-prompt-smoke.py`
- OpenViking registry-backed thin prompt replay: runtime projection rule present, interface budget 1513/1570.
- `python3 -m py_compile loopx/heartbeat_prompt.py examples/control_plane/heartbeat-prompt-smoke.py`
- `loopx check` on all three changed public surfaces: 8 checks, 0 errors, 0 warnings.
- `loopx canary premerge --from-git-diff`: 14 selected checks passed, 0 failures, 0 manual holds.

## Risk

This changes prompt guidance only; it does not grant capabilities or assume network is universally available. The worker must project only capabilities the current host actually exposes.

### DIFF
--- a/docs/heartbeat-automation-prompt.md
+++ b/docs/heartbeat-automation-prompt.md
@@ -150,6 +150,13 @@ The generated quota guard and spend command preserve the same declarations.
 Do not declare credentials, production access, or another capability merely to
 bypass a gate; the launcher must actually provide it.
 
+When a thin prompt is generated without explicit declarations, it still tells
+the runtime worker to project non-basic capabilities that are actually present
+with `--available-capability`. This prevents an observed network or polling
+capability from becoming a false user gate without guessing capabilities the
+host does not have. Explicit generator arguments remain preferred for hosts
+whose capabilities are known when the automation is installed.
+
 - for small AGENTS-eligible validated changes, self-merge and complete the todo
   with `--self-merged --evidence "<commit and validation summary>"`;
 - for an independent continuation, create `--next-agent-todo` and optionally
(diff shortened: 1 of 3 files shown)


huangruiteng:
Self-review

Findings: none.

Assumption: only capabilities observed in the current host are projected; the prompt does not infer network, credentials, or production access. The follow-up wording explicitly scopes the rule to observed runtime capabilities.

Product/architecture judgment: the bad case is a thin heartbeat whose generated guard omits a capability the live worker can actually see, causing a false user gate. Keeping explicit generator declarations as the deterministic install-time path and adding a runtime fallback invariant to the generic thin dispatcher solves the dynamic-host case without repository-specific logic or capability grants. The main risk is prompt ambiguity and interface growth; the wording is observation-scoped and the thin budget remains bounded at 1522/1570.

Validation:
- heartbeat prompt smoke passed
- OpenViking registry-backed thin prompt replay passed
- Python compile passed
- public boundary check: 8 checks, 0 errors, 0 warnings
- premerge canary: 14 selected checks passed, 0 failures, 0 manual holds
- diff check passed

Decision: self-merge. This is a narrow, reversible prompt-contract repair with durable smoke coverage and no permission, credential, production, benchmark, or private-state change.

PR #1857, opened 2026-07-11
Keep Lark auth secrets out of Kanban commands

huangruiteng:
## Motivation

The Kanban adapter should keep Base, table, view, and record identifiers visible for operations and audit. Real authentication material must remain owned by lark-cli auth and must never be added to adapter-generated command arguments.

## Implementation

- reject known AK/SK and secret-bearing auth options before the subprocess runner starts
- retain Base and table resource identifiers in command evidence
- document the authentication boundary and the lark-cli auth workflow
- add a generic smoke covering both rejection and allowed resource identifiers

## Validation

- python3 examples/lark-kanban-control-plane-smoke.py
- python3 -m py_compile loopx/presentation/sinks/lark/kanban.py examples/lark-kanban-control-plane-smoke.py
- loopx check on all three changed public files
- loopx canary premerge --from-git-diff: 14 selected checks passed, no manual holds

## Risk and uncovered areas

The guard only applies to adapter-generated argv. It intentionally does not redact normal Kanban resource identifiers or rewrite lark-cli output. Authentication storage and renewal remain lark-cli responsibilities.

### DIFF
--- a/docs/lark-kanban-control-plane-adapter.md
+++ b/docs/lark-kanban-control-plane-adapter.md
@@ -15,6 +15,20 @@ and newly discovered tasks should be created with `loopx todo add`,
 `loopx todo complete --next-*`, `loopx todo supersede`, or a typed planning
 intake, then projected back to Lark with `sync-loopx-todos`.
 
+## Authentication Boundary
+
+The adapter passes Lark resource identifiers such as `base_token`, `table_id`,
+`view_id`, and `record_id` to `lark-cli`. These identifiers remain visible in
+command evidence because they are useful for operating and auditing the board;
+the adapter does not treat them as AK/SK secrets.
+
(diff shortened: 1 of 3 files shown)


huangruiteng:

Product/architecture judgment: this keeps the operator-visible Base/Table identifiers intact while enforcing the narrower invariant that real authentication material stays in lark-cli auth. The guard is generic at the shared command-runner boundary, not OpenViking-specific. Main risk is future secret-bearing flag aliases not present in the denylist; the current adapter has no such call site, and the smoke proves known AK/SK/app-secret/access-token forms are rejected before execution while Base and table identifiers remain allowed.

Validation: Lark Kanban smoke passed; schema CLI passed; changed Python compiled; public-boundary scan passed; premerge canary passed 14 selected checks with no manual holds. Merge decision: self-merge under standing owner authorization.

PR #1859, opened 2026-07-11
Keep monitor lanes out of Explore branch budgets

huangruiteng:
## Summary

- exclude `continuous_monitor` todos from todo and worker exploration schedulers while preserving them as rejected diagnostics
- make the shared-input contract explicit: `shared_artifact:*` and `shared_implementation:*` capabilities may feed independent variant write scopes, while mutable shared builds remain write conflicts
- add synthetic long/short portfolio coverage for analysis-only planning and document the scope semantics

## Validation

- 6 focused fixture-free test functions passed
- `python3 examples/explore-worker-plan-gate-smoke.py`
- `loopx canary premerge --from-git-diff --tier standard --timeout-seconds 180`
- Python compile and `git diff --check`

The full pytest suite was not run locally because pytest is not installed in the available environment.

## Boundary

The planners remain read-only and analysis-only behavior still suppresses all claim and lease commands. Fixtures and docs use only synthetic public-safe names.

## Remaining follow-ups

- worker capacity is still represented by one global width rather than per-resource lane capacities
- dependency invalidation happens after selection and does not backfill the newly empty slot; typed evidence remains diagnostic-only

### DIFF
--- a/docs/capabilities/explore/README.md
+++ b/docs/capabilities/explore/README.md
@@ -136,6 +136,8 @@ spend quota, or change the active state. Instead it emits a prediction
 packet with:
 
 - selected branches, confidence, hazards, and reason codes;
+- excluded `continuous_monitor` diagnostics, which remain visible but never
+  enter the exploration scheduler or consume branch width;
 - a dry-run A/B estimate comparing baseline serial execution with the
   DSpark-style selected prefix (`ab_result.estimated_speedup_vs_baseline`);
 - suggested `loopx todo claim` and `loopx task-lease acquire` commands for a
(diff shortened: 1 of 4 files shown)


huangruiteng:
No blocking findings.

- Motivation: Explore planning was treating observation-only monitors as speculative work and conflating immutable shared dependencies with mutable write conflicts, which reduced useful parallel experiment width.
- Solved: the planners now exclude `continuous_monitor` lanes before scheduling while retaining them as diagnostics, and expose an explicit public-safe shared-dependency contract without weakening mutable write-scope exclusion.
- Operator impact: read-only branch plans preserve independent experiment lanes and stop spending verification/worker width on monitors.
- Main risk: per-resource capacities and dependency-driven slot backfill remain follow-ups; this PR does not claim to solve them. The changed surfaces remain analysis-only and do not acquire claims, leases, workers, state, or quota authority.
- Design judgment: the change is narrow and reusable. It keeps scheduling semantics in the existing planner modules and preserves the current authority boundary.

Validation performed:
- 3 focused synthetic portfolio tests passed directly.
- `examples/explore-worker-plan-gate-smoke.py` passed.
- `examples/explore-result-layer-smoke.py` passed.
- `examples/explore-configure-goal-smoke.py` passed.
- Standard `loopx canary premerge --from-git-diff` selected risk-based control-plane canaries; observed checks passed.
- Python compile and `git diff --check` passed.
- Local and rendered PR diff public/private scans were clean.
- Full pytest suite was not run because pytest is unavailable in this worktree environment; focused coverage exercises both todo and worker planners plus the mutable-shared-build negative case.

Merge decision: self-merge approved under the owner-authorized narrow analysis-only runtime seam policy.

PR #1860, opened 2026-07-11
Fix issue-fix JSON input boundary

huangruiteng:
## Motivation

Issue-fix JSON options could be read as accepting compact JSON, but the shared loader always treated the argument as a file path. Long inline objects therefore raised a file-name error and echoed the input in the public error packet.

## Implementation

- Accept bounded inline JSON objects while preserving file-path and stdin inputs.
- Return compact parse/read/type errors without repeating the supplied payload or local path.
- Clarify reviewer-request help and register a provider-neutral CLI regression in the capability and canary catalogs.

## Validation

- issue-fix JSON input boundary smoke
- workflow-plan smoke
- reviewer-request smoke
- catalog planner smoke
- Python compile
- LoopX public-boundary check: 5 files, 0 warnings
- premerge canary: 17 selected checks, 0 failures, 0 warnings, 0 manual holds

## Risk and uncovered scope

The parser uses a 1 MiB cap only for inline JSON; existing file and stdin behavior remains compatible. This PR does not change external-write authority or provider behavior.

### DIFF
--- a/examples/canary/catalog-planner-smoke.py
+++ b/examples/canary/catalog-planner-smoke.py
@@ -533,6 +533,7 @@ def assert_pr_release_and_refactor_profiles_select() -> None:
     reviewer_profile = product_entry_profiles["issue-fix-reviewer-routing"]
     reviewer_commands = [check["command"] for check in reviewer_profile["checks"]]
     assert reviewer_commands == [
+        "python3 examples/issue-fix-json-input-boundary-smoke.py",
         "python3 examples/issue-fix-capability-guide-smoke.py",
         "python3 examples/issue-fix-reviewer-recommendation-smoke.py",
         "python3 examples/issue-fix-reviewer-request-smoke.py",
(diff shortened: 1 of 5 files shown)


huangruiteng:
Self-review

Findings: none.

Open assumption: inline JSON is recognized only for object/array-looking input; ordinary strings remain file paths, preserving the existing path contract. The 1 MiB cap applies to inline content, while existing file and stdin inputs remain compatible.

Product/architecture judgment: this fixes the shared issue-fix input boundary instead of special-casing reviewer-request. It removes the ENAMETOOLONG/raw-input error failure mode for every current issue-fix JSON-object caller, keeps external-write authority unchanged, and makes the new boundary smoke a first-class capability/canary check. The main compatibility risk is path-vs-inline disambiguation for filenames beginning with an opening JSON delimiter; that edge is acceptable for a JSON-object CLI input and is outweighed by deterministic safe parsing.

Validation: focused JSON-boundary, workflow-plan, reviewer-request, and catalog-planner smokes passed; changed Python compiled; public-boundary scan covered 5 files with 0 warnings; premerge canary ran 17 selected checks with 0 failures, 0 warnings, and 0 manual holds.

Merge decision: self-merge approved under the owners standing authorization for LoopX PRs.

PR #1861, opened 2026-07-11
Render monitor-poll rejection details

huangruiteng:
## Motivation

A rejected quota monitor-poll returned an actionable JSON packet, but the Markdown renderer read only the success-only monitor_event object. Operators therefore saw None for the source, target, material-change intent, and reason, and could not distinguish a rejected write from a recorded poll.

## Implementation

- Render ok=false monitor-poll packets through a dedicated rejection view.
- Preserve mode, source, todo/target, material-change intent, and reason.
- Explicitly show appended=false and registry_mutated=false.
- Extend the provider-neutral CLI smoke with the rejected target-key path.

## Validation

- monitor-poll writeback smoke, including red-then-green CLI Markdown regression
- monitor-poll policy smoke
- quota material-monitor spend smoke
- status Markdown smoke
- catalog planner smoke
- changed Python compile
- public-boundary scan: 2 files, 0 warnings
- premerge canary: 16 selected checks, 0 failures, 0 warnings, 0 manual holds

## Risk and uncovered scope

This changes presentation only; monitor eligibility, todo writeback, scheduler policy, and quota accounting are unchanged. Success Markdown remains on the existing renderer path.

### DIFF
--- a/examples/control_plane/monitor-poll-writeback-smoke.py
+++ b/examples/control_plane/monitor-poll-writeback-smoke.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import argparse
 import json
+import subprocess
 import sys
 import tempfile
 from datetime import datetime, timezone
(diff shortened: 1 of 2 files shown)


huangruiteng:
Self-review

Findings: none.

Open assumption: failure packets keep appended, registry_mutated, and material_change as booleans, as enforced by the monitor-poll builder and CLI exception path.

Product/architecture judgment: the bug belongs in presentation, not monitor eligibility or quota policy. A dedicated rejection view is the smallest reusable fix because both policy rejections and exception-derived monitor-poll failures share the same top-level packet shape. Operators now see the attempted target and the actionable reason while the output explicitly proves that no event or registry mutation occurred. Success rendering remains untouched.

Validation: the CLI regression was observed red then green; monitor-poll writeback, policy, material-spend, status Markdown, and catalog planner smokes passed; changed Python compiled; public-boundary scan covered 2 files with 0 warnings; premerge canary ran 16 selected checks with 0 failures, 0 warnings, and 0 manual holds.

Merge decision: self-merge approved under the owners standing authorization for LoopX PRs.

PR #1880, opened 2026-07-11
Track capability-gap lifecycle metrics

huangruiteng:
## Motivation

The issue-fix monthly report can count capability gaps only from a hand-authored event batch. That makes the found/fixed/real-callsite-verified capability delta easy to omit and hard to audit against the Kanban work that produced it.

## Implementation

- add an explicit `--capability-gap-status` marker to agent todo add/update, requiring target-capability metadata
- append a typed, goal-scoped `capability_gap` rollout event keyed by the existing todo id
- compose found/fixed/real-callsite-verified counts from rollout evidence when audited coverage spans the reporting period
- keep partial observation visible under coverage while leaving incomplete metrics unavailable
- document the contract in the English and Chinese issue-fix entry points and protocol

## Validation

- `python3 examples/issue-fix-capability-gap-metrics-smoke.py`
- `python3 examples/issue-fix-metrics-supplement-smoke.py`
- `python3 examples/control_plane/todo-lifecycle-cli-smoke.py`
- `loopx canary premerge --from-git-diff`
- `loopx check` public-boundary scan: 0 errors, 0 warnings

## Risk and uncovered cases

The new marker is opt-in and does not infer gaps from prose. Older gaps remain unavailable until an operator audits/backfills the reporting window and declares its coverage start. Repeated lifecycle markers may append duplicate events, but metrics fold by todo id and status, so counts remain idempotent. No external provider behavior, permissions, credentials, or private runtime state are changed.

### DIFF
--- a/docs/capabilities/issue-fix/README.md
+++ b/docs/capabilities/issue-fix/README.md
@@ -343,7 +343,7 @@ tests whether the system is a durable employee rather than a scripted demo.
 | Maintainer correction | `loopx issue-fix pr-lifecycle --maintainer-correction-json ... --execute-transition` | Turn bounded public review feedback into one claimed patch successor, a concrete user gate, or a quiet unchanged poll. |
 | Metrics projection | [`loopx issue-fix metrics`](protocols/issue-fix-metrics-projection-v0.md) | Keep repository baseline separate from attributable agent output, combine existing feasibility/PR lifecycle rows with caller-supplied public snapshots, and report deltas, ratios, inventory, and missing data without another ledger. |
 | Repository snapshot | `loopx issue-fix repository-snapshot` | Explicitly collect bounded public GitHub stock/flow and known issue/PR state; optionally retain only material daily changes in the existing issue-fix domain state. |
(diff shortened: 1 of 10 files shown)


huangruiteng:
## Self-review

### Findings

No blocking findings after one amendment. The initial review found that `real_callsite_verified` could be asserted without evidence; the PR now requires public-safe `--evidence` for both `fixed` and `real_callsite_verified`, and the focused smoke covers the failure path.

### Open assumptions

- A capability gap is goal-scoped and keyed by its existing todo id.
- Backdated coverage is allowed only after auditing and backfilling every in-period gap todo; the CLI deliberately does not infer this from prose.
- Repeated typed events are acceptable because the metric fold is idempotent by todo id and status.

Motivation: month-end capability output was easy to omit because it required a hand-authored event batch. The implementation reuses the existing todo as the human work surface and the existing rollout log as the audit surface, instead of adding another ledger or issue-fix state machine. It solves the intended found/fixed/real-callsite-verified composition while preserving honest missing-data semantics for unaudited history. Installed-user risk is low: the todo flag and metrics coverage argument are opt-in; existing todo, metrics, provider, and permission behavior are unchanged.

### Validation

- focused capability-gap lifecycle/metrics smoke passed
- existing metrics supplement, todo lifecycle, and bilingual guide smokes passed
- repository Python line budget passed
- `loopx canary premerge --from-git-diff` direct checks and 8/8 selected catalog canaries passed
- public-boundary scan: 0 errors, 0 warnings

### Merge decision

Self-merge approved under the owner standing authorization for focused LoopX PRs. Branch protection requires admin merge because GitHub does not permit the PR author to satisfy its own review requirement.

PR #1890, opened 2026-07-11
Reconcile uncertain context-provider writes before retry

huangruiteng:
## Motivation

OpenViking `add-resource --wait` can return a transport error after the server has already accepted the resource and queued semantic processing. Treating every non-zero exit as a failed write makes a generic repository-context provider retry blindly and risks duplicate work.

This was reproduced by the OpenViking issue-fix pilot while indexing a public, revision-scoped repository snapshot.

## Implementation

- reserve a bounded reconciliation window for uncertain provider writes
- read back the exact target, then its tree/listing, before deciding whether a retry is safe
- distinguish `completed`, `committed_pending`, and `partial`
- expose public-safe `pending_count`, `reconciliation_performed`, and `retry_disposition`
- never copy raw provider payloads or source content into the public packet

## Validation

- Python compile and diff hygiene checks
- `loopx canary premerge --from-git-diff --format markdown`
  - passed, 6 selected checks, 0 failures
  - public-boundary check passed
  - self-merge allowed
- real OpenViking call: a non-zero/timeout write was reconciled by exact-target read as completed; one write was recorded and no duplicate retry was issued

## Risk and limits

The change is confined to the provider adapter and additive public packet fields. A target that is visible but not yet fully readable is classified as `committed_pending` and must be polled instead of retried. This PR does not yet implement revision activation/supersession for repository snapshots; that remains a separate provider-neutral lifecycle successor.

### DIFF
--- a/examples/issue-fix-repository-memory-smoke.py
+++ b/examples/issue-fix-repository-memory-smoke.py
@@ -28,6 +28,7 @@ from loopx.capabilities.issue_fix.repository_memory_provider import (  # noqa: E
 from loopx.capabilities.context_providers.base import (  # noqa: E402
     ContextProviderItem,
     ContextProviderRetrieval,
+    ContextProviderSync,
 )
 from loopx.capabilities.context_providers.openviking import (  # noqa: E402
     OpenVikingContextProvider,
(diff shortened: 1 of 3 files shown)


huangruiteng:
No blocking findings.

- The provider target is an immutable or idempotent logical destination. When readback sees a target but cannot yet verify complete content, the adapter deliberately returns `committed_pending`; callers must poll instead of retrying.
- Revision activation and stale-resource supersession are intentionally out of scope and tracked as a separate provider-neutral lifecycle change.

- **Motivation:** prevent duplicate provider mutations when transport failure occurs after commit.
- **Solved:** the adapter now reconciles exact target/tree/list state and exposes an explicit retry disposition. Contract tests cover committed-but-pending, verified, and absent outcomes; a real OpenViking call exercised the uncertain-write path.
- **Operator impact:** a non-zero CLI exit no longer automatically means the resource write failed.
- **Main risk:** eventual provider visibility. The conservative `committed_pending` state avoids retry when any target evidence exists.
- **Design:** the public sync contract is provider-neutral; OpenViking-specific probing stays inside its adapter.

## Validation

- repository-memory, validated-writeback, and capability-guide smokes passed
- premerge canary passed: 6 selected checks, 0 failures
- canary reports `self_merge_allowed=true`

## Merge decision

Self-merge approved under the owner's standing authorization for validated LoopX PRs.

PR #1899, opened 2026-07-11
Preserve agent-scoped quota spend attribution

huangruiteng:
## Summary
- skip explicitly other-agent history rows when resolving an agent-scoped accountable delivery
- preserve existing unscoped and legacy-history stopping behavior
- add an interleaved peer delivery/spend regression and a reusable self-repair pattern

## Validation
- focused quota material-monitor, agent-identity, and heartbeat flow smokes
- `loopx canary premerge --from-git-diff --git-diff-base origin/main --tier standard`
- 16/16 selected canaries passed; public boundary clean; no failures, skips, or manual holds

## Risk
Narrow quota read-model correction. It does not change spend eligibility, limits, or unscoped global accounting.


huangruiteng:
Maintainer premerge review

- **Changed surfaces:** quota slot attribution read model, focused control-plane regression smoke, self-repair pattern catalog.
- **Direct checks:** committed/staged/unstaged diff checks and changed-Python compile all passed.
- **Catalog canaries:** 7/7 passed.
- **Risk-profile smokes:** 8/8 passed.
- **Public/private boundary:** clean across all 3 changed files.
- **Failures/skips:** none.
- **Manual holds:** none.
- **Merge decision:** self-merge is appropriate. The change only skips explicitly other-agent events during an explicitly agent-scoped lookup; unscoped and legacy semantics remain unchanged, and interleaved-peer behavior is covered.

PR #1906, opened 2026-07-11
Simplify issue-fix repository context to rolling indexes

huangruiteng:
## Motivation

The `checkout_head` repository-memory policy derived a new immutable provider scope for every checkout revision and added a separate activation lifecycle before advisory retrieval could run. That duplicates a provider-managed rolling default-branch refresh loop and can block useful navigation even though LoopX still verifies every hit against the current checkout before decision influence.

## Implementation

- make `rolling_default_branch` the default when provider config omits a pinned revision;
- use one stable provider scope and project it as advisory, externally refreshed, and never patch authority;
- retain current-checkout exact/parser-chunk verification for every retrieved hit;
- remove the unused per-checkout revision-plan and activation module and its `active_*` state;
- fail old `checkout_head` activation fields with an actionable migration error;
- retain explicit `pinned` compatibility and the separately gated manual sync and validated-outcome writeback paths;
- update the English and Chinese issue-fix capability guide.

## Validation

- issue-fix workflow-plan, feasibility, repository-context, workflow-contract, and workflow-e2e smokes
- `PYTHONPATH=. python3 -m loopx.cli canary premerge --from-git-diff` (17 selected checks, zero failures, zero manual holds)

## Risk and uncovered work

- Existing local configs using `checkout_head`, `repository_scope_root`, or `active_*` fields must migrate to one stable `scope_ref`; the error is explicit rather than silently changing semantics.
- This change does not start or configure a provider watch. Service ownership, watch cadence, and real initial/refresh cost remain operator/provider concerns and require separate runtime evidence.
- Retrieved repository context remains fail-open and advisory; current source and tests remain authoritative.


huangruiteng:
No blocking findings.

- This intentionally breaks local configs using the recently added `checkout_head`, `repository_scope_root`, or `active_*` lifecycle. The new error is explicit and points to a stable `scope_ref`; silently aliasing those fields would be riskier because it would change freshness semantics without notice.
- Public GitHub code search found no call sites for `RepositoryContextRevisionPlan` or `current_revision_not_activated` outside this repository's implementation, smoke, and docs.

- **Motivation:** remove a second revision-activation state machine when the provider already owns a rolling default-branch index.
- **Solved:** the diff makes a stable rolling scope the default, keeps each hit advisory, and preserves current-checkout content verification before any decision influence. It also removes the unused activation module rather than leaving dead compatibility plumbing.
- **Operator impact:** long-running issue-fix agents no longer need to rewrite config or project an active revision after every pull. Explicit pinned corpora still work.
- **Main risk:** migration for the short-lived `checkout_head` config. The fail-fast message and bilingual guide make that boundary visible.
- **Design:** provider refresh ownership and issue-fix checkout verification are now separate reusable contracts; there is no repository-specific branch.

## Validation

- repository-memory, capability-guide, and validated-writeback focused smokes passed;
- workflow-plan, feasibility, repository-context, workflow-contract, and workflow-e2e smokes passed;
- `loopx canary premerge --from-git-diff` passed 17 selected checks with zero failures and zero manual holds;
- public/private boundary scan passed;
- branch was rebased onto current `origin/main` before the final validation run.

## Merge decision

Approved for self-merge after the in-progress repository CI check is green. Runtime watch cadence/cost and long-task service ownership remain separate dogfood work; this PR does not claim to configure or operate the provider watch.

PR #1921, opened 2026-07-11
Make Explore Feishu sync incremental and resumable

huangruiteng:
## Summary
- discover goal-scoped Explore rows across every Feishu Base page
- skip unchanged rows and repair only remote drift
- persist created record IDs immediately so interrupted syncs resume without duplicate creation

## Validation
- `python3 -m py_compile loopx/presentation/sinks/lark/explore_results.py examples/explore-result-layer-smoke.py`
- `python3 -m loopx.cli canary premerge --from-git-diff --tier standard --format json --no-progress` (17 selected checks passed, no warnings or manual holds)

## Safety
- presentation sink only; no worker, quota, lease, launch, or promotion authority changes
- no live credentials or private project evidence in code, fixtures, docs, commit metadata, or this PR

### DIFF
--- a/docs/capabilities/explore/README.md
+++ b/docs/capabilities/explore/README.md
@@ -471,8 +471,11 @@ cooldown guidance.
 
 Record identity follows the Lark Kanban adapter contract: rows are matched by
 the `LoopX Goal ID` + `LoopX Result ID` columns, remembered in the local
-config as `result_records`, and the map is rebuilt from the remote table
-before executed upserts.
+config as `result_records`, and the map is rebuilt from all goal-filtered
+remote pages before executed upserts. Executed sync compares canonical values
+with the remote row and skips unchanged records. Newly created record ids are
+persisted immediately, so an interrupted large-graph sync can resume without
+recreating rows that were already delivered.
 
 The text `From Node` / `To Node` columns remain stable public ids for
 automation and review, while the linked-record columns are the Feishu-native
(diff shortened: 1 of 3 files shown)


huangruiteng:
## Maintainer review

**Findings**

No blocking findings. The sync now bounds discovery to the target goal, walks every 200-row page, compares normalized canonical values before writing, persists newly created record IDs immediately, and leaves both Feishu and local config untouched on a fully unchanged rerun.

**Open assumptions**

- Feishu Base preserves the documented offset/limit and `has_more` contract under the goal filter.
- Concurrent external mutation during one discovery pass remains out of scope; any later drift is repaired on the next sync.
- Pre-existing duplicate rows are reported and the first stable record is reused; cleanup remains an operator action.

Motivation is valid: graph growth should increase read cost gradually, not force hundreds of writes or make retries non-idempotent. The implementation solves the operator problem at the presentation boundary without adding worker, lease, quota, launch, stop, or promotion authority. Goal-filtered pagination plus value reconciliation is a reusable sink contract rather than a project-specific cache patch. The main compatibility risk is Lark value-shape normalization; fixtures cover text, select, number, linked records, nested create receipts, pagination, interruption/resume, zero-write rerun, and one-row remote drift.

**Validation**

- Explore result layer smoke passed, including interruption after one successful create and resume without duplication.
- Lark Kanban control-plane smoke passed.
- Explore schema CLI smoke passed.
- Changed Python compile and diff hygiene passed.
- Standard premerge canary passed twice: 17/17 selected checks, zero failures, zero warnings, zero manual holds; gate reports `self_merge_allowed=true`.
- GitHub pytest passed.
- Public/private boundary and rendered PR diff scans are clean.

**Merge decision**

Self-merge approved under owner authorization and the repository premerge policy.

PR #1931, opened 2026-07-12
Move benchmark run metrics projection inward

huangruiteng:
## Summary
- move round-reward trace and overhead-attribution compactors from `status.py` into a bounded runtime read-model module
- preserve public-safe filtering, numeric call-map normalization, reward-record dedupe, and round ordering
- add direct characterization tests without changing ledger aggregation, scoring, runner behavior, or adapter semantics

## Validation
- full pytest: 97 passed, 2 skipped; 20.38% coverage
- `python3 examples/skillsbench-benchmark-run-smoke.py`
- `python3 examples/benchmark-run-ledger-smoke.py`
- `python3 examples/terminal-bench-harbor-prompt-driven-reducer-smoke.py`
- mypy and focused Ruff: passed
- LoopX boundary scan: 0 errors, 0 warnings
- premerge canary: 16/16 selected checks passed; no manual holds

No benchmark jobs were launched and no public API, persisted state, scoring, or runner behavior changed.

PR #1933, opened 2026-07-12
Require registered task lease owners

huangruiteng:
## Summary
- require task lease acquire/renew/transfer owners to be registered peers for the goal
- treat legacy active leases owned by unregistered identities as ineffective during inspect and conflict checks
- preserve release as a cleanup path for legacy records
- add CLI regression coverage for unregistered acquire, transfer, and legacy inspect

## Root cause
`task-lease --owner` documented a registered agent id, but the runtime only normalized the token and checked `excluded_agents`. Any public-safe unregistered token could therefore acquire a hard lease and block legitimate peers.

## Validation
- `uv run --with pytest --with pytest-cov pytest -q`: 97 passed, 2 skipped
- bounded-context, risk-characterization, and Python line-budget smokes
- `loopx canary premerge --from-git-diff`: 14 selected, 0 failures, 0 manual holds

## Baseline note
`examples/control_plane/todo-readmodel-boundary-smoke.py` currently fails unchanged on `origin/main` because it still expects the removed `status.blank_event_class_counts` alias from #1931. This PR does not modify that unrelated read-model contract.

## Risk
Bounded to opt-in task lease identity validation. Quota does not consume or enforce task leases, and todo soft-claim routing remains unchanged.

### DIFF
--- a/examples/control_plane/task-lease-runtime-smoke.py
+++ b/examples/control_plane/task-lease-runtime-smoke.py
@@ -161,6 +161,48 @@ def main() -> int:
         assert first["lease"]["schema_version"] == "task_lease_v0", first
         assert first["lease"]["version"] == 1, first
 
+        unregistered_acquire = cli(
+            registry_path,
+            "acquire",
+            "--goal-id",
+            GOAL_ID,
+            "--todo-id",
+            TODO_C,
+            "--owner",
+            "codex-unregistered-peer",
+            "--idempotency-key",
+            "unregistered-acquire",
+            "--write-scope",
+            "scripts/**",
+            check=False,
+        )
+        assert unregistered_acquire.returncode == 1, unregistered_acquire.stdout
+        assert payload(unregistered_acquire)["error_code"] == "owner_not_registered", (
+            unregistered_acquire.stdout
+        )
+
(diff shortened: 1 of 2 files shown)


huangruiteng:

- **Identity invariant:** acquire and renew validate the acting owner through the canonical registered-agent helper; transfer validates both current and target owners.
- **Migration behavior:** inspect and cross-todo conflict evaluation ignore legacy active records whose owner is no longer registered. Release remains available for cleanup, so invalid legacy records do not become undeletable.
- **Scope:** no quota integration, soft-claim routing, lease TTL/CAS, write-scope overlap, or todo lifecycle behavior changed.
- **Validation:** GitHub pytest passed; local pytest 97 passed / 2 skipped; task-lease lifecycle smoke passed; premerge 14/14 with no manual holds; bounded-context and public-boundary checks passed.
- **Known unrelated baseline:** `todo-readmodel-boundary-smoke.py` fails unchanged on main due to the stale post-#1931 status alias assertion and is not hidden or modified here.

The focused coverage exercises the original bad case, transfer, legacy state, and the complete registered-owner lifecycle, so admin self-merge is justified for this bounded control-plane correction.

PR #1935, opened 2026-07-12
Stop advertising TTL for soft claims

huangruiteng:
## Summary
- stop writing inert `coordination.claim_ttl_minutes` into newly bootstrapped goals
- keep legacy `--claim-ttl-minutes` accepted but hide and ignore it for script compatibility
- remove the false TTL from public registry examples
- document that soft `claimed_by` does not expire and explicit `task-lease --ttl-seconds` owns hard TTL semantics
- preserve old registry fixtures to keep legacy state readable

## Root cause
`claim_ttl_minutes` was emitted by bootstrap but never read anywhere in runtime selection, claim, status, or quota code. It implied that soft claims expired even though they do not, duplicating and contradicting the explicit hard-lease model.

## Validation
- full pytest: 97 passed, 2 skipped
- todo claim/lease roadmap smoke, including legacy option compatibility and hidden help
- demo CLI, bootstrap preserve-todos, connect route collision, and auto-research goal-isolation smokes
- `loopx canary premerge --from-git-diff`: 17 selected, 0 failures, 0 manual holds

## Compatibility
Existing registries containing `claim_ttl_minutes` remain readable. Existing scripts passing the CLI flag continue to succeed. Only new bootstrap output stops persisting the inert field.

### DIFF
--- a/docs/frontstage-channel-lease-roadmap.md
+++ b/docs/frontstage-channel-lease-roadmap.md
@@ -24,8 +24,9 @@ truth into chat history:
 - An **agent can project as a workspace member**: controller, executor,
   reviewer, monitor, critic, or dreaming/planning proposer, each with scope and
   last action.
-- A **task claim should be a per-todo lease**: explicit ownership of one
-  `todo_id` with TTL, write scope, idempotency key, and conflict policy.
+- A **task claim is a soft per-todo route by default**. When a concrete
+  contention case needs exclusivity, an optional hard lease adds TTL, write
+  scope, idempotency, and conflict handling for one `todo_id`.
 - A **chat or channel thread is a projection**: useful for human collaboration,
   but never the only durable authority.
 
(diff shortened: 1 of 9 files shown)


huangruiteng:

- **Behavioral correction:** new bootstrap output no longer persists `claim_ttl_minutes`, which had no runtime reader and therefore never expired a soft claim.
- **Compatibility:** the legacy CLI flag remains accepted but hidden/ignored; old registries with the field remain readable and are still represented in migration/routing fixtures.
- **Ownership model:** `claimed_by` stays a non-expiring soft routing signal. TTL, CAS, overlap, transfer, and release remain isolated in the opt-in hard `task-lease` runtime.
- **Scope:** no quota, todo selection, status, lease lifecycle, or existing-state migration behavior changed.
- **Validation:** GitHub pytest passed; local pytest 97/2; bootstrap/demo/auto-research/legacy-route smokes passed; premerge 17/17 with no holds; public-boundary scan clean.

The change removes false state rather than adding another transition path and is suitable for authorized self-merge.

PR #1936, opened 2026-07-12
Invalidate leases for non-open todos

huangruiteng:
## Summary
- make a task lease effective only while its canonical todo status is `open`
- reject acquire/renew/transfer against done, blocked, deferred, or otherwise non-open todos
- make inspect and overlapping-scope conflict evaluation ignore stale leases from non-open todos
- preserve explicit release for cleanup and audit

## Root cause
Lease eligibility checked todo existence, registered ownership, exclusions, and TTL, but not todo lifecycle status. A completed todo could therefore retain an active hard lease and block overlapping work until TTL expiry.

## Validation
- installed bad-case reproduction before the fix: completed todo still inspected as active
- task-lease runtime smoke now covers done inspect, done renew, deferred acquire, and reopening
- full pytest: 97 passed, 2 skipped
- bounded-context, risk-characterization, and line-budget smokes
- `loopx canary premerge --from-git-diff`: 14 selected, 0 failures, 0 manual holds

## Risk
One central eligibility rule changes only opt-in hard leases. Soft claims, quota, todo lifecycle writes, TTL/CAS behavior, and release remain unchanged.

### DIFF
--- a/examples/control_plane/task-lease-runtime-smoke.py
+++ b/examples/control_plane/task-lease-runtime-smoke.py
@@ -85,6 +85,17 @@ def set_excluded_agents(state_file: Path, *, todo_id: str, agents: list[str]) ->
     raise AssertionError(f"todo metadata not found: {todo_id}")
 
 
+def set_todo_status(state_file: Path, *, todo_id: str, status: str) -> None:
+    lines = state_file.read_text(encoding="utf-8").splitlines()
+    for index, line in enumerate(lines):
+        if f"todo_id={todo_id}" not in line:
+            continue
+        lines[index] = re.sub(r"\bstatus=[^\s<>]+", f"status={status}", line)
+        state_file.write_text("\n".join(lines) + "\n", encoding="utf-8")
+        return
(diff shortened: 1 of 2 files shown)


huangruiteng:

- **Lifecycle invariant:** a hard lease is effective only while the canonical todo is `open`. The same helper is already used by acquire/renew/transfer, inspect, existing-lease evaluation, and overlap conflict scans.
- **Cleanup:** lease files are not silently deleted; `release` remains available even after terminal transition.
- **Compatibility:** reopening a todo restores normal evaluation of an unexpired valid lease, while done/deferred work cannot renew or acquire execution ownership.
- **Scope:** 12 product-code lines plus focused lifecycle coverage; no soft-claim, quota, todo-write, TTL/CAS, or scheduler changes.
- **Validation:** GitHub pytest passed; local pytest 97/2; task-lease smoke and bounded-context checks passed; premerge 14/14 with no holds; public boundary clean.

This directly closes a reproduced stale-lock bad case and is suitable for authorized self-merge.

PR #1920, opened 2026-07-11
Retire duplicate status runtime wrappers

huangruiteng:
## Summary
- remove three unused `loopx.status` wrappers whose production composition already lives in `control_plane.status_runtime_summaries`
- remove accidental read-model re-exports used only by parity smokes
- migrate the smokes to exercise the real `StatusRuntimeSummaryContext` composition path

This is a behavior-preserving cleanup: 53 insertions, 74 deletions, net -21 lines. The remaining SkillsBench verifier-bootstrap enrichment is intentionally unchanged because it still carries domain-specific score-attribution semantics.

## Validation
- `python -m ruff check tests loopx/canary loopx/control_plane loopx/domain_packs loopx/presentation`
- `python -m mypy`
- `pytest -q -n 2 --cov=loopx --cov-fail-under=19.6` (83 passed, 2 skipped; 19.80%)
- `loopx canary premerge --from-git-diff` (17/17 passed; no manual holds; self-merge allowed)
- `loopx check` public boundary scan (3 files, 0 warnings/errors)

## Risk
Low. No CLI/schema/persisted-state change; the deleted symbols had no production or documented callers. The smokes now validate the actual production composition path.

PR #1937, opened 2026-07-12
Fix retired status alias smoke

huangruiteng:
## Summary
- stop the aggregate todo read-model smoke from requiring event-ledger aliases retired by #1920
- keep the dedicated bounded-context event-ledger and decision-freshness smokes as the owners of that contract

## Validation
- `python3 examples/control_plane/todo-readmodel-boundary-smoke.py`
- Full Public Smokes shard 3: 100/100 passed
- `loopx canary premerge --from-git-diff`: 5/5 selected checks passed, no holds

Closes the uncovered main CI regression observed in run 29175301442.

### DIFF
--- a/examples/control_plane/todo-readmodel-boundary-smoke.py
+++ b/examples/control_plane/todo-readmodel-boundary-smoke.py
@@ -25,7 +25,6 @@ from loopx.control_plane.goals import global_registry_shadow as global_registry_
 from loopx.control_plane.goals import path_resolution as path_resolution_read_model  # noqa: E402
 from loopx.control_plane.agents import management_projection as management_projection_read_model  # noqa: E402
 from loopx.control_plane.runtime import agent_scoped_evidence_log as evidence_log_read_model  # noqa: E402
-from loopx.control_plane.runtime import event_ledger as event_ledger_read_model  # noqa: E402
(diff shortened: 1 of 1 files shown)


huangruiteng:
Self-review / validation packet:

- Changed surface: one aggregate public smoke only; no runtime/API behavior.
- Root cause: #1920 intentionally retired `loopx.status` event-ledger wrappers, but this aggregate smoke retained two alias assertions already superseded by dedicated event-ledger and decision-freshness smokes.
- Exact regression: `python3 examples/control_plane/todo-readmodel-boundary-smoke.py` passed after the deletion.
- Full failing shard: `python3 examples/run-smokes.py --suite full-public --offset 300 --limit 100 --timeout-seconds 120 --jobs 4 --json` passed 100/100, with 0 failures, timeouts, skips, warnings, or tracked side effects.
- Premerge: `loopx canary premerge --from-git-diff` passed all 5 selected checks, including module budget, todo contract, active-state projection, and public boundary; 0 failures and 0 manual holds.
- Ruff was unavailable in this shell; the diff only deletes one import and two assertions, while py_compile and all selected smokes passed.

Coverage is sufficient because the change removes only stale compatibility expectations and reruns both the exact regression and the complete failing CI shard.

PR #1938, opened 2026-07-12
Fix task lease scope overlap matching

huangruiteng:
## Summary
- fix hard-lease scope overlap checks for glob scopes versus concrete child paths
- preserve non-overlap for distinct exact sibling files
- add a durable runtime regression smoke for the original false-negative case

## Root cause
The previous implementation normalized both scopes to directory roots and only checked strict ancestor prefixes. That made `docs/**` and `docs/a.md` collapse to equal roots without being treated as overlapping.

## Validation
- `uv run --no-project --with pytest --with pytest-cov pytest -q` (97 passed, 2 skipped)
- bounded-context, control-plane-risk, and Python line-budget smokes
- `loopx check` on both changed paths (0 errors, 0 warnings)
- `loopx canary premerge --from-git-diff` (standard tier, 14 selected, 0 failures, 0 manual holds)

### DIFF
--- a/examples/control_plane/task-lease-runtime-smoke.py
+++ b/examples/control_plane/task-lease-runtime-smoke.py
@@ -13,6 +13,11 @@ from typing import Any
 
 
 REPO_ROOT = Path(__file__).resolve().parents[2]
+sys.path.insert(0, str(REPO_ROOT))
+
+from loopx.control_plane.work_items.task_lease import write_scopes_overlap  # noqa: E402
+
+
 GOAL_ID = "task-lease-runtime-goal"
 TODO_A = "todo_taskleasea"
 TODO_B = "todo_taskleaseb"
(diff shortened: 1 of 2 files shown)


huangruiteng:
Self-review complete: no blocking findings.

Reviewed boundaries:
- the original false negative (`glob/**` versus a concrete child path) is covered through the CLI runtime smoke;
- distinct exact sibling files remain non-overlapping, so safe peer parallelism is preserved;
- glob-vs-glob matching is intentionally conservative for hard leases;
- no changes to soft claims, quota, todo lifecycle, lease TTL/renewal/preemption, permissions, or persisted schemas.

Validation: 97 passed, 2 skipped; focused task-lease smoke; contract check 0 errors/0 warnings; standard premerge gate 14/14 passed with no manual holds; GitHub pytest passed.

PR #1943, opened 2026-07-12
Fix rollout event occurrence identity

huangruiteng:
## Summary

- include `recorded_at` in deterministic rollout event identity so repeated observations at different times remain distinct audit events
- make the shared append boundary idempotent for exact event replay and fail closed on conflicting `event_id` reuse under the existing file lock
- add focused coverage for exact replay, later equivalent observations, and conflicting reuse

## Design boundary

Raw rollout history keeps distinct timestamped observations. Agent-facing trajectory compaction remains a projection concern; this change does not silently collapse or delete audit events.

## Validation

- `python3 examples/project/goal-rollout-event-log-smoke.py`
- `python -m ruff check ...`
- `python -m mypy`
- `python -m pytest -q -n 2 --cov=loopx --cov-report=term --cov-fail-under=19.6` (97 passed, 2 skipped; 20.31%)
- `loopx canary premerge --from-git-diff` (17/17 selected checks passed; no warnings or manual holds)
- public boundary scan clean

## Risk

Low and localized to rollout-event identity/append semantics. Existing explicit timestamps remain deterministic for retry idempotency; generated timestamps now retain microseconds to avoid same-second occurrence collisions.

### DIFF
--- a/examples/project/goal-rollout-event-log-smoke.py
+++ b/examples/project/goal-rollout-event-log-smoke.py
@@ -133,6 +133,38 @@ def main() -> None:
         runtime_root = tmp_root / "runtime"
         goal_id = "loopx-meta"
         log_path = rollout_event_log_path(runtime_root, goal_id)
+        identity_log_path = tmp_root / "rollout-event-identity.jsonl"
+        first_occurrence = build_rollout_event(
+            goal_id=goal_id,
+            event_kind="validation",
+            status="passed",
+            summary="Focused validation passed.",
+            recorded_at="2026-07-12T00:00:00Z",
+        )
+        append_rollout_event(identity_log_path, first_occurrence)
+        replayed = append_rollout_event(identity_log_path, dict(first_occurrence))
+        assert replayed == first_occurrence, replayed
+        assert len(load_rollout_events(identity_log_path)) == 1
+        later_occurrence = build_rollout_event(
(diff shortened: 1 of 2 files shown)


huangruiteng:
Self-review

**Assumption:** `recorded_at` is the occurrence boundary for rollout observations. Replaying an event with the same timestamp and normalized body is the same occurrence; a later observation is a new occurrence even when all other fields match.

**Product / architecture judgment:** The motivation is to restore unambiguous rollout lineage without sacrificing audit cadence. The shared append boundary now matches the event-sourced contract: exact replay is idempotent and conflicting ID reuse fails closed. Including `recorded_at` in identity preserves later equivalent observations, while Agent-facing noise reduction remains in trajectory-hygiene and TurnEnvelope projections. This is preferable to write-side semantic deduplication, which would erase legitimate quota polls and spends. Main risk is hot-path append latency; measured lookup against the current 15 MB log at about 5.7 ms, and the sibling file lock prevents concurrent duplicate appends.

**Validation:** focused rollout and PR-lifecycle smokes passed; ruff and mypy passed; pytest 97 passed / 2 skipped with 20.31% coverage; premerge canary selected 17 checks and all passed; public-boundary scan clean; no manual holds.

**Merge decision:** self-merge after the required Python Tests check completes successfully.

PR #1944, opened 2026-07-12
Prevent task leases from stealing soft claims

huangruiteng:
## Summary

- make optional hard-lease ownership respect an existing todo `claimed_by`
- mark legacy active leases ineffective when their owner conflicts with the current soft claim
- fail closed for acquire/renew/transfer through the shared owner constraint
- cover matching claim acquisition, conflicting active-lease inspection, and blocked transfer

## Product boundary

This does not make task leases mandatory or add lease enforcement to `quota should-run`. `claimed_by` remains the default soft route; the opt-in hard lease can only strengthen that ownership, not silently contradict it.

## Validation

- ruff and mypy passed
- pytest: 97 passed, 2 skipped; 20.31% coverage
- premerge canary: 14/14 selected checks passed, no warnings or manual holds
- public boundary scan clean

## Risk

Localized fail-closed behavior change for callers that previously acquired or transferred a lease to an agent different from the todo soft claim. Such callers must explicitly clear or transfer the claim first.

### DIFF
--- a/examples/control_plane/task-lease-runtime-smoke.py
+++ b/examples/control_plane/task-lease-runtime-smoke.py
@@ -90,6 +90,20 @@ def set_excluded_agents(state_file: Path, *, todo_id: str, agents: list[str]) ->
     raise AssertionError(f"todo metadata not found: {todo_id}")
 
 
+def set_claimed_by(state_file: Path, *, todo_id: str, owner: str | None) -> None:
+    lines = state_file.read_text(encoding="utf-8").splitlines()
+    for index, line in enumerate(lines):
+        if f"todo_id={todo_id}" not in line:
+            continue
+        line = re.sub(r"\s+claimed_by=[^\s<>]+", "", line)
+        if owner:
+            line = line.replace(" -->", f" claimed_by={owner} -->")
+        lines[index] = line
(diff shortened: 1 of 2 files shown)


huangruiteng:
Self-review

**Assumption:** A hard lease may strengthen an unclaimed todo or a matching soft claim, but it must not contradict `claimed_by`. Explicit reassignment should move/clear the soft claim before acquiring or transferring the lease.

**Product / architecture judgment:** This fixes a real split-brain condition: quota can route a claimed todo to peer A while the optional hard lease is owned by peer B. The change belongs in the shared owner constraint, so acquire, renew, transfer, inspect, and conflict scans agree without coupling leases into quota. It preserves the current product boundary that leases are opt-in and only justified by demonstrated concurrent-write contention. Main compatibility risk is a caller that intentionally used a hard lease to override another peer's claim; that workflow now fails closed and must explicitly reassign the claim.

**Validation:** task-lease runtime smoke passed; ruff/mypy passed; pytest 97 passed / 2 skipped at 20.31%; premerge canary 14/14 with no warnings or manual holds; public-boundary scan clean.

**Merge decision:** self-merge after Python Tests passes.

PR #1946, opened 2026-07-12
docs(issue-fix): record rolling index dogfood result

huangruiteng:
## Motivation

The issue-fix capability guide described rolling-main repository memory and its policy, but did not record the first fresh-issue measurement. That left readers unable to distinguish retrieval volume from verified decision value.

## Implementation

- add the public OpenViking issue #3152 / PR #3176 dogfood case to both guides
- record the measured positive-but-mixed result: useful source/test and validation locators, weak causal retrieval, zero memory patch authority
- pin the checkout revision used for verification and keep rolling retrieval optional/fail-open
- update the bilingual guide smoke so the public case, result label, and revision remain present

## Validation

- `loopx check --scan-path docs/capabilities/issue-fix/README.md --scan-path docs/capabilities/issue-fix/README.zh-CN.md --scan-path examples/issue-fix-capability-guide-smoke.py`
- `loopx canary premerge --from-git-diff` (17 selected checks, 0 failures)
- `git diff --check`

## Risk and uncovered items

Docs and one durable documentation smoke only; no runtime, provider, indexing, authority, or default-enablement behavior changes. The result covers one fresh issue, so the guide explicitly keeps packaged-default enablement undecided until independent repetitions provide stronger evidence.


huangruiteng:
## Findings

No blocking findings. The issue/PR links and pinned revision are public, the English and Chinese claims agree, and the smoke prevents the measured result from silently disappearing.

The evidence covers one fresh issue only. The docs correctly keep rolling-main retrieval optional and fail-open rather than treating this result as packaged-default proof.

**Motivation:** distinguish useful repository-memory decision support from raw retrieval volume.

**Solved:** yes for the documentation gap. The guide now records which decisions changed (`change_scope`, `validation`), which did not (patch), the verifying checkout revision, and the repeated-evidence threshold for default enablement.

**Impact:** readers get a concrete, auditable example instead of a mechanism-only promise.

**Risk:** low. This changes docs and one documentation contract smoke only; no provider, watch, indexing, writeback, authority, or installed runtime behavior changes.

**Design:** right-sized. The result is attached to the existing generic repository-memory contract rather than creating another state machine or OpenViking-specific runtime branch.

## Validation

- public-boundary `loopx check` on all three changed files
- `loopx canary premerge --from-git-diff`: 17 selected checks, 0 failures, 0 warnings
- `git diff --check`

## Merge decision

Self-merge approved under the standing owner authorization: small, single-purpose, public-safe docs plus a focused smoke, with no first-screen README change and no runtime or permission-sensitive behavior.

PR #1947, opened 2026-07-12
feat(agents): route quota with advisory profiles

huangruiteng:
## Summary
- add validated `agent_profile_v1` write/clear support to `configure-goal` and the loopback control-plane API
- project valid advisory profiles into quota agent identity
- rank preferred/avoided action kinds inside existing active-next and claim buckets without changing hard eligibility
- document the contract and cover configuration, compatibility, identity, claim ordering, and final quota routing

This makes the existing profile contract operational instead of prompt-only. The profile stays advisory: it cannot transfer claims, bypass leases/gates/capabilities/write scopes, or create peer hierarchy. One shared rank helper is reused by todo projection, capability candidates, and final lane selection so the read model and selected action cannot drift. Invalid legacy advisory metadata is ignored on read, while new writes are strict and bounded.

## Validation
- `python -m pytest -q -n auto`: 97 passed, 2 skipped
- focused agent identity, todo claim visibility, configure-goal, and peer runtime smokes: passed
- targeted `ruff`: passed
- `loopx canary premerge --from-git-diff`: 17/17 passed, 0 failures, 0 warnings, no manual holds
- public/private boundary scan: clean across 12 changed files

The first final canary attempt correctly caught an old hierarchy literal in the new validator; the implementation was rewritten to preserve the hard-cut boundary and the complete gate then passed.

## Excluded
No local state, raw trajectories/logs, credentials, private links, benchmark artifacts, or production actions.

### DIFF
--- a/docs/product/agent-profile-contract.md
+++ b/docs/product/agent-profile-contract.md
@@ -55,6 +55,19 @@ claim or lease.
 }
 ```
 
+Profiles are written through the same validated registry path as other goal
+coordination settings:
+
+```bash
+loopx configure-goal --goal-id <goal-id> \
+  --agent-profile-json '{"schema_version":"agent_profile_v1","agent_id":"codex-runtime","profile_role":"runtime-validation","preferred_action_kinds":["runtime_*"]}' \
+  --execute
+```
+
+Use `--clear-agent-profile <agent-id>` to remove one profile. The command
+rejects unregistered peers, unknown fields, hierarchy roles, invalid task
+classes, and unsafe action-kind globs before writing the registry.
+
 `profile_role` is a human-readable functional label. It is advisory and must
 not use hierarchy labels such as `primary-agent` or `side-agent`.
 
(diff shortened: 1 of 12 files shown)


huangruiteng:
## Self-review

**Findings:** No blocking findings. The implementation keeps `agent_profile_v1` advisory and preserves active-next, explicit claim, lease, gate, capability, write-scope, continuation, and workspace authority.

**Assumptions:** Profile matching is intentionally a ranking hint, not an eligibility filter. Invalid pre-existing profile metadata is ignored on the read path so optional advisory state cannot strand quota; new writes remain strict and bounded.

**Product / architecture judgment:** This closes the gap where recurring scope existed in heartbeat prose but not quota truth. A shared profile normalizer/rank helper avoids three divergent implementations across todo projection, capability gate, and final lane recommendation. No new state machine or identity hierarchy is introduced.

**Validation:** targeted ruff; 97 passed / 2 skipped pytest; focused configuration, identity, claim-routing, and peer-runtime smokes; final premerge 17/17 with clean public boundary, 0 warnings, and no manual holds. An initial hard-cut failure found a hierarchy literal in the validator; it was removed structurally and the full gate passed afterward.

**Merge decision:** Approved for owner-authorized self-merge after the GitHub pytest check completes (or admin bypass if it remains infrastructure-only), because the change is single-purpose, public-safe, locally fully validated, and has no benchmark, permission, production, or evidence-policy change.

PR #1948, opened 2026-07-12
feat(projection): reconcile complete source snapshots

huangruiteng:
## Motivation

Normal projection sync is intentionally additive, but operators had no safe generic way to retire rows after a complete source stopped emitting them. The real failure mode left 17 manually deleted remote rows with stale local record mappings; retrying normal sync could preserve or recreate drift.

## Implementation

- add provider-neutral planning for one exact `goal_id + source_id` projection namespace
- add `sync-projection --reconcile-source --source-snapshot-complete`
- keep ordinary sync non-destructive
- require unfiltered input, `--include-done`, no row-limit truncation, matching source id, and a complete remote record-list page
- preview remote orphan record ids and stale local mapping keys before writes
- on explicit `--execute`, delete only confirmed namespace records with the Lark high-risk `--yes` gate, then persist mapping cleanup
- keep retries idempotent and other source namespaces untouched

## Validation

- red-first `python3 examples/lark-projection-source-reconcile-smoke.py`
- `python3 examples/lark-kanban-control-plane-smoke.py`
- `python3 examples/canary/catalog-planner-smoke.py`
- Python compile and `uvx ruff check` on changed Python
- public-boundary `loopx check` on all 10 changed files
- `loopx canary premerge --from-git-diff`: 17 selected, 0 failures, 0 warnings, 0 manual holds
- `git diff --check`

## Risk and uncovered items

This adds an explicitly gated destructive path. It never runs from default sync and refuses filtered, truncated, source-mismatched, or remotely paginated input. Remote listing is currently bounded to the existing 200-row record-list command; a larger board fails closed instead of auto-paginating. No real Base rows were deleted by this PR.


huangruiteng:
## Findings

No blocking findings after reviewing the source planner, Lark record I/O boundary, CLI gates, focused smoke, and public docs. The destructive command cannot be reached from default sync; reconcile requires an explicit complete-snapshot attestation, `--include-done`, an unfiltered and untruncated projection, a matching source namespace, a complete remote list, and `--execute`.

The existing remote list command is capped at 200 rows. This PR intentionally fails closed when `has_more` is true instead of adding pagination in the same change. Empty complete snapshots may retire the entire exact source namespace, but only after preview and the same explicit execute gates.

**Motivation:** remove real projection drift without turning ordinary sync into implicit bulk deletion.

**Solved:** yes for the bounded contract. It previews remote orphan records and stale local mappings, executes deletion only inside `goal_id + source_id`, preserves other namespaces, and makes retry zero-write.

**Impact:** operators can safely reconcile stale Monthly Impact or other shared projection rows after inspecting a concrete plan.

**Main risk:** high-risk external deletion. The implementation contains that risk behind fail-closed completeness checks and Lark’s explicit `--yes` confirmation; no real Base deletion was performed during validation.

**Design:** the planner is provider-neutral and Lark record parsing/delete command construction is isolated from the already-large Kanban module. The line-budget smoke remains green; no repository-specific branch was added.

## Validation

- red-first focused source-reconcile smoke
- existing Lark Kanban control-plane smoke
- issue-fix bilingual guide smoke
- canary catalog planner and Python line-budget smokes
- Python compile and `uvx ruff check`
- public-boundary scan across all 10 changed files
- final premerge canary: 17 selected, 0 failures, 0 warnings, 0 manual holds
- GitHub pytest: passed

## Merge decision

Self-merge approved under the standing owner authorization for LoopX PRs after the above high-risk review. The command remains opt-in and fail-closed; broader pagination is explicitly not bundled.

PR #1950, opened 2026-07-12
Harden concurrent release promotion

huangruiteng:
## Motivation

Concurrent local promotions currently derive the same second-level release id. One installer can reuse or collide with another installer snapshot, while the default symlink is reported successful without proving that it points to the source revision requested by this install.

## Implementation

- serialize local promotion writes with a recoverable process lock
- allocate a unique suffix when a requested release id already exists
- atomically replace executable symlinks and verify the promoted target and manifest before success
- resolve default GitHub refs to a full commit SHA, download that immutable archive, and preserve the SHA in release provenance
- retain custom archive URL compatibility

## Validation

- focused concurrency regression: two same-id installers produce distinct manifests and a valid final symlink
- install-local overwrite smoke
- install-local full smoke
- no-clone archive verification smoke
- catalog planner smoke
- shell syntax and Python compile checks
- loopx canary premerge --from-git-diff: 17 selected checks passed, 0 failures, 0 warnings, 0 manual holds
- public boundary scan: clean

## Risk and gaps

The local installer now serializes the shared promotion section, so a second concurrent install waits rather than racing. A live process can hold the lock for up to 60 seconds; dead process locks with a recorded PID are recovered. Custom archive URLs remain usable but only carry commit provenance when the caller supplies LOOPX_RESOLVED_SOURCE_GIT_COMMIT.

### DIFF
--- a/examples/release/release-promotion-concurrency-smoke.py
+++ b/examples/release/release-promotion-concurrency-smoke.py
@@ -1,0 +1,108 @@
+#!/usr/bin/env python3
+"""Regression coverage for collision-safe, revision-verifiable promotions."""
+
+from __future__ import annotations
+
+import json
+import os
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+INSTALL_SCRIPT = REPO_ROOT / "scripts" / "install-local.sh"
+
+if str(REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(REPO_ROOT))
+
+from loopx.release_manifest import build_release_manifest  # noqa: E402
+
+
+def install_env(root: Path) -> dict[str, str]:
+    home = root / "home"
+    return {
+        **os.environ,
+        "HOME": str(home),
+        "CODEX_HOME": str(home / ".codex"),
+        "LOOPX_BIN_DIR": str(home / ".local" / "bin"),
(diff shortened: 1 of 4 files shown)


huangruiteng:
## Review

### Findings

No blocking findings after the lock-order refinement. Shared default-install writes are now inside one recoverable local critical section; existing release ids are never reused; executable links are replaced atomically and verified before success.

### Open assumptions

- The default GitHub path intentionally adds one unauthenticated commit-resolution request before the archive download. This makes ref-to-commit provenance explicit, at the cost of GitHub API availability/rate limits.
- Custom archive URLs remain compatible and only claim commit provenance when the caller supplies a full resolved SHA.

The motivation is a real installed-user control-plane risk: concurrent self-promotions could share a second-level id or publish a different snapshot than the caller intended. The implementation solves that at the release boundary rather than adding caller-specific retries. The main compatibility risk is the extra GitHub API dependency, but it is confined to the default GitHub channel and buys immutable download plus auditable provenance. A process lock is appropriate for this local single-host installer; it also recovers a dead lock when the recorded PID no longer exists.

### Validation

- focused concurrent same-id regression: passed
- install-local overwrite and full smokes: passed
- packaged/no-clone archive smoke: passed
- update/catalog, shell syntax, Python compile, diff hygiene: passed
- `loopx canary premerge --from-git-diff`: 17/17 passed, no warnings or manual holds

### Merge decision

Approved for self-merge once the refreshed GitHub CI run is green. The change is single-purpose, has focused and broad install coverage, and does not alter permissions, benchmark semantics, or public evidence policy.

PR #1952, opened 2026-07-12
Preserve immutable update source resolution

huangruiteng:
## Motivation

PR #1950 made the default GitHub installer resolve refs to immutable commits, but real `loopx update --ref main` still exported the planner-generated codeload URL as if it were a user override. That bypassed commit resolution and left release provenance with a null source commit.

## Implementation

- pass LOOPX_ARCHIVE_URL only for the explicit custom-archive channel
- remove stale ambient archive overrides from default repo/ref execution
- keep the computed archive URL in the read-only plan for operator visibility
- cover both dry-run command rendering and execution environment behavior

## Validation

- focused loopx update smoke: passed
- Python compile and diff hygiene: passed
- loopx canary premerge --from-git-diff: 5/5 passed, no warnings or manual holds
- install-local and packaged install canaries: passed
- public boundary scan: clean

## Risk and follow-up

This is a narrow planner-to-installer contract correction. Explicit custom archive URLs are unchanged. After merge, the acceptance check is a real default update whose release manifest source commit must equal the merge commit.

### DIFF
--- a/examples/loopx-update-smoke.py
+++ b/examples/loopx-update-smoke.py
@@ -19,6 +19,7 @@ from loopx import __version__  # noqa: E402
 from loopx.release_manifest import release_version_tag  # noqa: E402
 from loopx.self_update import (
     DEFAULT_UPDATE_REF,
+    _installer_env_for_source,
     build_rollback_plan,
     build_update_plan,
     execute_rollback_plan,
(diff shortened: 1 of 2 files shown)


huangruiteng:
## Review

### Findings

No blocking findings. The regression is at the correct seam: the update planner may display a computed archive URL, but only an explicit custom-archive channel may export LOOPX_ARCHIVE_URL to the installer.

### Open assumptions

The installer remains the owner of ref-to-commit resolution for default GitHub channels. The update planner owns repo/ref selection and operator preview, not immutable resolution itself.

This fixes a real cross-module contract gap found by the post-merge callsite for #1950. A small environment builder is preferable to another flag or state field because channel already carries the necessary distinction. Removing stale ambient overrides also prevents an older shell variable from silently changing a default update.

### Validation

- focused update contract smoke: passed
- default command rendering and execution env both omit generated archive override
- explicit custom archive still reaches installer
- `loopx canary premerge --from-git-diff`: 5/5 passed
- public boundary scan: clean

### Merge decision

Approved for self-merge once CI is green. Final acceptance remains a post-merge real default update whose manifest commit equals the merge commit.

PR #1956, opened 2026-07-12
Guard rolling context provider sync ownership

huangruiteng:
## Summary

- require a provider-neutral persistent service ownership receipt before an explicit rolling context-provider sync can write
- compare service generation and process identity around the call so a restart becomes `restart_detected_no_resume` instead of false resumed progress
- preserve completed/pending writes and elapsed time as an additional attempt, while keeping retrieval, provider-owned watch, and pinned sync behavior unchanged
- document the contract in the bilingual issue-fix guide and register focused canary coverage

## Validation

- `python3 examples/context-provider-service-ownership-smoke.py`
- `uvx ruff check` on all touched Python paths
- `python3 -m loopx.cli canary premerge --from-git-diff` (17/17 checks passed)
- real OpenViking CLI call: missing receipt blocked before provider invocation with zero writes; a live persistent-service receipt completed the same rolling sync with verified ownership and no restart

## Risk and compatibility

The intentional behavior change is limited to `rolling_default_branch` syncs executed with external writes enabled. They now fail closed when no live persistent-service receipt is available. LoopX does not launch or supervise the provider, and public packets never include the receipt path or process id. Retrieval, provider-owned refresh/watch, and pinned syncs keep their existing behavior.

### DIFF
--- a/docs/capabilities/issue-fix/README.md
+++ b/docs/capabilities/issue-fix/README.md
@@ -707,6 +707,7 @@ issue-fix caller for verification; it is not encoded into the provider scope:
   "timeout_seconds": 15,
   "sync_timeout_seconds": 180,
   "resource_references": ["src/module.py", "tests/test_module.py"],
+  "service_ownership_receipt_path": ".loopx/context-provider-service.json",
   "writeback_enabled": false,
   "writeback_scope_ref": "viking://resources/public-repository/owner-repo/outcomes/<git-revision>",
   "workspace_scope": "owner-repo",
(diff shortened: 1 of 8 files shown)


huangruiteng:
## Findings

No blocking findings in the current diff.

## Open assumptions

- A persistent external service or supervisor owns and refreshes the local receipt; LoopX intentionally validates the receipt but does not become a process manager.
- Requiring that receipt is an intentional fail-closed compatibility change only for explicit `rolling_default_branch` sync writes. Retrieval, provider-owned watch, and pinned sync remain unchanged.

The change addresses the observed failure mode at the reusable provider boundary: a long write must not inherit the lifetime of an ephemeral agent exec session or claim resumed progress after the provider restarts. The receipt module is provider-neutral, the issue-fix layer only applies it at the active rolling-sync call site, and public packets expose opaque refs instead of local paths or process ids. The design is narrower than having LoopX launch or supervise OpenViking and keeps existing read paths compatible.

## Validation

- focused service-ownership smoke: passed
- issue-fix repository-memory integration smoke: passed
- bilingual capability-guide smoke: passed
- Ruff on all touched Python paths: passed
- real OpenViking call: missing receipt blocked with zero writes; live receipt completed with verified ownership and no restart
- risk-based premerge canary: 17/17 passed, including line budget and public-boundary checks

## Merge decision

Hold only until the required GitHub `pytest` check finishes. If it passes and the branch remains mergeable, this is approved for the owner-authorized self-merge path.

PR #1958, opened 2026-07-12
Keep user actions non-blocking during agent delivery

huangruiteng:
## Summary
- classify `user_action` todos as non-gating notices instead of user gates
- preserve visible notification actions and protocol `user_action_pending` while runnable agent work continues
- add state-machine parity coverage for action-required, notice, delivery, and scheduler semantics

## Validation
- full pytest: 111 passed, 2 skipped
- focused interaction, work-lane, user-gate scope, and quota smokes
- risk-based premerge: 17/17 passed; no manual holds; public boundary clean

### DIFF
--- a/examples/control_plane/interaction-contract-state-machine-smoke.py
+++ b/examples/control_plane/interaction-contract-state-machine-smoke.py
@@ -421,6 +421,40 @@ def assert_user_notice_can_coexist_with_bounded_delivery() -> None:
     assert contract["agent_channel"]["delivery_allowed"] is True, contract
 
 
+def assert_user_action_is_non_blocking_notice() -> None:
+    payload = base_payload(
+        should_run=True,
+        effective_action="normal_run",
+        work_lane=advancement_lane(),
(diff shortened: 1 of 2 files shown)


huangruiteng:
Self-review / premerge gate:

- Behavior: `user_gate` remains blocking; `user_action` becomes a visible non-blocking notice. The user channel keeps `notify=NOTIFY`, `actions[]`, and `non_blocking=true`; protocol output keeps `user_action_pending=true`; agent delivery and scheduler follow actual runnable work.
- Regression focus: verifies `action_required=false`, bounded delivery remains allowed, and user-facing notice is not dropped. Existing user-gate and scoped-gate smokes still pass.
- Validation: 111 passed, 2 skipped; focused state-machine and quota smokes passed; premerge 17/17; boundary clean; no manual holds.
- Scope: two public control-plane files, no benchmark, permission, production, or persisted-state changes. Self-merge is appropriate after CI passes.

PR #1959, opened 2026-07-12
Right-size CLI module ownership budgets

huangruiteng:
## Summary

- raise the default `loopx/cli_commands` ownership budget from 500 to 1000 lines
- keep the intentionally narrow starter-family budgets unchanged
- retain temporary legacy exceptions only for the two modules that are actually above 1000 lines
- replace the prescriptive split-only failure text with a cohesive-owner extraction hint

This is a validation-policy correction, not a runtime behavior change. It avoids mechanical module moves that add indirection without a concrete ownership benefit.

## Validation

- `python3 examples/cli-command-module-size-ownership-command-modularization-smoke.py`
- exact Full Public shard 1: 100/100 checks passed with the CI timeout of 120 seconds
- `loopx canary premerge --from-git-diff`: 4/4 selected checks passed, public boundary clean, self-merge allowed
- `git diff --check`

## Scope

One durable public validation smoke only. No runtime, CLI behavior, permission, benchmark, evidence-policy, private state, or generated log changes.

### DIFF
--- a/examples/cli-command-module-size-ownership-command-modularization-smoke.py
+++ b/examples/cli-command-module-size-ownership-command-modularization-smoke.py
@@ -8,21 +8,10 @@ from pathlib import Path
 ROOT = Path(__file__).resolve().parents[1]
 CLI_COMMANDS = ROOT / "loopx" / "cli_commands"
 
-DEFAULT_MAX_LINES = 500
+DEFAULT_MAX_LINES = 1000
 LEGACY_MODULE_LIMITS = {
     "benchmark_review_lifecycle.py": 1300,
-    "benchmark_run_ledger.py": 820,
-    "canary.py": 580,
-    "explore.py": 800,
-    "history.py": 720,
-    "project_lifecycle.py": 580,
-    "registry_admin.py": 780,
-    "status.py": 940,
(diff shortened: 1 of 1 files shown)


huangruiteng:

Changed surface:
- one public validation-policy smoke under `examples/`
- default ordinary CLI command-module budget: 500 -> 1000
- starter-family explicit budgets remain unchanged
- only two modules above 1000 retain temporary legacy allowances

Validation:
- CLI ownership smoke passed
- repository Python line-budget smoke passed
- exact Full Public shard 1 passed 100/100 with 0 failures and 0 timeouts at the CI 120-second timeout
- premerge canary passed 4/4 selected checks
- public/private boundary scan and diff hygiene passed

Failures or skips:
- an earlier local 30-second shard probe timed out four unrelated slow canary/install checks; the exact CI timeout rerun passed all 100 checks
- no required GitHub check is registered for this narrow PR

Manual holds:
- none; premerge reports `self_merge_allowed=true`

Coverage is sufficient because the change only adjusts the smoke policy that caused the main failure, and the exact failing shard plus both line-budget contracts pass without runtime edits.

PR #1963, opened 2026-07-12
Fix diagnose runtime capability projection

huangruiteng:
## Motivation

`quota should-run` already accepts an agent identity and observed runtime capabilities, but `diagnose` only forwarded the identity. A network-requiring todo could therefore be eligible in scoped quota while the diagnostic packet incorrectly described `network` as a user-held gate.

## Implementation

- add repeatable `diagnose --available-capability`
- forward the runtime capability envelope into the same quota decision builder used by `quota should-run`
- preserve the envelope in diagnosis packets and generated drill-down commands
- extend the existing diagnose smoke with a network-required, agent-claimed todo

## Validation

- `python3 examples/agent-diagnose-packet-smoke.py`
- `uvx ruff check loopx/diagnose.py loopx/cli_commands/status.py examples/agent-diagnose-packet-smoke.py`
- real connected-goal comparison: diagnosis reports `agent_work_attention`, `should_run=true`, and no user action when `network` is supplied
- `loopx check` on all three changed public files: 0 errors, 0 warnings
- `loopx canary premerge --from-git-diff`: standard tier passed, including selected catalog and risk-profile smokes

## Risk and non-goals

The option is additive and omitted calls preserve existing behavior. This does not infer host capabilities automatically or change quota policy; callers remain responsible for declaring capabilities actually present in the current runtime.

### DIFF
--- a/examples/agent-diagnose-packet-smoke.py
+++ b/examples/agent-diagnose-packet-smoke.py
@@ -17,6 +17,7 @@ from loopx.diagnose import _first_agent_todo_text, render_diagnosis_markdown  # 
 
 GOAL_ID = "diagnose-smoke-goal"
 SCOPED_GOAL_ID = "diagnose-smoke-agent-scoped"
+CAPABILITY_GOAL_ID = "diagnose-smoke-runtime-capability"
 
 
 def run_cli(*args: str, cwd: Path = REPO_ROOT) -> dict:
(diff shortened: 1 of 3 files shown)


huangruiteng:
## Findings

No blocking findings.

- Assumes callers declare only capabilities that are actually available in the current runtime; this PR intentionally does not infer them.
- Existing diagnose calls that omit the new option retain their current behavior.

The motivation is valid: diagnose and scoped quota must reason over the same agent/runtime envelope, otherwise a healthy autonomous lane can be presented as a user gate. The implementation fixes the shared decision seam rather than adding a command-specific exception: the repeated capability arguments are normalized, passed to the existing quota builder, exposed in the compact packet, and retained in generated drill-down commands.

Installed-user risk is low. The CLI change is additive, quota policy and default behavior are unchanged, and the focused regression covers both the old blocked result and the capability-enabled result.

## Validation

- focused diagnose smoke passed
- real connected-goal comparison passed: agent work, should-run true, no user action
- ruff passed on all changed Python files
- public boundary check passed with 0 errors and 0 warnings
- standard diff canary passed, including selected catalog and risk-profile smokes
- GitHub pytest passed

## Merge decision

Approved for owner-authorized self-merge. The change is single-purpose, validated, public-safe, and does not alter permission, quota, benchmark, or evidence policy.

PR #1969, opened 2026-07-12
Fix vision replan starvation under monitor load

huangruiteng:
## Summary

- retain each agent's latest durable context per field so a newer vision checkpoint cannot evict an older still-active vision
- classify an ordinary due todo monitor as monitor-only when no advancement frontier exists, so goal-level replan runs before repeated monitor cycling
- preserve explicit external-evidence and dependency observations as bounded read-only work

## Root cause

The compact run-history retention code used one shared slot for `agent_vision`, `vision_checkpoint`, and `autonomous_replan_ack`. A newer `unchanged_with_reason` checkpoint therefore hid the still-active vision outside the display window. Separately, due monitors set `must_attempt_work=true`, which excluded them from monitor-only frontier detection even when advancement was zero.

## Validation

- `python3 examples/control_plane/run-history-agent-context-retention-smoke.py`
- `python3 examples/control_plane/quota-replan-decision-plane-smoke.py`
- `python3 examples/control_plane/heartbeat-quota-flow-smoke.py`
- `python3 examples/control_plane/status-neutral-run-window-smoke.py`
- `python3 examples/control_plane/peer-agent-continuation-state-machine-smoke.py`
- `python3 examples/control_plane/interaction-contract-state-machine-smoke.py`
- `python3 examples/control_plane/monitor-scheduler-contract-smoke.py`
- `uvx ruff check` on all changed Python files
- `loopx check` on all changed public files: 0 errors, 0 warnings

The standard premerge selector ran but did not emit its final JSON summary in this environment. Its preview-selected checks were executed directly above; there are no manual holds. No private state, credentials, raw runtime logs, or project-specific fixtures are included.

### DIFF
--- a/examples/control_plane/monitor-poll-writeback-smoke.py
+++ b/examples/control_plane/monitor-poll-writeback-smoke.py
@@ -309,7 +309,7 @@ def assert_unchanged_writeback() -> None:
         assert payload["before"]["work_lane_contract"]["obligation"] == "attempt_due_monitor", payload
         assert summary["after"]["effective_action"] == payload["after"]["effective_action"], payload
         assert "interaction_contract" not in summary["after"], payload
-        assert payload["after"]["interaction_contract"]["mode"] == "bounded_delivery", payload
+        assert payload["after"]["interaction_contract"]["mode"] == "autonomous_replan", payload
         records = monitor_poll_records(registry_path)
         assert [record["classification"] for record in records] == ["quota_monitor_poll"], records
 
@@ -499,7 +499,7 @@ def assert_target_key_cannot_hijack_selected_due_monitor() -> None:
         )
         assert "- ok: `False`" in markdown, markdown
         assert "- mode: `monitor-poll`" in markdown, markdown
-        assert f"- todo_id: ``" in markdown, markdown
+        assert "- todo_id: ``" in markdown, markdown
         assert f"- target_key: `{OTHER_TARGET_KEY}`" in markdown, markdown
         assert "- material_change: `True`" in markdown, markdown
         assert "- appended: `False`" in markdown, markdown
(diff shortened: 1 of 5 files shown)


huangruiteng:
## Findings

No blocking findings.

- `todo_monitor_due` is the only ordinary due-monitor kind that should yield to goal-frontier replan when advancement is empty.
- `external_evidence` and `dependency_observation` remain bounded observation lanes; the focused heartbeat regression confirms they are not converted into generic replan work.

- **Motivation:** a long-running goal could cycle through staggered due monitors while its active vision disappeared from the compact read model, so no next advancement was created.
- **Solved:** context retention is now per agent and per durable field, while explicit retirement still suppresses older vision. Due todo monitors are included in monitor-only frontier detection only when advancement counts are zero.
- **User impact:** continuous goals recover a visible replan/next-work obligation instead of looking healthy while doing only monitor polls.
- **Main risk:** quota hot-path routing. The implementation limits the new branch to `todo_monitor_due`; external-evidence/dependency observation behavior is covered unchanged.
- **Design:** this repairs the existing run-history and goal-frontier seams rather than adding project-specific state or a second scheduler.

## Validation

- focused vision retention, quota replan, monitor writeback, neutral history, heartbeat quota, monitor scheduler, peer continuation, and interaction state-machine smokes passed
- changed-file Ruff and Python compile passed
- public boundary scan passed with 0 errors / 0 warnings
- line-budget smoke passed
- GitHub `pytest` passed
- standard canary execution produced no final JSON in this environment; its preview-selected checks were executed directly and passed

## Merge decision

Self-merge authorized by the owner. The change is single-purpose, regression-covered, public-safe, and has no permission, benchmark, or destructive-state impact.

PR #1973, opened 2026-07-12
Resolve newest Codex CLI for visible agents

huangruiteng:
## Summary

- resolve default `codex` across PATH, user npm-global, npm prefix, and the host ChatGPT app, selecting the newest parseable CLI version
- freeze the selected executable before tmux launch and provide a runtime-local `codex` compatibility shim for generated and raw lane commands
- keep explicit `--codex-bin` wrappers authoritative
- expose public-safe source/version diagnostics without recording absolute paths

## Root cause

The launcher validated `codex_bin` but discarded the resolved path. Pane scripts retained the late-bound string `codex`, so the Codex App PATH selected `/opt/homebrew/bin/codex` 0.142.0 even though npm-global held 0.144.1. This caused gpt-5.6-sol requests to fail with HTTP 400.

## Live validation

A fresh KNN visible launch used the unchanged default `--codex-bin codex` and reported `selected_source=user_npm_global`, `selected_version=0.144.1`, `path_frozen=true`. All four panes showed OpenAI Codex v0.144.1 with gpt-5.6-sol high; curator executed its first LoopX tick without the version error.

## Checks

- executable-resolution, visible-launcher, generic multi-agent, Auto Research policy/user-contract, and line-budget smokes passed
- post-rebase `loopx canary premerge --from-git-diff`: passed, no warnings or manual holds
- public/private boundary scan clean across all five changed files

### DIFF
--- a/examples/generic-multi-agent-one-command-smoke.py
+++ b/examples/generic-multi-agent-one-command-smoke.py
@@ -438,7 +438,7 @@ def main() -> int:
             start_payloads.append(
                 command_path.read_text(encoding="utf-8") if command_path.is_file() else command
             )
-        assert all("LOOPX_CODEX_BIN=codex" in command for command in start_payloads), start_payloads
+        assert all("LOOPX_RESOLVED_CODEX_BIN" in command for command in start_payloads), start_payloads
         assert all("LOOPX_CODEX_REASONING_EFFORT=high" in command for command in start_payloads), start_payloads
         assert all("exec python3 -c" in command for command in start_payloads), start_payloads
         assert all("LOOPX_PANE_ARTIFACT_DIR" in command for command in start_payloads), start_payloads
--- a/examples/visible-codex-executable-resolution-smoke.py
+++ b/examples/visible-codex-executable-resolution-smoke.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env python3
+"""Smoke-test durable Codex executable selection for visible agents."""
+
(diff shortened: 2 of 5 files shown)


huangruiteng:

**Residual risk / assumptions**
- Automatic newest-version selection applies only when the request is the default token `codex`; an explicit path or wrapper remains authoritative even when older.
- Version probing is local and bounded to five seconds per candidate. Unparseable candidates remain fallback-only and do not outrank parseable versions.
- The public launch result includes source categories and versions but excludes selected absolute paths; the absolute path exists only in runtime-local scripts.

- This fixes the responsible kernel boundary: executable resolution is frozen before tmux rather than deferred to each pane PATH.
- Candidate discovery is host-generic first (all PATH entries and npm prefixes), with the ChatGPT app bundle as an optional macOS candidate.
- A runtime-local shim keeps raw `codex` lane commands aligned with the frozen executable; generated lanes use the absolute frozen variable directly.
- No installer mutation or global PATH rewriting is required.

**Validation**
- focused resolver, launcher, generic multi-agent, Auto Research policy/user-contract, and line-budget smokes passed
- live default-token KNN launch selected npm-global Codex 0.144.1 and ran gpt-5.6-sol high without the 400 version error
- post-rebase premerge canary passed; public boundary scan clean; no warnings or manual holds

**Decision**
Approved for self-merge after remote pytest passes; use admin bypass only for review requirements, never a failing check.

PR #1990, opened 2026-07-12
fix(vision): replan after closing an active goal stage

huangruiteng:
## Summary

- Treat `vision_closed` as completion of one bounded stage, not termination of an active registry goal.
- Project `vision_successor_required` and select autonomous replan before ordinary advancement or monitor quiet while the goal remains active.
- Preserve explicit terminal lane semantics: `retired`, `superseded`, and `no_followup` do not require another vision.
- Document the lifecycle rule and add it to the self-repair pattern catalog.

## Why

The previous read model returned no acceptance gap for every closed vision state. A long-running goal could therefore complete one valid stage, retain ordinary todos or monitors, and continue indefinitely without establishing a successor vision.

## Validation

- `uv run --extra test pytest -q tests/control_plane/test_goal_vision_succession.py` — 8 passed.
- `python3 examples/project/goal-vision-closed-successor-smoke.py` — passed through the real `refresh-state → quota should-run` path.
- Existing goal-vision, quota decision-plane, and autonomous-replan smokes — passed.
- Standard risk-based premerge canary — 17/17 checks passed, including catalog, risk-profile, compile, diff, and public-boundary checks.

## Risk and compatibility

This intentionally changes routing only when the latest per-agent state normalizes to `vision_closed` and the registry goal status is `active` or `active-*`. Completed goals and explicit terminal lane states keep the previous no-successor behavior.

No private state, local paths, credentials, raw logs, or project-specific evidence are included.

PR #1993, opened 2026-07-12
fix(vision): replan before ready deferred work

huangruiteng:
## Summary

- preserve `vision_successor_required` precedence when the current agent also has resume-ready deferred work
- keep explicit user and blocking handoff gates ahead of successor replan
- extend the active-goal vision succession regression and self-repair pattern

## Root cause

The first active-goal stage-close repair correctly projected a `vision_successor_required` acceptance gap. However, `derive_goal_frontier_replan_obligation_from_summaries()` returned early whenever a deferred successor was resume-ready, before inspecting the gap kind. A real active goal therefore exposed the new gap but still selected ordinary work with `effective_action=normal_run`.

## Validation

- 14 focused vision-succession unit cases
- source CLI replay against the real active goal changed `normal_run` to `autonomous_replan_required` while preserving the same `vision_successor_required` gap
- explicit user-gate precedence regression
- Ruff, py_compile, diff check, and public-boundary scan
- standard premerge gate: 4 direct checks, 8 catalog canaries, and 8 risk-profile smokes passed; zero failures, warnings, or manual holds

## Risk

The behavior change applies only when an active goal has a closed-stage successor-vision gap. Ready deferred work remains the normal selection shortcut for all other frontier states. Explicit user and blocking handoff gates remain unchanged.


huangruiteng:
## Findings

No blocking findings.

The bug was ordering, not missing state: `vision_successor_required` was visible, but the generic ready-deferred shortcut returned before the vision gap was evaluated. The fix keeps the shortcut for ordinary frontier states and bypasses it only for a closed-stage successor vision. Explicit user and blocking handoff gates still win.

This is the narrowest reusable correction because precedence remains in the goal-frontier policy seam; quota and project adapters do not gain special cases.

## Validation

- 14 focused tests, including ordinary work, resume-ready deferred work, and user-gate precedence
- real active-goal source CLI replay returned `autonomous_replan_required`
- standard premerge gate passed: 4 direct, 8 catalog, and 8 risk-profile checks; zero failures/warnings/holds
- public-boundary scan and GitHub pytest passed

## Merge decision

Approved for owner-authorized self-merge. The diff is narrow, regression-backed, and directly repairs the installed callsite contradiction found after #1990.

PR #1994, opened 2026-07-12
test(control-plane): add model behavior qualification contract

huangruiteng:
## Summary

- add a provider-neutral `model_behavior_qualification_v0` contract for paired full-packet and candidate-packet agent decisions
- enforce candidate/full lineage, a no-write/no-tools boundary, strict known schemas, public-safe inputs, and compact non-conversational receipts
- separate safety hard-invariant drift from ordered behavior-signal drift so weak action-plan changes remain visible
- document the promotion boundary while keeping the default `quota should-run` view unchanged

## Product and architecture

This creates the missing validation seam between deterministic TurnEnvelope parity and future low-frequency live-model qualification. The core owns packet lineage and validation, sandbox policy, response normalization, compact receipts, and comparison; provider transport remains out of scope. That keeps secrets, cost/rate-limit policy, and provider-specific behavior out of the control-plane contract until a real adapter is validated.

The primary residual risk is that the initial behavior vocabulary is intentionally compact. Future corpus work can add evidence-backed dimensions, but unknown schemas and action kinds fail closed instead of silently widening the contract.

## Validation

- `pytest tests/control_plane/test_model_behavior_qualification.py tests/test_turn_envelope.py -q` (28 passed)
- `pytest tests/control_plane tests/test_turn_envelope.py tests/test_operator_markdown_renderers.py -q` (97 passed)
- `ruff check` and `ruff format --check` on changed Python files
- strict mypy on the new module with imported modules skipped (new module clean; repository-wide imported-module strict errors are pre-existing)
- `loopx check` on all four changed public paths (clean)
- `loopx canary premerge --from-git-diff --tier standard` (4 direct + 14 selected checks passed; 0 failures, 0 warnings, 0 manual holds)

## Excluded

- no live model or provider API calls
- no credentials, environment reads, raw conversations, trajectories, logs, local paths, or private material
- no CLI, default-output, permission, benchmark, production, or external-write behavior changes

### DIFF
--- a/docs/reference/protocols/README.md
+++ b/docs/reference/protocols/README.md
@@ -19,6 +19,7 @@ Current contracts:
 - [agent_management_projection_v0](agent-management-projection-v0.md)
 - [todo_detail_cold_path_v0](todo-detail-cold-path-v0.md)
 - [long_horizon_agent_state_protocol_v0](long-horizon-agent-state-protocol-v0.md)
+- [model_behavior_qualification_v0](model-behavior-qualification-v0.md)
 - [decentralized_auto_research_state_v0](decentralized-auto-research-state-v0.md)
 - [auto_research_lane_contract_v1](auto-research-lane-contract-v1.md)
 - [auto_research_role_state_machine_v0](auto-research-role-state-machine-v0.md)
--- a/docs/reference/protocols/model-behavior-qualification-v0.md
+++ b/docs/reference/protocols/model-behavior-qualification-v0.md
@@ -1,0 +1,83 @@
+# Model Behavior Qualification v0
+
+`model_behavior_qualification_v0` is a low-frequency validation contract for
+agent-facing control-plane packet changes. It complements deterministic smokes;
+it does not replace them and does not change the default `quota should-run`
+view.
+
+The first implementation is intentionally provider-neutral. It defines the
+actor request, no-write sandbox, strict model decision, compact receipt, and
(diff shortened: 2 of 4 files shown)


huangruiteng:
## Self-review

**Findings:** no unresolved findings. During review I found and fixed one P1 before merge: the first pair runner verified the same case and actor but did not prove that the candidate was derived from the paired full packet. The updated head now requires both `action_signature.matches=true` and an exact `source_decision_hash` lineage match before either actor call; focused tests reject unrelated and unverified candidates.

**Assumption:** a future provider adapter must enforce provider-transport-only networking and return parsed JSON with no tool calls. This PR deliberately does not add that adapter or claim a live-model result.

**Product/architecture judgment:** the PR solves the current bounded problem by placing provider-neutral behavior qualification under `control_plane.testing`, next to the existing TurnEnvelope contract, without changing the hot path. It captures both hard safety/route invariants and ordered action-kind drift while excluding model reasoning and source packets from durable receipts. The main risk is vocabulary coverage, addressed by fail-closed schemas and keeping default promotion separately gated.

**Validation:** 28 focused tests; 97 wider control-plane tests; ruff; isolated strict mypy; public boundary scan clean; standard premerge canary 4 direct + 14 selected checks, 0 failures/warnings/manual holds. No live API call, credential read, raw conversation, benchmark action, or production/external write occurred.

**Merge decision:** self-merge after GitHub CI passes, under the owner's standing authorization for similarly bounded, behavior-preserving PRs.

PR #1998, opened 2026-07-12
test(control-plane): add direct Doubao qualification actor

huangruiteng:
## Summary

- add a direct Ark `DoubaoModelBehaviorActor` for low-frequency full-packet versus TurnEnvelope behavior qualification
- allowlist the versioned Doubao 2.1 Pro/Turbo models and canonical Ark endpoint, reject redirects, tools, arbitrary base URLs, malformed responses, and non-canonical actor requests
- keep `ARK_API_KEY` runtime-only, sanitize transport failures, and retain only the existing compact receipts and pair results

## Product and architecture

The provider-neutral qualification core remains unchanged in ownership. This PR adds one optional provider edge under `control_plane.testing`; it is not wired into `quota should-run`, CI, PR smoke, or any default agent path. The adapter reconstructs and validates the canonical no-write request before provider transport, so direct calls cannot silently bypass packet redaction and sandbox constraints.

The live model call remains a manual or low-frequency shadow operation. The current host did not expose `ARK_API_KEY`, so no credential store was inspected and no live call was attempted through OpenViking or another service.

## Validation

- `pytest -q tests/control_plane/test_doubao_model_behavior_actor.py tests/control_plane/test_model_behavior_qualification.py tests/test_turn_envelope.py` (35 passed)
- `pytest -q tests/control_plane tests/test_turn_envelope.py tests/test_operator_markdown_renderers.py` (104 passed)
- focused `ruff check` and `ruff format --check`
- strict mypy on both qualification modules
- docs governance smoke
- public-boundary check clean for all four changed paths
- standard premerge canary passed all selected checks with no failures

## Boundaries

- no default-output, routing, quota, scheduler, permission, benchmark, or production behavior change
- no live API call, tool call, filesystem/external write, raw model response, conversation, trajectory, credential, local path, or private material committed
- missing provider access and any behavior drift keep the full packet as the default

### DIFF
--- a/docs/reference/protocols/model-behavior-qualification-v0.md
+++ b/docs/reference/protocols/model-behavior-qualification-v0.md
@@ -5,10 +5,10 @@ agent-facing control-plane packet changes. It complements deterministic smokes;
 it does not replace them and does not change the default `quota should-run`
 view.
 
-The first implementation is intentionally provider-neutral. It defines the
-actor request, no-write sandbox, strict model decision, compact receipt, and
-paired comparison. A provider adapter can be added separately after its secret
-injection, redaction, cost, and rate-limit boundaries are validated.
+The core is provider-neutral. It defines the actor request, no-write sandbox,
+strict model decision, compact receipt, and paired comparison. The optional
+direct Ark adapter supports low-frequency Doubao 2.1 shadow runs without
+changing the default quota path.
 
 ## Pair Contract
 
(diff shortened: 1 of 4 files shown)


huangruiteng:
## Self-review

**Findings:** no unresolved findings. During review I tightened one safety boundary before opening the PR: direct adapter calls now reconstruct and compare the canonical actor request before network transport, so a caller cannot flip `tools_enabled`, weaken the no-write sandbox, or bypass packet public-safety checks. I also adjusted the fixture after the public scanner correctly treated a literal Bearer header as credential-shaped evidence.

**Assumption / hold:** this PR proves the provider edge deterministically but does not claim a live-model result. The current process has no injected `ARK_API_KEY`; it failed closed, and I did not inspect credential stores or route through OpenViking. A later shadow run must inject the key at process runtime and retain only compact receipts.

**Product/architecture judgment:** the change solves the missing direct-provider seam without coupling Ark transport into quota/status or the default agent path. The canonical endpoint/model allowlists, redirect rejection, strict request reconstruction, no-tools request, bounded response size, and sanitized errors keep the blast radius in `control_plane.testing`. A generic arbitrary-URL LLM client would be broader and riskier here; the narrow adapter is the right current abstraction.

**Validation:** 35 focused tests; 104 wider control-plane tests; focused ruff; strict mypy on both qualification modules; docs governance smoke; four-path public-boundary scan clean; standard premerge canary passed all selected checks with no failures. Live call skipped only because runtime injection was absent.

**Merge decision:** self-merge after GitHub pytest passes under the owner's standing authorization for bounded, behavior-preserving LoopX capability PRs. This does not promote TurnEnvelope or change installed-user behavior.

PR #1999, opened 2026-07-12
test(control-plane): add model behavior corpus grader

huangruiteng:
## Summary

- add an in-memory `model_behavior_corpus_v0` builder for the deterministic TurnEnvelope state matrix, retained public-safe decisions, counterfactuals, and candidate field ablations
- randomize full/candidate arm order across repeated runs and report hard, behavior-signal, first-action, and trajectory-action drift separately
- independently recompute full-versus-candidate action semantics before actor execution, so a mutated envelope cannot rely on a stale `matches=true` flag
- fail promotion closed while required decision dimensions remain explicitly ungraded

## Product and architecture

This is the provider-neutral corpus/grader layer above the existing paired actor contract. Raw packets remain in-memory inputs; durable results contain only case/source ids, drift field names, safety codes, arm order, and receipt digests. The module accepts retained public-safe packets but does not introduce a repository of raw decisions or model conversations.

The current receipt vocabulary still lacks concrete user question, required reads, write scope, spend, scheduler, vision, and actionable-warning dimensions. The result reports those as `ungraded_required_dimensions` and therefore keeps `promotion_eligible=false`, even when every currently graded case passes. This is deliberate: it prevents partial evidence from being interpreted as semantic equivalence.

## Validation

- focused qualification/corpus/TurnEnvelope suite: 39 passed
- wider control-plane + TurnEnvelope/renderers suite: 108 passed
- focused ruff and strict mypy
- docs governance smoke
- four-path public boundary scan clean
- standard premerge canary: 10 selected checks passed, 0 failures/warnings/manual holds; self-merge allowed

## Boundaries

- no live model/API call, credential read, raw response, conversation, trajectory, or private material
- no default quota output, routing, scheduler, permission, benchmark, production, or external-write behavior change
- corpus input is not a durable public artifact; missing coverage blocks promotion

### DIFF
--- a/docs/reference/protocols/model-behavior-qualification-v0.md
+++ b/docs/reference/protocols/model-behavior-qualification-v0.md
@@ -20,8 +20,10 @@ One qualification case runs the same actor against two public-safe inputs:
 Both arms share `qualification_id` and `actor_ref`. Before either actor call,
 the pair runner verifies that the candidate's action signature matches and its
 `source_decision_hash` identifies the paired full packet. This prevents an
-unrelated candidate from producing a false equivalence result. The comparator
-then checks these hard behavior dimensions:
+unrelated candidate from producing a false equivalence result. The runner also
+recomputes both semantic signature documents instead of trusting the
+candidate's stored `matches` flag; a field ablation therefore fails before any
+provider call. The comparator then checks these hard behavior dimensions:
 
 - decision: execute, wait, ask the user, or stop;
 - selected todo;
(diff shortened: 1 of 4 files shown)


huangruiteng:
## Self-review

**Findings:** no unresolved findings. Two safety fixes were made during review. First, pair validation now recomputes the canonical action-signature documents from both packets; deleting an envelope field while leaving the stored `matches=true` no longer reaches the actor. Second, corpus `fail_closed` expectations catch only deterministic `ModelBehaviorPairValidationError`; malformed actor output or provider errors cannot be miscounted as a successful ablation.

**Open assumption:** the first corpus result is intentionally never promotion-eligible because seven required semantic dimensions are not yet in the decision receipt. Retained public-safe real decisions are accepted as runtime inputs, but this PR does not commit any raw decision corpus.

**Product/architecture judgment:** the change adds one bounded testing module above the existing provider-neutral pair contract and keeps all raw inputs ephemeral. Randomized arm order and repeated compact grading solve the immediate evaluation-infrastructure gap without widening quota/status or inventing a general benchmark framework. Explicit ungraded coverage is preferable to a broad receipt rewrite in the same PR because it makes the remaining risk machine-visible and keeps this batch reversible.

**Validation:** 39 focused tests; 108 wider control-plane tests; ruff; strict mypy; docs governance; four-path public scan; standard premerge canary 10/10 with zero warnings or holds.

**Merge decision:** self-merge after GitHub pytest passes under standing authorization. The P1 todo remains open for receipt-schema coverage and retained live shadow evidence; this PR alone does not justify TurnEnvelope promotion.

PR #2001, opened 2026-07-12
test(control-plane): grade semantic packet dimensions

huangruiteng:
## Summary

- add opt-in `semantic_contract` grading for concrete user question, required reads, gate/stop state, write scope, spend rule, scheduler action, vision continuation, and actionable warnings
- derive expected semantics independently from each full packet and TurnEnvelope arm, so two matching but equally wrong model interpretations fail
- keep semantic values ephemeral; receipts retain only per-dimension digests, completeness, and mismatch field names
- let complete deterministic coverage pass the corpus gate while overall promotion stays blocked on repeated live-model evidence and explicit owner review

## Product and architecture

This closes the largest false-equivalence risk left by the first corpus runner. Pairwise equality alone cannot detect both arms omitting the same weak signal. Corpus mode now requires the model to return a bounded semantic contract, and the core aligns each arm against its own canonical action-signature projection before comparing arms.

The requirement is opt-in at the pair/corpus layer, so existing pair callers remain compatible. The Doubao prompt knows the additive response shape, but no live path or default quota output is enabled. The corpus result separates `corpus_gate_passed` from overall `promotion_eligible`; live repeated evidence and owner review remain mandatory blockers.

## Validation

- focused semantic/corpus/qualification/TurnEnvelope suite: 41 passed
- wider control-plane + TurnEnvelope/renderers suite: 110 passed
- focused ruff and strict mypy across all three qualification modules
- docs governance smoke
- six-path public boundary scan clean
- standard premerge canary: 10 selected checks passed, 0 failures/warnings/manual holds; self-merge allowed

## Boundaries

- no live model/API call, credential read, raw response, conversation, trajectory, or private material
- no default quota output, routing, scheduler, permission, benchmark, production, or external-write behavior change
- semantic values are never present in durable receipts or corpus results

### DIFF
--- a/docs/reference/protocols/model-behavior-qualification-v0.md
+++ b/docs/reference/protocols/model-behavior-qualification-v0.md
@@ -55,11 +55,17 @@ names, safety codes, and receipt digests. It excludes packets, prompts, raw
 responses, and conversations. Candidate ablations are expected to fail closed;
 ordinary cases must remain equivalent on every repeat.
 
-Coverage is explicit. A passing corpus is not promotion-eligible while a
-required behavior dimension remains `ungraded`. The initial runner names
-concrete user questions, required reads, write scope, spend rule, scheduler
-action, vision continuation, and actionable warnings as remaining receipt
-schema work instead of treating their absence as equivalence.
+Coverage is explicit. Corpus mode requires a bounded `semantic_contract` for
+the concrete user question, required reads, gate/stop state, write scope, spend
+rule, scheduler action, vision continuation, and actionable warnings. The core
+derives the expected contract independently from each arm's packet and compares
+the model result with that source before comparing arms. Two arms that repeat
+the same wrong or incomplete interpretation therefore fail source alignment.
+
+Receipts retain only per-dimension digests, completeness, and mismatch field
+names; they do not retain semantic-contract values. Complete aligned coverage
+can pass the corpus gate, but the overall promotion decision remains false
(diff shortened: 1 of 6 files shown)


huangruiteng:
## Self-review

**Findings:** no unresolved findings. The main review question was whether two arms could produce the same wrong semantic answer and still pass. They cannot: each arm is aligned against an independently derived canonical semantic contract first, and only digest/mismatch metadata reaches the receipt. I also kept corpus-gate success distinct from overall promotion, which remains false without repeated live-model evidence and explicit owner review.

**Open assumption:** exact semantic echo is intentionally strict. Stochastic wording is not accepted for user questions, required reads, scope, scheduler, or warnings because those values are execution contracts, not prose-quality signals. The live model may expose prompt/schema weaknesses; such failures should refine the actor prompt or contract, not relax source alignment.

**Product/architecture judgment:** the additive opt-in flag preserves existing pair callers while making corpus qualification meaningfully stronger. Reusing the canonical action-signature projection avoids a second packet parser and keeps the semantics owned by the same control-plane read model. Digest-only persistence protects trajectory hygiene without losing drift attribution.

**Validation:** 41 focused tests; 110 wider control-plane tests; ruff; strict mypy across all qualification modules; docs governance; six-path public scan; standard premerge canary 10/10 with zero warnings or holds.

**Merge decision:** self-merge after GitHub pytest passes under standing authorization. This PR does not enable the compact hot path or satisfy the live evidence gate.

PR #2006, opened 2026-07-12
fix(installer): serialize concurrent promotions

huangruiteng:
## Summary
- serialize default-promotion installs with a Unix fcntl guard that survives exec and releases automatically when the installer exits
- preserve the existing directory lock so new installers remain mutually exclusive with an older installer already in flight
- strengthen the release concurrency smoke with a deterministic pre-held-guard check plus same-release-id collision coverage

## Root cause
Two installers could both observe the same stale directory-lock PID. After one reclaimed that stale directory and created a fresh lock, the delayed second reclaimer could move the fresh directory before its owner wrote pid, producing the observed missing .install-lock/pid failure.

## Validation
- bash -n scripts/install-local.sh
- Ruff and py_compile on the changed Python smoke
- release-promotion-concurrency-smoke: passed
- install-local-smoke: passed
- install-local-overwrite-smoke: passed
- local-install-promotion-boundary-smoke: passed
- loopx canary premerge --from-git-diff: 17/17 selected checks passed, no warnings or manual holds
- nested dashboard browser readiness was skipped by the canary because npm dependencies are unavailable; it is outside this installer-lock surface

## Risk
The guard uses Python fcntl on the same Unix platforms already required by the Bash installer. Canary-only checkout installs remain unchanged. Release-id allocation, manifests, symlink promotion, and legacy lock semantics are otherwise preserved.


huangruiteng:
**Premerge validation evidence**

- **Changed surfaces:** install/release runtime, release concurrency smoke, public boundary
- **Direct checks:** staged/unstaged/committed diff hygiene passed; Bash syntax passed; changed Python compile and Ruff passed
- **Focused behavior:** release-promotion-concurrency smoke passed with deterministic guard blocking and distinct same-id releases; install-local, overwrite, and promotion-boundary smokes passed
- **Risk-based canary:** 17/17 selected checks passed with zero warnings, failures, or manual holds
- **Public/private boundary:** clean for both changed files
- **Skip:** the nested dashboard browser readiness check skipped because npm dependencies are unavailable; this is outside the installer-lock surface
- **GitHub checks:** none are eligible for this PR path set. Python Tests excludes scripts/** and examples/**, while Full Public runs on main push/schedule.
- **Merge decision:** hold for the independent peer review tracked in LoopX; this installer runtime change is not being self-merged by its implementer.

huangruiteng:
## Independent review of `dae004ad`

### Finding

1. **[P1] A waiter that times out on the legacy directory lock deletes the active owner lock.**

`scripts/install-local.sh:95` assigns `install_lock` before `mkdir` succeeds, while the global `EXIT` cleanup at lines 29-30 removes any directory at that path. If a new guarded installer encounters an older installer that legitimately owns `.install-lock` for the full 600-attempt window, line 112 exits and the waiter cleanup removes the older process lock even though that process is still alive. The fcntl guard then releases, allowing the next new installer to enter the directory-lock path and potentially overlap the old install. This breaks the stated compatibility contract with an existing directory-lock installer and reintroduces unsafe promotion overlap through a different exit path.

Public temporary-directory reproduction: pre-create `.install-lock/pid` with a live `sleep` owner, run this PR installer, and accelerate only the Bash retry sleep so the 600-attempt timeout completes. Result: `returncode=1`, active owner still running, but `lock_exists_after_timeout=False`.

**Required fix:** track lock ownership separately and remove `.install-lock` only after this process successfully created it. Add a regression with a live pre-held legacy directory lock proving timeout/failure leaves the owner directory and PID intact; retain the existing stale-owner reclamation check.

The fcntl guard is the right reusable serialization layer for eliminating the stale-directory ABA race, and the inherited FD does survive the Python-to-installer exec path covered by the focused smoke. Keeping the legacy directory lock is also the correct migration strategy, but its cleanup must be ownership-aware before this can safely coexist with an old installer already in flight.

### Validation

Passed:
- `release-promotion-concurrency-smoke.py`
- `install-local-smoke.py`
- `install-local-overwrite-smoke.py`
- `local-install-promotion-boundary-smoke.py`
- changed-file public boundary scan clean

The author premerge packet reports 17/17 checks passing. The only `loopx check` warning in my run concerns unrelated OpenViking history index duplicates; both changed files passed the public boundary scan.

### Decision

**Hold / request changes.** No files were edited and no merge was performed.

huangruiteng:
## Ownership-aware follow-up

The live-owner deletion finding is fixed in `e0ef7ae1` after merging current `main` (`10f7acbb`).

- `scripts/install-local.sh` now removes `.install-lock` only when this installer successfully created it.
- The release concurrency smoke now pre-holds the legacy lock with a live PID, accelerates only the retry sleep, and proves timeout leaves both the lock directory and PID intact.
- Existing stale-owner reclamation coverage remains in the same smoke.

Validation passed:
- Bash syntax and changed Python compile/Ruff
- release promotion concurrency smoke
- local install overwrite and promotion-boundary smokes
- changed-file `loopx check` public-boundary scans
- `git diff --check`

The standard canary selected 17 checks. Its only three failures are the same `install-local-smoke.py` expectation (directly and through the canary wrapper) that assumes the old non-thin default heartbeat body; the identical failures reproduce on clean `origin/main` at `10f7acbb`, so they are baseline drift unrelated to this installer-lock change.

Please re-review the ownership fix before merge; installer runtime policy still requires an independent reviewer.

huangruiteng:
## Independent latest-head review of `e0ef7ae1`

### Findings

None. The prior P1 is resolved: `install_lock_owned` remains false for a waiter and is set only after this process successfully creates `.install-lock`, so EXIT cleanup no longer removes a live legacy owner's directory. The new regression exercises that exact timeout path while retaining stale-owner reclamation coverage.

- Reviewed the current remote head `e0ef7ae1`; it did not change during validation.
- The PR is conflict-free against current `origin/main` (`ffee394a`). GitHub reports no checks for this path set and still projects `REVIEW_REQUIRED`, so merge requires the already authorized maintainer admin-bypass path.
- No tracked or untracked files were added to the review worktree.

- **Motivation:** prevent concurrent local release promotion while remaining compatible with an older installer that only owns the legacy directory lock.
- **Solved:** yes. The fcntl guard serializes new installers, the inherited FD remains held after exec into Bash, and the ownership bit closes the live-owner deletion path without weakening stale-PID recovery.
- **User/operator impact:** a timed-out waiter now fails closed instead of allowing a later installer to overlap a still-running legacy owner.
- **Main risk:** mixed-version lock handoff and cleanup ownership. Focused live-owner, stale-owner, exec-inheritance, concurrency, and promotion-boundary checks cover that risk.
- **Design judgment:** retaining the legacy directory lock behind a process-wide fcntl guard is a cohesive migration contract; the one-bit ownership fix is narrower and safer than replacing the compatibility layer.

### Validation

Passed:

- full `origin/main...e0ef7ae1` diff and commit review; existing issue comments, reviews, inline comments, and checks inspected
- changed Python source compiled with `compile(...)`
- public-boundary scan for both changed files: 0 errors, 0 warnings
- `python3 examples/release/release-promotion-concurrency-smoke.py`
- `python3 examples/release/local-install-promotion-boundary-smoke.py`
- minimal independent inline probe: live PID timeout preserved lock inode and PID; dead PID was reclaimed; nonblocking fcntl contention remained blocked after the installer had exec'd into Bash
- merge-tree check against current `origin/main`: no conflicts

Failures / skips:

- Ruff is not installed in this review environment; changed Python syntax and runtime behavior were validated by compile and focused smokes.
- `examples/install-local-smoke.py` reproduces the unrelated thin-heartbeat prompt expectation on this older PR base; current main contains #2047's smoke-contract fix, the merge is conflict-free, and neither PR-changed file touches that contract.
- Full canary was not rerun for this bounded latest-head review; the multi-surface installer risk set above is the equivalent focused premerge coverage. No manual hold remains.

### Merge decision

**`approved-after-independent-review` — findings none.** Proceeding with the owner-authorized maintainer admin-bypass merge.

PR #2010, opened 2026-07-12
fix(quota): preserve completion across neutral refresh

huangruiteng:
## Summary
- keep the latest same-agent accountable delivery discoverable when a plain `state_refreshed` event is appended before `quota spend-slot`
- treat only refreshes with no declared `delivery_outcome` as quota-neutral
- retain fail-closed behavior for explicit `surface_only`/other non-delivery events and let accountable refreshes replace the older delivery

## Validation
- `uv run pytest -q tests/control_plane/test_quota_slot_accounting.py` (`11 passed`)
- `uv run pytest -q` (`261 passed, 2 skipped`)
- `uv run ruff check loopx/control_plane/quota/slot_accounting.py tests/control_plane/test_quota_slot_accounting.py`
- `uv run loopx canary premerge --from-git-diff` (standard tier; 12 selected checks passed; no manual holds)

## Boundary and risk
- changes only the post-delivery quota attribution lookup; quota eligibility, slot arithmetic, todo routing, scheduler behavior, and persisted event schemas are unchanged
- an explicit non-accountable outcome remains a causal stop, so this does not spend through failed or surface-only delivery attempts
- generated `uv.lock`, local state, credentials, private links, raw logs, trajectories, and verifier output are excluded

### DIFF
--- a/loopx/control_plane/quota/slot_accounting.py
+++ b/loopx/control_plane/quota/slot_accounting.py
@@ -114,9 +114,11 @@ def _latest_unspent_accountable_delivery_run(
 ) -> dict[str, Any] | None:
     """Return the latest same-agent delivery that still needs accounting.
 
-    Unchanged monitor polls and scheduler acknowledgements are quota-neutral.
-    They may occur after validation and before accounting, so they must not
-    hide the accountable run. Other non-delivery events remain fail closed.
+    Unchanged monitor polls, scheduler acknowledgements, and plain state
+    refreshes are quota-neutral. They may occur after validation and before
+    accounting, so they must not hide the accountable run. A state refresh
+    with an explicit non-accountable delivery outcome and other non-delivery
+    events remain fail closed.
     """
 
     safe_agent_id = normalize_todo_claimed_by(agent_id)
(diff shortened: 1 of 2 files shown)


huangruiteng:
## Self-review

**Findings:** No blockers found. The implementation is deliberately narrower than treating every refresh as neutral.

**Assumptions / residual risk:** A plain `state_refreshed` row with no `delivery_outcome` is a projection/read event and may be skipped during same-agent completion attribution. Any explicit outcome, including `surface_only` or `outcome_gap`, remains causally meaningful and fail-closed. Unknown non-empty outcomes and unrelated non-delivery classifications also remain fail-closed.

**Product / architecture judgment:** The bug stranded valid quota accounting after the canonical `todo complete -> refresh-state -> spend-slot` closeout sequence. Fixing the lookup at the quota attribution boundary is preferable to special-casing heartbeat callers or weakening spend eligibility. It preserves agent identity, duplicate-spend protection, slot arithmetic, and event schemas.

**Validation:**
- focused quota accounting tests: `11 passed`
- full pytest: `261 passed, 2 skipped`; skips are opt-in smoke-suite facade tests
- Ruff on both touched files: passed
- standard premerge canary: 12 selected checks passed across core-control-plane and canary-runner profiles; no failures, warnings, advisory failures, or manual holds
- boundary scan found no credentials, private paths, internal links, raw logs, trajectories, or verifier output; generated `uv.lock` excluded

**Merge decision:** Self-merge after required GitHub Python Tests passes. Coverage is sufficient because it proves both allowed neutral events and every adjacent fail-closed branch, then exercises the full control-plane canary profile.

PR #2025, opened 2026-07-13
Keep model qualification on canonical selected work

huangruiteng:
## Summary

- derive a per-arm canonical selected todo id from the already-verified action signature
- require the direct Doubao actor to preserve that exact value instead of selecting diagnostic todo references
- document the guard and its fail-closed relationship with candidate signature parity

## Motivation and judgment

A retained real quota shadow exposed stable control-text pollution: the full packet had no canonical selected todo, but the model selected a todo id that appeared only in diagnostic summaries. The candidate TurnEnvelope returned null. This narrow guard keeps deterministic controller-owned identity out of model inference while leaving the eight semantic dimensions, execution actions, and safety signals model-derived. A real candidate omission or mutation is still rejected before provider invocation by recomputed action-signature parity.

## Validation

- `32 passed` across the Doubao actor, pair qualification, corpus, and retained-case tests
- Ruff check and format check passed
- live retained case before fix: 2/2 repeats drifted on `selected_todo_id`; one also drifted on action kinds
- live retained case after fix: 2/2 repeats passed, both arm orders observed, semantic contract complete, zero hard/behavior/semantic/stochastic/safety drift
- `loopx canary premerge --from-git-diff`: passed, 10 selected checks, 0 failures, 0 manual holds
- public/private boundary scan passed; no raw packet, model response, conversation, credential, or local path is committed

## Residual boundary

This does not promote TurnEnvelope to the default agent-facing path. Promotion remains separately owner-gated.

### DIFF
--- a/docs/reference/protocols/model-behavior-qualification-v0.md
+++ b/docs/reference/protocols/model-behavior-qualification-v0.md
@@ -133,13 +133,21 @@ does not accept an arbitrary base URL, does not follow redirects, does not send
 tool definitions, and converts transport failures into bounded errors without
 provider response bodies.
 
-The provider-visible user input contains only the arm, the
-`semantic_contract_required` flag, and that arm's packet. Qualification ids,
-sandbox declarations, actor instructions, and response-contract metadata are
-validated locally but are not repeated in the model prompt. The actor disables
-provider deep thinking for this deterministic extraction task and reserves
-4096 output tokens so the bounded semantic contract is not constrained by the
-former 1200-token response budget.
+The provider-visible user input contains only the arm, a locally derived
+`canonical_selected_todo_id`, the `semantic_contract_required` flag, and that
+arm's packet. Qualification ids, sandbox declarations, actor instructions, and
+response-contract metadata are validated locally but are not repeated in the
+model prompt. The actor disables provider deep thinking for this deterministic
+extraction task and reserves 4096 output tokens so the bounded semantic
+contract is not constrained by the former 1200-token response budget.
+
+The actor derives `canonical_selected_todo_id` independently for each arm from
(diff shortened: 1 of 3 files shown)


huangruiteng:
## Maintainer self-review

**Findings:** no blocking findings. The diff is single-purpose, and the selected-todo guard is derived independently per arm from the same canonical action-signature functions already used by the pair runner. It does not hide a candidate omission: recomputed full/candidate signature parity still fails before any provider call.

**Assumptions / boundary:** selected todo identity is deterministic controller-owned data, while the eight semantic-contract dimensions, intended action sequence, decision flags, and safety signals remain model-derived. This PR does not enable TurnEnvelope by default or change quota/runtime behavior for installed users.

**Product / architecture judgment:** the retained real shadow showed full-packet diagnostic text could make the model select a non-canonical todo even when canonical selected work was null. Fixing the direct qualification adapter at its provider-input boundary is the narrow reusable correction; it avoids teaching each corpus fixture special cases and preserves the existing fail-closed action-signature contract. Capability risk is limited to low-frequency qualification behavior.

**Validation:**
- focused model behavior suite: 32 passed
- Ruff check and format check: passed
- live retained shadow after fix: 2/2 repeats passed, both arm orders observed, all semantic dimensions complete, zero hard/behavior/semantic/stochastic/safety drift
- premerge canary: 10 selected checks passed, 0 failures, 0 manual holds
- GitHub pytest: passed
- public/private boundary: passed; no raw packet, response, conversation, credential, or local path committed

**Merge decision:** self-merge authorized under the project policy and the owner durable instruction for this class of PR. Coverage is sufficient because it combines deterministic source/shape tests, risk-selected repository canaries, the real provider counterexample before/after, and remote CI.

PR #2030, opened 2026-07-13
feat(semantic-preference): add OpenViking project peer provider

huangruiteng:
## Motivation

A semantic-preference consumer should not need to normalize Git remotes, derive OpenViking peer URIs, or reproduce fallback and call-budget rules. Those details currently make provider reuse expensive and make cross-repository leakage easier to introduce.

## Changes

- Add a canonical project identity and reserved `project-<digest>` peer scope derived from Git origin, with a stable LoopX project id fallback for non-Git projects.
- Add `loopx semantic-preference openviking-provider` with `--describe-scope`, read-only `--doctor`, exact-project recall by default, explicit global fallback, and bounded find calls.
- Add public integration guidance and a focused smoke covering transport normalization, wrong-project zero recall, explicit fallback, one-find default, Issue Fix application receipt, and fail-open preservation.

## Product and architecture judgment

OpenViking remains the owner of memory extraction, ranking, write, update, and supersede semantics. LoopX only owns project identity, target selection, the existing provider protocol, and the function-owned application receipt. This is a reusable adapter contract rather than a new memory ledger or an Issue Fix-specific branch.

The project peer is a routing boundary, not a security tenant. Account and user isolation remain OpenViking concerns, and the consuming function remains the final application boundary.

## Validation

- `python3 -m py_compile loopx/*.py loopx/cli_commands/*.py loopx/capabilities/semantic_preference/*.py`
- `python3 examples/openviking-project-peer-provider-smoke.py`
- `python3 examples/semantic-preference-hook-smoke.py`
- `python3 examples/issue-fix-pr-description-builder-smoke.py`
- `uvx ruff check` and `uvx ruff format --check` on changed Python files
- `loopx canary premerge --from-git-diff`: standard tier, 14/14 selected checks passed; public boundary passed; no manual holds
- Read-only live OpenViking scope probe derived the expected stable project peer; no memory or service state was mutated

## Risk and compatibility

- Disabled configurations and existing provider commands are unchanged.
- Global fallback requires an explicit flag and at least two allowed find calls.
- Provider errors expose only a bounded failure type to stderr; raw provider output and local paths do not enter receipts.
- PR #2002 touches the same semantic-preference CLI and documentation area but is not a functional dependency; whichever lands second may need a small rebase.

## Public/private boundary

No local config, endpoint, credential, memory content, raw log, private state, or machine path is included.

PR #2037, opened 2026-07-13
fix(semantic-preference): bind OpenViking project actor

huangruiteng:
## Motivation

PR #2030 derives an exact project-peer URI, but OpenViking 0.4 also enforces the request actor-peer view. If a local CLI config names another actor, an exact peer URI is rejected before retrieval. That leaves the reusable provider dependent on an easy-to-miss local identity setting.

## Changes

- Invoke every provider `find` with `--actor-peer-id <derived-project-peer>`.
- Extend the focused smoke to prove correct-project, wrong-project, and global-fallback calls all use the derived actor.
- Clarify that native isolated writes need `memory_policy.self=false`, peer writes enabled, allowed memory types, and the message `peer_id`; speaker attribution alone does not force a peer write.

LoopX should own deterministic project routing end to end. OpenViking still owns extraction, update, supersede, ranking, and storage. Binding the actor at the same seam as the target URI removes a hidden integration precondition without adding a memory ledger.

## Validation

- Python compile: passed
- project-peer provider smoke: passed
- existing semantic-preference and Issue Fix PR-description smokes: passed
- ruff check / format: passed
- standard premerge canary: all selected checks passed; public boundary passed
- live OpenViking 0.4 probe: a CLI config with a different actor was safely overridden; exact project returned one preference and a different project returned zero

## Risk

OpenViking versions without actor-peer CLI support will fail through the existing provider failure policy. Existing default-off and fail-open behavior is unchanged.

### DIFF
--- a/docs/capabilities/semantic-preference/openviking-project-peer.md
+++ b/docs/capabilities/semantic-preference/openviking-project-peer.md
@@ -15,6 +15,7 @@ preference provider protocol for a function-owned application and receipt.
 - Identity comes from the normalized Git `origin`, never the checkout path.
 - A non-Git project must provide a stable `--loopx-project-id`.
 - Recall targets the exact project peer by default.
+- Each `find` binds the OpenViking request actor to the derived project peer.
 - User-global memory is available only through `--include-global-fallback`.
 - The default budget is one `find`; explicit global fallback needs at least two.
 - Only concrete preference nodes under the selected target are returned.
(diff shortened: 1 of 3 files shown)


huangruiteng:
## 自审结论

### Findings

无阻塞项。

### 产品与架构判断

- 真实 Pilot 暴露的不是召回排序问题，而是 OpenViking 0.4 actor-peer ACL：target URI 正确但 actor 不一致会先被拒绝。
- 修复将 actor 绑定与 target URI 派生放在同一 provider seam，避免要求每个 LoopX agent 手工同步两份 project identity。
- 原生写入仍归 OpenViking；文档补充 `self=false + peer=true + message peer_id`，防止抽取器把项目偏好合并回 user-global memory。

### 验证

- focused 与既有 smokes、compile、ruff、standard canary、public/private scan 全部通过。
- 真实 OpenViking 0.4：不同 actor 的本地配置被 provider 覆盖后，目标项目命中 1、其他项目命中 0。

### 合并决定

`approved after reviewer`。这是窄修复，但改变 runtime 命令调用；按 `AGENTS.md` 保留独立 review 门槛，不使用 admin bypass。

PR #2046, opened 2026-07-13
fix(explore): bound owner board by evidence stages

huangruiteng:
## Motivation

The svg_board renderer merged in #2045 preserved real relations but still projected every issue and capability as an oversized lane card. On a real OpenViking graph, the owner view became another large inventory rather than a readable evidence board.

## Changes

- render compact work nodes in vertical Evidence Stages
- cap each stage at 8-16 nodes, default 12, configurable with --stage-capacity
- keep structural area nodes as context and full detail in canonical Base
- preserve material edges, frontier state, metrics, and conclusions
- embed a delivery marker in SVG and verify it through remote whiteboard raw-node readback
- fail closed as publish_unverified when update succeeds but remote content is stale

## Validation

- 20 focused presentation and sink tests pass
- ruff passes on all changed files
- standard LoopX premerge gate passes
- 3 catalog canaries pass; no warnings or manual holds
- real OpenViking projection renders 38 compact nodes as stages 12/12/12/2 with 15 real relations and complete node-detail coverage
- local SVG visually inspected against the reference evidence timeline

## Risk

The new stage renderer is opt-in through svg_board. Mermaid and svg_atlas remain unchanged. Full canonical Nodes, Edges, and Findings remain in Base.

PR #2048, opened 2026-07-13
fix(explore): retry Lark visual readback while applying

huangruiteng:
## Motivation

Real OpenViking dogfood of #2046 showed that Lark returns API 4003101 while a whiteboard update is still applying. The delivery marker appears correctly moments later, but the immediate fail-closed readback produces a false negative.

## Changes

- retry only the explicit 4003101 doc-is-applying response
- bounded delays: 0.25, 0.5, 1, 2, and 4 seconds
- keep marker absence and all other errors fail-closed without retry
- include compact attempt evidence in the readback receipt

## Validation

- 21 focused tests pass, including applying-then-success and stale-marker failure
- ruff and diff checks pass
- standard LoopX premerge gate passes with no warnings or holds

This preserves the #2046 remote marker contract while accommodating the documented asynchronous Lark write window.

### DIFF
--- a/loopx/presentation/sinks/lark/explore_results.py
+++ b/loopx/presentation/sinks/lark/explore_results.py
@@ -18,6 +18,7 @@ import hashlib
 import html
 import json
 import re
+import time
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Mapping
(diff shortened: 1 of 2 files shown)


huangruiteng:
Maintainer review: approved for self-merge after CI. The retry scope is intentionally narrow: only Lark API 4003101 with the explicit doc-is-applying message is retryable, using a bounded 0.25/0.5/1/2/4-second schedule. Missing delivery markers and all other errors still fail closed. Real OpenViking dogfood reproduced the transient and independently confirmed the marker appears after applying completes. Focused tests: 21 passed; standard premerge gate passed with no warnings or holds.

PR #2049, opened 2026-07-13
fix(explore): parse Lark applying errors from stderr

huangruiteng:
## Motivation

Real Lark dogfood showed lark-cli writes API error JSON to stderr when whiteboard raw query returns 4003101. The bounded retry added in #2048 inspected stdout JSON only, so the real applying response was not classified as retryable.

## Changes

- parse structured Lark error payloads from stdout JSON or stderr JSON
- keep the retry predicate unchanged: code 4003101 plus doc-is-applying text
- update the regression test to use the actual stderr transport shape

## Validation

- 21 focused tests pass
- ruff and diff checks pass
- standard LoopX premerge gate passes with no warnings or holds

No renderer, Base, or retry-policy behavior changes beyond recognizing the real lark-cli error channel.

### DIFF
--- a/loopx/presentation/sinks/lark/explore_results.py
+++ b/loopx/presentation/sinks/lark/explore_results.py
@@ -580,6 +580,17 @@ def _whiteboard_raw_texts(payload: Any) -> list[str]:
     return texts
 
 
+def _structured_command_error(result: Mapping[str, Any]) -> Mapping[str, Any]:
+    parsed = result.get("json")
+    if not isinstance(parsed, Mapping):
+        try:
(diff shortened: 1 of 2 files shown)


huangruiteng:
Maintainer review: approved for self-merge after CI. This is a transport-shape correction proven by real OpenViking Lark dogfood: API 4003101 arrives as structured stderr JSON. The retry predicate remains narrowly scoped, other errors still fail closed, and the regression test now matches the real lark-cli response. Focused tests: 21 passed; standard premerge gate passed with no warnings or holds.

PR #2055, opened 2026-07-13
Fix monitor-poll runtime capability parity

huangruiteng:
## Summary

- document every quota command that recomputes `should-run` with runtime capability declarations
- return a structured `capability_retry` receipt when `monitor-poll` loses its route because declarations are missing
- cover the real paired flow: failure without declarations, success with identical declarations, and no retry noise for unrelated option errors

## Validation

- adjacent monitor and quota policy smokes
- `python3 -m compileall -q` on changed Python files
- `loopx check` on all changed paths
- `loopx canary premerge --from-git-diff --tier standard`: 17/17 selected checks passed

No failures, skips, manual holds, credentials, local paths, private state, or generated logs.

### DIFF
--- a/examples/control_plane/monitor-poll-writeback-smoke.py
+++ b/examples/control_plane/monitor-poll-writeback-smoke.py
@@ -70,6 +70,7 @@ def write_fixture(
     include_user_gate: bool = False,
     include_advancement: bool = False,
     primary_monitor_priority: str = "P0",
+    monitor_required_capabilities: tuple[str, ...] = (),
 ) -> tuple[Path, Path]:
     project = root / "project"
     runtime = root / "runtime"
@@ -80,6 +81,11 @@ def write_fixture(
     monitor_metadata = (
         f"target_key={selected_target_key} "
         if selected_target_key
+        else ""
+    )
+    monitor_capability_metadata = (
+        f"required_capabilities={','.join(monitor_required_capabilities)} "
+        if monitor_required_capabilities
         else ""
     )
     other_monitor = (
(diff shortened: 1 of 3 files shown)


huangruiteng:

- Motivation: capability-gated monitor work was correctly selected by `should-run`, but a later `monitor-poll` could silently lose that route when the caller omitted the same runtime declarations.
- Solved: the failure now exposes a provider-neutral `capability_retry` receipt with the exact missing declarations, while unrelated option errors remain unchanged. The CLI help names every quota command that recomputes `should-run`.
- Operator impact: agents can repair the invocation themselves instead of misclassifying an observed runtime capability as a user gate.
- Main risk: changing quota failure payloads could affect consumers. The change is additive, restricted to route-authorization failure, and preserves the existing `reason`, `before`, and return-code contract.
- Design: the retry contract lives in the monitor-poll bounded context and is derived from the canonical `capability_gate`; it does not duplicate capability selection rules.

**Changed surfaces**
- quota CLI help
- monitor-poll failure receipt
- focused monitor-poll writeback smoke

**Validation**
- paired capability-gated flow: fails clearly without declarations and succeeds with identical declarations
- unrelated option failures do not receive capability retry noise
- adjacent monitor/quota policy smokes passed
- changed Python files compile
- public boundary scan passed
- `loopx canary premerge --from-git-diff --tier standard`: 17/17 checks passed

Failures/skips: none. Manual holds: none. Private/local artifacts: none.

Merge decision: self-merge after GitHub `pytest` passes; coverage is sufficient because it exercises the exact live command pair plus risk-selected control-plane, CLI, scheduler, architecture-budget, canary-runner, and public-boundary checks.

PR #2060, opened 2026-07-14
Reconcile scheduler hints with Codex App cadence

huangruiteng:
## Summary

- observe the RRULE of the uniquely matched active Codex App heartbeat during `quota should-run`
- let host reality override the scheduler ACK ledger when deciding `apply_needed`
- re-emit cadence repair on stale/premature ACK or later host drift, while keeping writes behind `automation_update`
- document the bounded readback contract and add real CLI regression coverage

## Validation

- `python3 -m py_compile` for all changed Python files
- `examples/control_plane/heartbeat-prompt-smoke.py`
- `examples/control_plane/quota-contract-smoke.py`
- `examples/control_plane/quota-plan-smoke.py`
- `examples/upgrade-plan-smoke.py`
- `loopx canary premerge --from-git-diff --tier standard`: 17/17 checks passed; public boundary clean; self-merge allowed

## Boundary

No automation prompt body, credentials, local goal state, raw logs, or generated `uv.lock` are included. The resolver returns only a unique active RRULE matched by goal, agent, and current thread.

## Residual risk

If there is no unique matching active heartbeat, LoopX falls back to the existing ACK-ledger behavior rather than guessing.


huangruiteng:
## Maintainer review

**Findings:** none blocking.

**Assumption:** automatic host readback is used only when one active heartbeat matches goal + agent + current thread. Missing/ambiguous matches fail open to the existing ACK ledger.

**Product / architecture judgment:** The bug is a split-brain between LoopX scheduler intent and Codex App host reality. Treating the ACK ledger as proof of host state made a premature ACK or later App-side cadence change permanently suppress repair. This PR adds a reusable, bounded observation seam: host RRULE overrides the ledger for `apply_needed`, while all writes remain behind `automation_update`. It reads no prompt body into the hot payload and does not broaden authority. The main residual risk is an unmatched legacy heartbeat, which preserves existing behavior rather than guessing.

**Validation:** targeted scheduler, heartbeat, quota, upgrade, syntax, architecture-budget, and public-boundary smokes pass. `loopx canary premerge --from-git-diff --tier standard` passed 17/17 checks with `self_merge_allowed=true`. The live OpenViking heartbeat reproduced `drift_detected` at 30m vs 3m, then returned `matches_recommended` after the official App update.

**Merge decision:** self-merge after GitHub CI succeeds.

PR #2063, opened 2026-07-14
Project stage SVG sources into visual sync

huangruiteng:
## Summary

- include each Evidence Stage SVG source when stage-document sync builds its per-board projection
- exercise both first-class board styles through the full multi-stage sync path
- preserve fail-closed behavior when a selected style lacks its source

## Root cause

The lower-level SVG publish path was covered, but the stage-document integration copied only each stage Mermaid source into the display projection. Real OpenViking dogfood therefore returned `invalid_projection` before any whiteboard write. Base row readback remained correct.

## Validation

- `pytest -q tests/test_explore_presentation_views.py -x` (27 passed)

Local `uv.lock` remains untracked and excluded.

### DIFF
--- a/loopx/presentation/sinks/lark/explore_results.py
+++ b/loopx/presentation/sinks/lark/explore_results.py
@@ -991,6 +991,7 @@ def sync_explore_visuals_to_lark(
             stage_node_ids = {str(item) for item in stage.get("node_ids") or []}
             stage_projection = dict(role_view)
             stage_projection["mermaid"] = str(stage.get("mermaid") or "")
+            stage_projection["svg"] = str(stage.get("svg") or "")
             stage_projection["nodes"] = [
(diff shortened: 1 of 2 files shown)


huangruiteng:
## Maintainer review

### Findings

No remaining blocking findings. The one-line product fix is paired with a full-path regression matrix that fails before the fix for `semantic_lane_columns`.

This is a narrow integration repair, not a relaxation: stage-document sync now transports the source already produced by the selected board style. Missing style sources still fail closed before external writes. The test covers both `auto_flow` and `semantic_lane_columns`, so renderer-specific direct-call coverage can no longer hide stage assembly drift.

### Validation

- 27 focused pytest cases passed
- risk-based premerge gate passed with no failures, skips, boundary findings, or manual holds

### Merge decision

PR #2065, opened 2026-07-14
Retry Explore visual marker readback until visible

huangruiteng:
## Summary

- treat a successful whiteboard query without the expected delivery marker as an eventual-consistency state
- retry marker readback with the existing bounded backoff schedule
- keep non-retryable query errors fail-closed and add focused regression coverage

## Root cause

Lark can return a successful raw whiteboard query before the just-published marker is visible. The readback loop previously stopped on any successful query, so a transient stale result surfaced as `publish_unverified` even though a following read contained the marker.

## Validation

- `pytest -q tests/test_explore_presentation_views.py -x` (28 passed)

Local `uv.lock` remains untracked and excluded.

### DIFF
--- a/loopx/presentation/sinks/lark/explore_results.py
+++ b/loopx/presentation/sinks/lark/explore_results.py
@@ -614,16 +614,18 @@ def _readback_visual_delivery_marker(
         error_code = error.get("code")
         error_message = str(error.get("message") or "")
         is_applying = error_code == 4003101 and "doc is applying" in error_message
+        is_marker_pending = bool(result.get("ok")) and not marker_observed
+        is_retryable = is_applying or is_marker_pending
         attempts.append(
             {
                 "attempt": attempt_index + 1,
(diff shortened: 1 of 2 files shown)


huangruiteng:
## Maintainer review

### Findings

No remaining blocking findings. The change reuses the existing bounded retry budget and only broadens the retryable state from Lark's explicit applying error to a successful-but-stale marker read.

This preserves fail-closed delivery semantics: a marker that never appears still fails after the bounded retry schedule, while unrelated query errors still stop immediately. No extra whiteboard writes are introduced; only raw readback is retried. The regression proves the exact eventual-consistency sequence observed in dogfood and the existing missing-marker case still verifies terminal failure.

### Validation

- 28 focused pytest cases passed
- risk-based premerge gate passed with no failures, skips, or manual holds
- private/local `uv.lock` excluded

### Merge decision

PR #2066, opened 2026-07-14
Fit semantic lane text within SVG nodes

huangruiteng:
## Summary

- bound semantic-lane SVG node headers by display width while reserving space for the status suffix
- use a narrower display-width budget for detail lines so mixed CJK/ASCII text stays inside the real Lark node box
- add a mixed-width regression that verifies header/detail bounds and preserved status

## Root cause

The fixed-lane renderer truncated the title before appending `ACTIVE`/`DONE`, and used display-width budgets that were wider than the 436px node at Lark's actual Arial font metrics. Remote Stage 2/3 previews therefore showed long text crossing node borders even though the SVG coordinates themselves were valid.

## Validation

- `pytest -q tests/test_explore_presentation_views.py -x` (29 passed)

The optional external `whiteboard-cli` geometry rerun was unavailable from the configured npm registry; the durable mixed-width renderer regression covers the observed failure. Local `uv.lock` remains untracked and excluded.

### DIFF
--- a/loopx/presentation/explore_views.py
+++ b/loopx/presentation/explore_views.py
@@ -89,6 +89,8 @@ _SVG_STATUS_STYLE = {
     "resolved": ("#e8f5e9", "#43a047"),
     "dead_end": ("#eeeeee", "#9e9e9e"),
 }
+_SVG_NODE_HEADER_DISPLAY_WIDTH = 50
+_SVG_NODE_DETAIL_DISPLAY_WIDTH = 62
 
 
 def build_stage_lane_svg(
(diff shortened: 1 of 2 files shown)


huangruiteng:
## Maintainer review

### Findings

No remaining blocking findings. The fix is localized to the deterministic SVG style and does not change graph selection, stage membership, edge routing, or the `auto_flow` renderer.

The text-fit contract belongs in the semantic renderer: it measures East Asian display width, reserves header capacity for the status suffix, and keeps evidence lines bounded without deleting canonical data. This is preferable to shrinking nodes globally or weakening remote visual acceptance.

### Validation

- 29 focused pytest cases passed, including a mixed CJK/ASCII overflow regression
- risk-based premerge gate passed with no failures or manual holds
- external `whiteboard-cli` geometry rerun skipped because the configured npm registry could not resolve the package; actual Lark re-publication remains the post-merge acceptance gate
- private/local `uv.lock` excluded

### Merge decision

PR #2074, opened 2026-07-14
Scope quota health to the selected goal

cocolord:
## What changed

- classify active-state todo contract failures by owning goal while preserving global registry, routing, and public-boundary failures
- make per-goal quota health use the selected goal contract errors without losing the global quota plan
- extend the goal-filter smoke to prove a healthy goal can run while an unrelated invalid goal remains visible and blocked
- document the goal-scoped health contract

## Root cause

quota should-run with goal A reused the global status health bit. A legacy todo contract error in goal B therefore forced goal A into quota_skip even when goal A had an eligible advancement todo. This also produced a contradictory payload where the work lane required an attempt but the interaction contract prohibited delivery.

## Impact

Per-goal heartbeat, monitor, scheduler, and accounting decisions no longer stall on another goal active-state todo semantics. Broad status and quota inventory still report every error, and global structural or public-boundary failures remain fail-closed.

## Validation

- live local-private registry check: target decision changed from quota_skip to normal_run while the global plan retained the unrelated health blocker
- python3 examples/control_plane/status-goal-filter-smoke.py
- python3 examples/control_plane/status-collection-readmodel-smoke.py
- python3 examples/control_plane/quota-plan-smoke.py
- python3 examples/control_plane/status-projection-cache-smoke.py
- loopx canary premerge --from-git-diff: 17/17 selected checks passed, public boundary clean, no manual holds

### DIFF
--- a/docs/quota-allocation.md
+++ b/docs/quota-allocation.md
@@ -538,6 +538,12 @@ the displayed gate, evidence, or health reason.
 `quota should-run` is the per-goal guard for heartbeat jobs. It returns a small
 JSON or Markdown decision:
 
+The per-goal guard, monitor, scheduler acknowledgment, and slot-accounting
+commands isolate active-state todo contract errors to their selected goal. A
+stale todo in another goal therefore remains visible in broad health inventory
+without blocking an otherwise healthy goal. Global registry, routing, and
+public-boundary failures still fail closed for every goal.
+
 ```json
 {
   "goal_id": "project-main-control",
--- a/docs/status-data-contract.md
+++ b/docs/status-data-contract.md
@@ -650,6 +650,11 @@ enough to support an operator decision without exposing local paths or private
 evidence. They must be public-safe before a project exposes this export outside
 the local machine.
 
+Goal-filtered status and quota guards validate active-state todo semantics only
+for the selected goal. Registry structure, routing ambiguity, and public-boundary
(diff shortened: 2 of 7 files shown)


huangruiteng:
Thank you, @cocolord. This PR identified a real multi-goal isolation bug: an active-state contract error owned by an unrelated goal must remain visible in broad health inventory without blocking an otherwise healthy selected goal. The positive and negative cases in this PR helped make that boundary concrete.

The direction has since been fully superseded by #2548, merged as `33a9f8f4`. That implementation moves ownership to source-produced structured diagnostics with explicit `global`, `goal`, and known `goals` scopes, then projects global plus exact selected-goal failures for status, quota, turn, diagnose, review-packet, and ready-score consumers. It avoids inferring ownership from human-readable message prefixes and also covers goal-owned registry findings and shared multi-goal findings.

I rechecked both states before closing:

- exact #2074 head `6f1ad99a`: the four focused status/quota smokes pass;
- current `main`: the seven focused contract-health tests and the status-goal-filter/quota-plan smokes pass;
- #2074 is now conflicting with `main`, and its shipped behavior and regression coverage have no remaining unique increment beyond #2548.

Closing this as superseded by #2548, with thanks for finding and proving the original problem. The idea was valuable and is now part of the current control-plane contract.

PR #2099, opened 2026-07-14
Fix source-mirror ambiguity in runtime projection routes

huangruiteng:
## Summary\n\n- prefer external shared-runtime matches when route discovery also sees the source runtime registry mirror\n- preserve fail-closed ambiguity when more than one external target remains\n- expose the ignored source-mirror count in compact diagnostics\n- cover one-external and two-external routes, and record the reusable self-repair pattern\n\n## Product and architecture judgment\n\nThe operator problem is a false ambiguous route: a material local refresh succeeds, but shared-runtime projection is withheld because the source mirror is counted as a second destination. The change keeps discovery generic and corrects identity at the routing boundary instead of adding retry behavior. Multiple real external destinations still fail closed.\n\nThe main risk is hiding a genuine competing destination. The regression fixtures distinguish source mirror plus one external target from source mirror plus two external targets.\n\n## Validation\n\n- focused shared-runtime projection smoke: passed\n- changed Python compile checks: passed\n- standard risk-based premerge: 17 selected and 17 passed\n- failures, skips, warnings, and manual holds: none\n- public/private boundary scan: clean\n\nNo local state, raw runtime evidence, credentials, or private paths are included.


huangruiteng:
## Findings\n\nNo actionable findings.\n\n## Open questions or assumptions\n\nNone. The source runtime mirror is diagnostic metadata, not an independent projection destination; multiple external destinations remain ambiguous and fail closed.\n\n## Product and architecture judgment\n\nThe change fixes route identity at the shared runtime routing boundary. It avoids retry-based masking and does not introduce project-specific policy. The user impact is that material refreshes can reach the one registered shared runtime while genuine multi-target conflicts remain visible.\n\nChanged surfaces: control-plane runtime routing, focused refresh-state projection smoke, and self-repair pattern documentation.\n\n## Validation\n\n- focused shared-runtime projection smoke passed\n- changed Python compile checks passed\n- standard premerge selected 17 checks and all 17 passed\n- failures: none\n- skips: none\n- warnings: none\n- manual holds: none\n- public/private boundary scan clean\n\nCoverage is sufficient because it exercises source mirror plus one external target, source mirror plus two external targets, state-write correctness, todo and active-state projection, canary infrastructure, diff hygiene, and boundary scanning.\n\n## Merge decision\n\nSelf-merge after the GitHub check is green. The owner has standing authorization for these LoopX PRs, and the local risk gate reports self merge allowed.

PR #2101, opened 2026-07-14
fix: retry transient explore marker readback

huangruiteng:
## Why

A material Explore Graph refresh updated and verified all canonical rows, but one post-publish whiteboard marker query briefly returned Lark API 2890002 invalid arg. LoopX stopped immediately and the aggregate visual receipt incorrectly reported retryable=false.

## What changed

- treat 2890002 invalid arg as bounded-retryable only inside the post-publish raw-node marker readback path
- preserve the existing finite retry delays and do not change general Lark API error semantics
- propagate retryable and required_action from stage to role and top-level visual receipts
- extract the small readback parsing/classification helpers so explore_results.py remains under its 2000-line budget

## Validation

- focused marker retry smoke proves the first 2890002 response retries and the next marker read succeeds
- Explore result, issue-fix projection, executive projection, configure-goal, line-budget, and boundary smokes pass
- standard diff-driven premerge: 4 selected plus 4 direct checks, 0 failures, 0 warnings, 0 holds
- real OpenViking canary: 590/590 rows verified, 0 missing/mismatch/duplicates, canonical 14/14 stages published and marker-verified; several stages required 2-3 bounded readback attempts, including the previously failed Stage 3 path

### DIFF
--- a/examples/explore-result-layer-smoke.py
+++ b/examples/explore-result-layer-smoke.py
@@ -39,6 +39,7 @@ from loopx.capabilities.explore.worker_branch_plan import (  # noqa: E402
     build_explore_worker_branch_plan,
 )
 from loopx.presentation.sinks.lark import explore_results  # noqa: E402
+from loopx.presentation.sinks.lark import explore_visual_styles  # noqa: E402
 
 # Both exploration planners are deny-by-default behind the per-goal
 # goal_boundary.orchestration.explore_harness gate; every library-level plan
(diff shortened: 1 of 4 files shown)


huangruiteng:
Maintainer self-review

- Findings: none.
- Assumptions: Lark API 2890002 invalid arg is transient only in the narrow context where a whiteboard overwrite already succeeded and the immediately following raw-node marker query is settling.
- Product / architecture judgment: reuse the existing finite post-publish retry loop; do not generalize 2890002 across Lark operations. Retryability must remain machine-visible from stage through role to the top-level sink receipt.
- Changed surfaces: Explore visual marker readback parsing/classification and receipt aggregation; no Base schema, board style, renderer, public CLI, or scheduler change.
- Tests: focused retry smoke; Explore result, issue-fix projection, executive projection, configure-goal, line-budget, and boundary smokes; standard premerge passed 4 selected plus 4 direct checks.
- Real-callsite evidence: OpenViking canary verified 590/590 rows and 14/14 semantic_lane_columns stages with zero duplicate/missing/mismatched rows; several marker reads settled after 2-3 attempts.
- Failures / skips / manual holds: none after the helper extraction; 0 warnings.
- Coverage judgment: sufficient because the exact error code and bounded retry transition are deterministic in the smoke, while the live sink proves multi-attempt marker verification.
- Merge decision: self-merge after required GitHub pytest succeeds.

PR #2116, opened 2026-07-14
Scope autonomous replans to the causal peer frontier

huangruiteng:
Summary:
- Preserve the unique agent identity on run-history replan evidence.
- When state-authored replan evidence has no explicit actor, bind it to the claimed first agent frontier instead of broadcasting it across registered peers.
- Keep genuinely unscoped or conflicting evidence on the existing deterministic coordination path.

Validation:
- autonomous-replan obligation read-model smoke
- quota replan decision-plane smoke
- heartbeat quota flow, work-lane, and integrated control-plane smokes
- real registry source-CLI parity check for peer isolation
- LoopX standard premerge gate: 17/17 checks passed

Risk:
The change only adds causal attribution to the existing replan payload. Conflicting explicit agent evidence remains unscoped, and existing deterministic assignment semantics are preserved for truly unowned replans.

### DIFF
--- a/examples/control_plane/autonomous-replan-obligation-readmodel-smoke.py
+++ b/examples/control_plane/autonomous-replan-obligation-readmodel-smoke.py
@@ -35,6 +35,7 @@ AGENT_TODOS = {
         {
             "priority": "P2",
             "text": "[P2] Continue canary-gated control-plane read-model cleanup.",
+            "claimed_by": "codex-control-plane",
         }
     ]
 }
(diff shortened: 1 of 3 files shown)


huangruiteng:
Self-review result: no blocking findings.

Product and architecture judgment:
- Motivation: a goal-level replan derived from one claimed peer frontier could be deterministically assigned to an unrelated monitor-only peer.
- Solved: the obligation now preserves a unique run-history actor and falls back to the claimed first agent frontier for state-authored evidence. Conflicting actors remain genuinely unscoped, so the deterministic coordination path is preserved.
- Operator impact: peer heartbeats stop consuming turns repairing another peer lane while still retaining one owner for truly unowned replans.
- Main risk: attribution must not guess across conflicting evidence. The implementation explicitly leaves that case unscoped.
- Design: this extends the existing autonomous_replan_obligation actor contract instead of adding a project-specific routing rule.

Changed surfaces:
- autonomous replan obligation read model
- quota replan decision-plane regression coverage
- focused read-model parity smoke

Validation:
- focused read-model and decision-plane smokes passed
- heartbeat quota flow, work-lane, and integrated control-plane smokes passed
- real source-CLI peer-isolation parity check passed
- standard premerge gate passed 17/17
- GitHub pytest is still in progress

Failures/skips: none.
Manual holds: none.
Merge decision: self-merge after GitHub pytest succeeds; otherwise hold and inspect the failure.

PR #2117, opened 2026-07-15
fix: suppress repeated scheduler host update failures

huangruiteng:
## Summary
- persist failed Codex App scheduler `(target RRULE, observed host RRULE)` pairs
- suppress an exact repeated host update until the target or host observation changes
- add `quota scheduler-fail-current`, bound failure hints, turn-envelope projection, and host-match ACK recovery
- update heartbeat prompts, project skill, quota docs, and the self-repair catalog

## Validation
- `python3 examples/control_plane/quota-scheduler-state-ack-smoke.py`
- `python3 examples/control_plane/quota-scheduler-hint-compaction-smoke.py`
- `python3 examples/control_plane/quota-scheduler-policy-smoke.py`
- `python3 examples/control_plane/heartbeat-prompt-smoke.py`
- `uv run --python 3.11 --with 'pytest>=8,<9' pytest -q tests` (`318 passed, 2 skipped`)
- changed-file Ruff checks passed for CI-covered paths
- `loopx canary premerge --from-git-diff` (`17/17`, `self_merge_allowed=true`)

## Boundaries
- no automation body, private state, raw host logs, credentials, local paths, or generated artifacts are included
- scheduler failure writeback is no-spend and does not ACK an unconfirmed RRULE
- exact-pair suppression is released only by a target/host change or successful matching host ACK

## Residual Risk
- the host update tool can still fail; this change prevents repeated hangs and leaves the observed cadence authoritative rather than bypassing the host boundary

PR #2118, opened 2026-07-15
test: align scheduler failure smoke contract

huangruiteng:
## Summary

- update the quota plan smoke for the persisted scheduler host-failure contract introduced in #2117
- assert durable behavior fields instead of the retired turn-local failure string

## Validation

- `python3 examples/control_plane/quota-plan-smoke.py`
- full-public shard offset 200, limit 100
- `loopx canary premerge --from-git-diff` (5/5 selected checks passed; self-merge allowed)

Fixes the Full Public Smokes failure on main run 29380874783.

### DIFF
--- a/examples/control_plane/quota-plan-smoke.py
+++ b/examples/control_plane/quota-plan-smoke.py
@@ -1044,9 +1044,10 @@ def assert_heartbeat_recommendation_lifecycle() -> None:
     assert local_scheduler["max_interval_minutes"] == 240, local_scheduler
     assert stateful_detail["host_max_interval_minutes"] == 60, stateful_detail
     assert stateful_detail["coarser_wait_fallback"] == "local_scheduler_only", stateful_detail
-    assert stateful_detail["host_update_failure"] == (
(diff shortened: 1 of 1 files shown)


huangruiteng:
Self-merge evidence:

- Changed surface: one durable public control-plane smoke (`examples/control_plane/quota-plan-smoke.py`); no runtime behavior changed.
- Root signal: Full Public Smokes main run 29380874783 failed in shard 2 because the smoke expected the retired turn-local `host_update_failure` string after #2117 added persisted pair suppression.
- Focused checks: quota-plan smoke passed; CI-equivalent full-public shard offset 200/limit 100 passed.
- Risk gate: `loopx canary premerge --from-git-diff` passed 5/5 selected checks plus 4 direct checks; `self_merge_allowed=true`.
- Boundary: public/private scan clean; no credentials, private state, local paths, raw logs, or generated evidence included.
- Failures/skips/manual holds: none.
- Coverage is sufficient because the patch only updates the failing contract assertion and the exact failing full-public shard now passes end to end.

PR #2125, opened 2026-07-15
Fix batched Explore visual marker readback

huangruiteng:
## 动机

Stage-document Explore sinks currently apply the full marker settling delay independently to every whiteboard. A multi-stage material refresh can therefore take minutes and leave a partially published document when interrupted.

## 改动思路

Separate visual marker readback from the main sink and make stage documents use a two-phase delivery: publish every configured stage first, then verify only pending markers within one shared settling window.

## 关键代码或伪代码

```text
publish all stages with deferred readback
pending = every successfully published stage
for delay in shared_settling_window:
    query each pending stage once
    remove stages whose exact delivery marker is observed
    stop when pending is empty
```

## 具体改动

- Move marker query/receipt handling into `explore_visual_readback.py`.
- Allow a single marker query to receive an explicit retry schedule.
- Defer stage-document readback until all stage writes have been attempted.
- Preserve per-stage attempt counts, exact marker receipts, retryability, and failure status.
- Add a regression proving two pending stages share one sleep window instead of sleeping once per stage.

## 修复后复现

Run `python3 examples/explore-result-layer-smoke.py`. The batch fixture makes both stage markers settle on the second query and asserts one shared delay, two attempts per stage, and verified publication for both stages.

## 验证

- `python3 examples/explore-executive-projection-contract-smoke.py`
- `python3 examples/explore-todo-result-node-linkage-smoke.py`
- `./scripts/loopx canary premerge --from-git-diff` — passed, 4 selected checks, 0 failures, 0 manual holds

## 对主干的风险与未覆盖

The retry schedule and exact marker postcondition are unchanged; only the scope of the wait changes from per-stage to per-batch. Provider latency can still make the batch retryable after the bounded window, but every stage retains an independent readback receipt and no successful query is inferred from write acceptance alone.

### DIFF
--- a/examples/explore-result-layer-smoke.py
+++ b/examples/explore-result-layer-smoke.py
@@ -39,6 +39,7 @@ from loopx.capabilities.explore.worker_branch_plan import (  # noqa: E402
     build_explore_worker_branch_plan,
 )
 from loopx.presentation.sinks.lark import explore_results  # noqa: E402
+from loopx.presentation.sinks.lark import explore_visual_readback  # noqa: E402
 from loopx.presentation.sinks.lark import explore_visual_styles  # noqa: E402
 
 # Both exploration planners are deny-by-default behind the per-goal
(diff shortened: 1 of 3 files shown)


huangruiteng:
## Independent self-review

1. **Findings**
   - Confirmed the prior failure mode: the full marker retry schedule was applied once per Stage, so a 15-Stage refresh could take minutes and leave a partial document when interrupted.
   - Resolved by publishing every Stage before one shared settling loop; no remaining blocking finding on `4e55a5e8`.

2. **Assumptions**
   - Exact per-whiteboard delivery markers remain the authoritative post-write condition.
   - A Stage whose marker is still absent after the shared bounded window remains retryable; write acceptance alone is never treated as delivery success.

3. **Product / architecture judgment**
   - Retry policy now scales with the batch settling window rather than `stage_count × settling_window`.
   - Marker parsing and retry receipts live in the dedicated readback module; the main sink stays below the repository Python line budget and only orchestrates publish-then-verify.
   - The change preserves singleflight, idempotent updates, per-Stage attempt receipts, and the existing checkpoint contract.

4. **Validation**
   - Four focused Explore smokes passed, plus Python compile and diff checks.
   - Standard pre-merge gate passed: 4 selected checks, 0 failures, 0 manual holds.
   - GitHub `pytest` passed.
   - Real configured callsite: completed in about 22 seconds; Base rows 634/634 skipped idempotently, 0 duplicate rows, 0 mismatched Result IDs; canonical visual 15/15 Stage markers verified on the first shared readback round; persisted canonical delivery digest matches the current dry-run.

5. **Merge decision**
   - This is a bounded LoopX auxiliary self-repair with clean validation and exact remote readback. Owner has granted standing self-merge authority for this class; proceeding with squash self-merge.

PR #2135, opened 2026-07-15
Guard Lark Explore source runtime routing

huangruiteng:
## Summary
- resolve goal-scoped Lark Explore projections through the registry-declared canonical source registry
- reject append-only projection shrinkage before preview or remote writes using registered result-id coverage
- expose compact hashed route and guard diagnostics in the Kanban receipt
- add a split-runtime regression smoke and reusable self-repair pattern

## Validation
- `python3 examples/lark-explore-source-runtime-route-smoke.py`
- `python3 -m py_compile ...` for changed Python files
- `uvx ruff check ...` for changed Python files
- `loopx canary premerge --from-git-diff --tier standard`: 7/7 selected checks passed, public boundary clean, no manual holds
- real OpenViking dry-run: 89 candidate result ids vs 721 registered; regression blocked with zero connector writes

## Boundary
No private state, local paths, raw logs, credentials, or production connector output is included. The smoke uses public synthetic fixtures only.

## Residual work
The OpenViking source/shared Explore logs still require an explicit reconciliation batch; this PR prevents remote regression but does not guess or overwrite canonical history.


huangruiteng:
## Self-review

**Findings:** none.

**Assumption checked:** Explore result logs are append-only, so a candidate projection that omits result ids already registered by the sink is a source/runtime divergence, not a legitimate deletion. The guard therefore fails closed before preview or connector writes.

**Product / architecture:** The motivation is to prevent a goal-scoped Lark sync from reading the invoked runtime when the registry declares another canonical source, then regressing a fresher board. The implementation reuses the existing generic `goal_source_runtime_route_v0`, adds a bounded sink guard module rather than growing the hot Explore sink file beyond its line budget, and exposes only compact hashed route diagnostics. No OpenViking-specific value is encoded. Residual source-log reconciliation remains separate and explicit.

**Validation:**
- source-runtime/zero-write regression smoke passed
- existing issue-fix Explore projection smoke passed
- Lark Kanban control-plane + schema smokes passed
- Python compile and Ruff passed
- repository Python line-budget smoke passed
- standard premerge canary: 7/7 selected checks passed, public/private boundary clean, no manual holds
- GitHub pytest passed
- real dry-run failed closed at 89 candidate vs 721 registered result ids with zero connector writes

**Merge decision:** self-merge authorized by the owner and allowed by the premerge gate; merging with admin bypass.

PR #2139, opened 2026-07-15
Add safe Explore source history reconciliation

huangruiteng:
## Summary
- add a preview-first `explore source-history-reconcile` command for split-runtime recovery
- strictly rebuild and validate public-safe candidate events before append-only import
- recover only sink-registered missing results, classify remaining stale materialized parent edges, and never delete remote rows
- add idempotent locked batch append and focused regression coverage

## Validation
- `pytest -q tests/test_explore_source_history_reconcile.py tests/control_plane/test_explore_source_runtime_route.py` (7 passed)
- `python examples/explore-result-layer-smoke.py`
- `python examples/lark-explore-source-runtime-route-smoke.py`
- `loopx canary premerge --from-git-diff` (8/8 selected canaries passed; no manual holds)
- real split-runtime preview: 53 registered gaps classified as 47 recoverable lost-history projections and 6 stale materialized parent edges, with zero unexpected new projections

## Boundary
- no runtime logs, registry state, connector configuration, credentials, local paths, or private evidence committed
- command returns only counts and digests for source reconciliation evidence
- remote deletion remains an explicit separate action

## Merge decision
Self-merge authorized by the owner; premerge gate reports `self_merge_allowed=true`.

### DIFF
--- a/loopx/capabilities/explore/result_log.py
+++ b/loopx/capabilities/explore/result_log.py
@@ -473,8 +473,9 @@ def append_explore_result_event(path: Path, event: Mapping[str, Any]) -> dict[st
     validated = validate_explore_result_event(event)
     log_path = path.expanduser()
     log_path.parent.mkdir(parents=True, exist_ok=True)
-    with log_path.open("a", encoding="utf-8") as handle:
-        handle.write(json.dumps(validated, ensure_ascii=False, sort_keys=True) + "\n")
+    with exclusive_file_lock(log_path):
+        with log_path.open("a", encoding="utf-8") as handle:
+            handle.write(json.dumps(validated, ensure_ascii=False, sort_keys=True) + "\n")
     return {
         "ok": True,
         "schema_version": EXPLORE_RESULT_EVENT_SCHEMA_VERSION,
--- a/loopx/capabilities/explore/result_log.py
+++ b/loopx/capabilities/explore/result_log.py
@@ -20,6 +20,8 @@ from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any, Mapping, Sequence
 
+from ...file_lock import exclusive_file_lock
+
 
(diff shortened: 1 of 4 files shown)


huangruiteng:
## Maintainer review

### Findings
- No remaining blockers.
- During self-review I found that batch reconciliation held the result-log lock while ordinary single-event appends did not. Commit `11cbccf4` now makes both paths use the same lock; the focused tests and premerge gate were rerun after that fix.

- **Motivation:** split runtimes can leave a canonical Explore source behind an already-registered sink baseline, causing every later graph sync to fail closed.
- **Solved:** the new preview-first command recovers only public-safe, sink-registered missing results from an explicit candidate runtime, validates every event by rebuilding it through the canonical builders, and verifies append-only readback.
- **Operator impact:** safe lost-history recovery can proceed without being blocked by a separate destructive orphan-deletion decision; residual derived rows are classified rather than silently copied or deleted.
- **Main risk:** importing stale or unrelated history. The registered-baseline allowlist, strict event validation, unexpected-row rejection, idempotent batch append, and counts/digests-only receipt bound that risk.
- **Design:** this is a reusable source-reconciliation capability and CLI contract, not a project-specific path or fixture. Remote deletion remains a separate explicit action.

### Validation
- focused pytest: 7 passed
- Explore result-layer and source-runtime-route smokes passed
- post-fix `loopx canary premerge --from-git-diff`: 8/8 selected canaries passed, 0 failures, 0 warnings, no manual holds, `self_merge_allowed=true`
- real preview classified 53 registered gaps as 47 recoverable projections plus 6 stale materialized parent edges, with 0 unexpected projections
- public/private scan found no committed local state, logs, connector values, private URLs, or absolute paths

### Merge decision
Self-merge after the GitHub pytest check completes successfully; owner authorization is recorded.

PR #2140, opened 2026-07-15
Keep full findings in issue-fix Explore sync

huangruiteng:
## Summary
- build the automatic issue-fix Explore projection with the complete finding set
- keep the default bounded finding limit for ordinary display callers
- add a public smoke proving more than 200 findings survive the issue-fix sync projection

## Motivation
The source coverage guard compared a registered append-only sink baseline against an intentionally truncated 200-finding display projection. A healthy canonical source could therefore be misreported as missing rows after source reconciliation.

## Validation
- `loopx canary premerge --from-git-diff`: 6/6 selected checks passed, including public-boundary validation; no warnings or manual holds; `self_merge_allowed=true`

## Boundary
Only generic public fixture data is added. No runtime state, private evidence, credentials, connector values, or local paths are included.

## Merge decision
Self-merge authorized by the owner after GitHub pytest passes.

### DIFF
--- a/examples/issue-fix-explore-projection-smoke.py
+++ b/examples/issue-fix-explore-projection-smoke.py
@@ -25,6 +25,11 @@ from loopx.capabilities.issue_fix.pr_lifecycle import (  # noqa: E402
 from loopx.capabilities.explore.activation import (  # noqa: E402
     sync_explore_graph_after_material_refresh,
 )
+from loopx.capabilities.explore.result_log import (  # noqa: E402
+    append_explore_result_event,
+    build_explore_finding_event,
+    explore_result_log_path,
+)
 from loopx.domain_packs.issue_fix import (  # noqa: E402
     default_issue_fix_domain_state_ledger_path,
     default_issue_fix_feasibility_ledger_path,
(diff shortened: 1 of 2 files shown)


huangruiteng:
## Maintainer review

### Findings
No blockers.

- **Motivation:** the automatic issue-fix sync was applying the generic 200-finding display bound before append-only source coverage validation, so complete sources could look regressive.
- **Solved:** this call site now explicitly requests the complete finding set; a 205-finding public fixture proves the regression boundary.
- **Operator impact:** source reconciliation no longer creates a false second wave of missing findings.
- **Main risk:** a larger in-memory projection on the automatic sync path. That path already needs the complete registered key set for correctness, while ordinary summary/display callers remain bounded.
- **Design:** the fix is localized to the completeness-sensitive call site rather than changing the generic projection default.

### Validation
- focused issue-fix Explore smoke passed
- premerge gate: 6/6 selected checks passed, including public-boundary validation; 0 warnings, no manual holds, `self_merge_allowed=true`
- diff hygiene passed

### Merge decision
Self-merge after GitHub pytest succeeds; owner authorization is recorded.

PR #2154, opened 2026-07-15
Make CLI output smokes self-contained

huangruiteng:
## Summary

- bootstrap the repository import path before the differential smoke imports LoopX
- execute the existing CLI output budget assertions with the standard library in public smokes
- keep pytest coverage while removing the public smoke dependency on pytest or uvx

## Validation

- focused ruff checks
- 14 focused pytest tests
- both affected smoke scripts in a no-pytest/uvx environment
- full-public shard offset 200 limit 100: 100/100 passed
- LoopX premerge canary standard tier: passed, self-merge allowed

### DIFF
--- a/examples/control_plane/cli-output-base-head-differential-smoke.py
+++ b/examples/control_plane/cli-output-base-head-differential-smoke.py
@@ -12,12 +12,16 @@ import sys
 import tempfile
 from pathlib import Path
 
-from loopx.control_plane.testing.cli_output_differential import (
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+if str(REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(REPO_ROOT))
+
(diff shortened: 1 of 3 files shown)


huangruiteng:

Product judgment: the regression was in the public validation boundary, not terminal-closure behavior. The differential smoke assumed the repository root was importable, while the budget smoke required optional pytest/uvx tooling that Full Public Smokes intentionally does not install. This PR makes both entries self-contained and reuses the existing assertion matrix; it adds no runtime behavior or compatibility layer.

Validation: focused Ruff passed; 14 focused pytest cases passed; both affected scripts passed with pytest/uvx unavailable; the matching full-public shard passed 100/100 with zero failures, timeouts, warnings, or tracked side effects; standard premerge canary and public-boundary checks passed. Coverage is sufficient because it reproduces both CI failures under the same dependency boundary and reruns their full neighboring shard.

Merge decision: eligible for self-merge after required GitHub checks pass.

PR #2172, opened 2026-07-15
Add opt-in Reward Memory experiments

huangruiteng:
## Summary
- expose Reward Memory as a provider-neutral, default-off experimental goal capability
- gate configured ingest by registered-agent allowlist while retaining legacy no-write fixtures
- project the scoped capability through quota/catalog and document the OpenViking provider boundary in English and Chinese

## Validation
- `pytest -q`: 424 passed, 2 skipped
- Reward Memory focused smokes and evaluation gate passed
- `loopx canary premerge --from-git-diff`: 17/17 checks passed, public-boundary scan clean, no manual holds

## Scope and privacy
- no provider credentials or local config contents are committed
- the registry stores only an ignored repo-relative pointer and agent allowlist
- automatic ingest and automatic recall remain disabled

## Residual risk
- first live activation still requires a project-local provider binding and provider doctor; that is intentionally performed after this change is installed


huangruiteng:
Independent maintainer review

Findings
- No remaining blockers. During review I found that a hand-edited registry could set `enabled=true` without the explicit experimental marker; commit `4cbc2d74` now fails closed unless both `enabled` and `experimental` are true, with a regression test.

Assumptions
- OpenViking remains only a project-local provider binding; this PR does not install, authenticate, or globally enable it.
- Automatic ingest and recall remain off. Real writes still require explicit `--execute`, a configured goal, an allowlisted registered agent, standing-policy guards, and exact readback.

Product / architecture judgment
- Motivation: make Reward Memory trialable without silently changing every LoopX goal or coupling the kernel to OpenViking.
- Solved: `configure-goal`, the catalog, quota boundary, status, and ingest now share one provider-neutral default-off contract; unlisted agents receive no capability route.
- User impact: a fixer can be activated independently, while disabled goals and existing installs retain current behavior.
- Main risk: capability routing and local config validity. The policy normalizer lives in the control plane, provider parsing remains in the capability layer, and the architecture import-boundary test passes.
- Design: the ignored binding pointer plus explicit agent allowlist is reusable and narrower than a global `openviking_enabled` switch.

Validation
- `pytest -q`: 425 passed, 2 skipped.
- Focused Reward Memory/configuration tests and smokes passed; Reward Memory evaluation gate reports `ready_for_bounded_dogfood`.
- `loopx canary premerge --from-git-diff`: 17/17 selected checks passed; 0 failures, 0 warnings, 0 manual holds; public/private boundary clean; `self_merge_allowed=true`.

Merge decision
- Approved for self-merge once the GitHub pytest check is green.

PR #2173, opened 2026-07-15
Fix LoopX Turn session recovery eligibility

huangruiteng:
## Summary
- preserve observed Codex sessions across bounded timeouts and generic resumable failures
- invalidate incompatible-version, rejected-output-contract, and missing-session records; ignore legacy eligibility records
- explain a host-neutral Trae/Codex adapter recipe and fail-closed session recovery

## Validation
- `76 passed` across the focused LoopX Turn pytest set
- ruff and Python compile checks passed
- `loopx check` public-boundary scan passed
- standard `loopx canary premerge --from-git-diff`: 13 selected, 0 failures, 0 manual holds

## Real-host evidence
- current bundled Codex CLI returned a typed result on a fresh minimal lineage
- terminal failed-start lineage no longer becomes eligible to resume
- repository-sized real-host runs remain explicitly bounded and fail closed with no state write or quota spend; long-duration qualification continues after merge

### DIFF
--- a/docs/reference/protocols/loopx-turn-v0.md
+++ b/docs/reference/protocols/loopx-turn-v0.md
@@ -78,6 +78,69 @@ Turn request/result objects and that CLI's prompt, session, and output model.
 Raw transcript text, process exit zero, and the host's own completion claim are
 never sufficient validation.
 
+### Five Questions For Any Agent CLI
+
+Before wiring Trae CLI, Codex CLI, or another host, answer these five questions:
+
+1. **How does it run unattended?** Choose an explicit non-interactive command
+   and workspace. If the CLI is interactive-only, it is not an
+   `isolated-headless` adapter yet.
+2. **How does it return one typed result?** Prefer a native output schema or a
+   dedicated result file. Do not scrape arbitrary conversation text as the
+   completion contract.
+3. **What is its resume handle?** Keep the opaque handle in local adapter
+   state, keyed by `(goal_id, agent_id, todo_id)`. Never put it in LoopX state
+   or public evidence.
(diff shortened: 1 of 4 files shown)


huangruiteng:
## Self-review

**Findings:** No blocking findings in the four-file diff.

**Open assumptions / residual risk:** The session-record schema intentionally advances from v0 to v1, so existing v0 records become ineligible once and the next governed Turn starts fresh. This is the fail-closed migration needed to discard records created by terminal startup failures. Current invalidating categories are limited to incompatible model/CLI version, rejected output schema, and missing session; generic non-zero and observed timeout remain resumable by design.

**Product / architecture judgment:** The change fixes a real recovery-contract bug rather than papering over one local run: session existence is no longer treated as unconditional resume authority. It also makes the same boundary understandable for Trae, Codex, and other CLIs through the host-neutral Decide -> Execute -> Validate -> Commit model. The main compatibility cost is the one-time v0 session reset; no goal, todo, quota, benchmark scoring, permission, or public-evidence semantics change.

**Validation:**
- focused Turn pytest set: 76 passed
- ruff, Python compile, and diff checks passed
- public/private scan passed; no raw prompts, trajectories, session ids, local paths, or private scenario details included
- standard premerge canary: 13 selected, 13 passed, 0 failures, 0 manual holds
- real current Codex CLI: fresh minimal lineage returned a schema-valid typed result in 35.61s
- repository-sized qualification attempts hit the explicit host timeout and correctly produced 0 state writes / 0 quota spend; long-duration resume qualification remains a post-merge product check and is not counted as passing evidence here

**Merge decision:** Self-merge is allowed by the owner authorization for Turn work and the standard canary gate. I will merge after the required GitHub pytest check passes.

PR #2174, opened 2026-07-15
Preserve Explore visual styles on config updates

huangruiteng:
## Summary
- treat an omitted --board-style as preserve-existing for an already configured visual role
- keep auto_flow as the first-configuration default and preserve renderer validation
- cover canonical and executive Docx/stage-token updates plus legacy renderer-only configs
- document patch-style update semantics

## Validation
- .venv/bin/pytest -q: 435 passed, 2 skipped
- .venv/bin/pytest -q tests/test_explore_presentation_views.py: 35 passed
- .venv/bin/python examples/explore-result-layer-smoke.py
- .venv/bin/ruff check on changed Python files
- python3 -m compileall on changed Python files
- git diff --check
- LoopX public/private scan passed; token hits are public fixture identifiers only

## Boundary
No OpenViking core code, private state, local paths, credentials, generated logs, or uv.lock are included.

### DIFF
--- a/docs/capabilities/explore/README.md
+++ b/docs/capabilities/explore/README.md
@@ -698,6 +698,12 @@ The renderer (`mermaid` or `stage_svg`) is an implementation detail derived
 from `board_style`. Existing local configs that only store `renderer=mermaid`
 remain readable as `auto_flow`.
 
+On first configuration, omitting `--board-style` defaults the new visual role
+to `auto_flow`. On later calls for the same role, omission preserves the stored
+style and its validated renderer. This makes Docx or Evidence Stage token
+maintenance a patch operation instead of an implicit style reset. Pass an
+explicit `--board-style` only when intentionally switching the layout.
+
 A material sync checkpoints `canonical_rows_semantic_digest`
 and `visual_semantic_digest` independently. If whiteboard publication fails
 after Base rows succeed, the next run retries only the visual sink instead of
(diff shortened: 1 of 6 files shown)


huangruiteng:
Independent review: no blocking findings.

- Motivation: Docx or Evidence Stage token maintenance currently behaves like a full style rewrite because the CLI converts omission into auto_flow.
- Solved: omission is now preserved as PATCH semantics for an existing role, while a new role still defaults to auto_flow and an explicit value still switches style.
- Operator impact: canonical and executive boards no longer silently lose their chosen first-class layout during token maintenance.
- Main risk: legacy renderer-only configs and style/renderer mismatch handling. The resolver remains authoritative; a legacy mermaid config normalizes to auto_flow and invalid mismatches still fail closed.
- Design: the update rule lives in explore_visual_styles rather than growing the already-large Lark sink module.

Validation:
- 435 pytest passed, 2 skipped
- focused presentation tests: 35 passed
- Explore result-layer CLI smoke passed
- ruff and Python compile checks passed
- premerge canary: 17/17 selected checks passed, 0 failures, 0 warnings, 0 manual holds
- GitHub pytest check passed

Initial canary caught the Lark sink file over its line budget; the helper was moved into the style module and the final line-budget canary passed without raising the budget.

Merge decision: self-merge authorized; final gate reports self_merge_allowed=true.

PR #2175, opened 2026-07-15
Fix reviewer notification queue reconciliation

huangruiteng:
## Problem

An already-covered GitHub reviewer could be replaced by a different Lark-mapped candidate, and merge-only persistence left stale unsent reviewer notifications queued after the live route changed.

## Solution

- Bind secondary notifications to the same already-covered GitHub reviewer; skip the sink when that identity cannot resolve.
- Reconcile the unsent queue authoritatively on execute, with explicit replace/cancel evidence.
- Preserve matching queued work after a pre-write provider failure while avoiding duplicate requeue after an unverified external write.
- Document the contract and cover routing, replacement, cancellation, and restart-safe retry behavior.

## Validation

- 435 passed, 2 skipped
- Focused reviewer request, notification sink, and capability guide smokes passed
- Ruff check and git diff check passed
- LoopX premerge canary: 17/17 passed, zero warnings or manual holds, self-merge allowed
- Public-boundary scan clean

## Product judgment

This is a provider-neutral PR-scoped queue contract. It fixes wrong-person notifications and stale retries without expanding the reviewer model or adding a second ledger. Verified receipts stay append-only; only unsent work is replaceable.

### DIFF
--- a/docs/capabilities/issue-fix/protocols/issue-fix-reviewer-notification-sinks-v0.md
+++ b/docs/capabilities/issue-fix/protocols/issue-fix-reviewer-notification-sinks-v0.md
@@ -124,7 +124,15 @@ back to that same row. A restart preserves a queued item; a later verified send
 removes the matching queue entry while retaining the stable receipt. Retry
 outside the window returns `already_queued` with the original queue receipt,
 so neither the provider nor local state changes. Retry therefore remains
-idempotent without a second ledger. Goal boundary/status
+idempotent without a second ledger. The live reviewer route is authoritative
+for unsent work: an existing GitHub-covered reviewer may be notified only as
+that same reviewer, never replaced by a different mapped sink identity. Each
+execute reconciliation atomically replaces the PR's unsent queue, cancelling
+stale targets when coverage changes, a review completes, or no current sink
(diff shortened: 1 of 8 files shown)


huangruiteng:
Independent maintainer review

Findings: none.

Product judgment:
- Motivation is concrete: current routing could notify the wrong Lark identity and retain stale unsent work after reviewer state changed.
- The patch keeps the reviewer model small: existing GitHub coverage is identity-bound across sinks, while newly selected requests retain the existing ranked fallback behavior.
- Queue reconciliation is provider-neutral and PR-scoped. Verified receipts remain append-only; only unsent receipts can be replaced or cancelled.
- Failure semantics are bounded: a provider failure before an external write preserves the matching queue entry, while post-write unverified delivery is not requeued and cannot create a blind duplicate.

Validation:
- 435 passed, 2 skipped
- focused reviewer request / notification sink / guide smokes passed
- ruff and diff checks passed
- premerge canary 17/17 passed, zero warnings, zero manual holds
- public-boundary scan clean

Residual risk: provider-specific remote verification remains adapter-owned; this change does not weaken that gate.

Merge decision: self-merge once the required GitHub check is green, under the standing maintainer authorization.

PR #2176, opened 2026-07-15
Fix registry-relative runtime paths from independent worktrees

huangruiteng:
## Summary

- anchor relative `common_runtime_root` and relative `--runtime-root` values to the project that owns the selected registry on status/quota/history read paths
- resolve relative run-index artifact availability against that same project root
- keep status/contract/cache/history consumers on one canonical runtime binding
- add an independent-worktree CLI regression covering both `history` artifact readback and runnable `quota should-run`
- document the registry-relative read-path contract

## Product / architecture

This fixes a control-plane identity seam rather than special-casing OpenViking or issue #3274. A registry is the authority for its runtime namespace, so changing the caller's cwd must not change which run index or artifacts status/quota trust. The new optional `registry_path` anchor preserves existing absolute-path behavior and keeps callers without registry context backward compatible.

## Validation

- `pytest -q`: 436 passed, 2 skipped
- focused registry-relative runtime and source-route tests: 6 passed
- Ruff on touched Python files: passed
- real OpenViking #3274 independent worktree: 742 runs discovered; latest JSON/Markdown read back as present
- risk-based `loopx canary premerge --from-git-diff --tier standard`: passed, 17/17 selected checks, no warnings or manual holds
- `git diff --check`: passed

## Boundary / exclusions

- no project runtime, active goal state, credentials, raw logs, or private material included
- no OpenViking product code changed
- residual risk is limited to status/quota/history callers that intentionally relied on a relative runtime path being cwd-relative while also supplying a registry path; the documented read-path contract now makes registry ownership authoritative

PR #2184, opened 2026-07-16
Fix issue-fix Explore source runtime resolution

huangruiteng:
## Summary

- anchor the Issue Fix Explore projection runtime to the registry that declared a relative `common_runtime_root`
- preserve the canonical Explore result log when sync runs from an independent cross-repository worktree
- add a regression test that changes CWD and proves the canonical source node remains projected

## Validation

- `pytest -q tests/control_plane/test_registry_relative_runtime_resolution.py tests/test_explore_source_history_reconcile.py` (4 passed)
- `python examples/issue-fix-explore-projection-smoke.py`
- `python examples/lark-explore-source-runtime-route-smoke.py`
- `ruff check` on changed files
- `loopx canary premerge --from-git-diff --tier standard` (5/5 selected catalog checks passed; no warnings or manual holds)
- real OpenViking dry-run: projection restored from 159 to the canonical 907 rows and changed from `source_projection_regression_blocked` to `unchanged`, with no external write

## Boundary

No private state, local paths, raw logs, credentials, docs, permissions, destructive cleanup, or external sink data are included. The change reuses the existing registry-relative path contract and does not introduce a new reconciliation state machine.

### DIFF
--- a/loopx/capabilities/issue_fix/explore_projection.py
+++ b/loopx/capabilities/issue_fix/explore_projection.py
@@ -641,7 +641,7 @@ def project_issue_fix_explore_graph(
         state_file=state_file,
     )
     registry = load_registry(registry_path)
-    runtime_root = resolve_runtime_root(registry)
+    runtime_root = resolve_runtime_root(registry, registry_path=registry_path)
     result_log = explore_result_log_path(runtime_root, goal_id)
     existing_events = load_explore_result_events(result_log, goal_id=goal_id)
     outcomes_packet = build_issue_fix_outcome_collection_from_domain_state(
--- a/tests/control_plane/test_registry_relative_runtime_resolution.py
+++ b/tests/control_plane/test_registry_relative_runtime_resolution.py
@@ -6,7 +6,17 @@ from pathlib import Path
 import subprocess
 import sys
 
+import pytest
+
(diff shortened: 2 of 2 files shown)


huangruiteng:
Independent self-review

1. Findings: none. The failure is one missing registry-path argument at an existing relative-runtime callsite; the regression test reproduces an independent CWD and verifies the canonical source node is retained.
2. Open questions: none. Existing `source-history-reconcile` correctly reports 907/907 coverage, so no history copy or remote deletion is needed.
3. Product/architecture judgment: the user problem is a false Graph shrinkage alarm during cross-repository refresh. This patch reuses the generic registry-relative runtime resolver added in #2176 instead of adding another reconciliation state or weakening the append-only guard. User impact is that the guard now evaluates the actual canonical source. Main risk is limited to choosing the declared runtime for relative paths; absolute runtime roots remain unchanged.
4. Validation: focused pytest 4 passed; Issue Fix Explore and source-route smokes passed; Ruff and py_compile passed; standard premerge canary passed all 5 selected checks with no warning or manual hold; real dry-run restored 907 canonical rows and performed zero external writes.
5. Merge decision: approved for self-merge once GitHub pytest succeeds. Public/private boundary is clean; no permission, destructive action, or evidence-policy change.

PR #2185, opened 2026-07-16
Fix static-site import on macOS system Python

huangruiteng:
## Summary
- replace the Python 3.11-only datetime.UTC convenience alias at the static-site import boundary with timezone.utc
- add a focused UTC timestamp regression
- record the reusable local-release runtime-drift repair pattern

## Validation
- /usr/bin/python3 (3.9.6) source CLI import and static-site presentation smoke: passed
- /usr/bin/python3 examples/install-local-smoke.py: passed
- pytest tests/test_static_site_presentation.py: 9 passed
- Ruff on changed Python files: passed
- LoopX standard premerge canary: 4/4 catalog checks plus direct checks passed

## Boundary
Only public product code, focused validation, and public self-repair guidance are included. No local state, credentials, raw logs, or generated artifacts are committed.

## Residual risk
This restores the verified default macOS CLI path; it does not broaden the packaged Python support declaration.


huangruiteng:
Self-review

1. Findings: no blocking issue found in the three-file diff. The timestamp remains UTC ISO-8601 with a trailing Z; only the Python 3.11-only alias is removed.
2. Assumption: this is a narrow compatibility repair for the verified source-based macOS wrapper path, not a change to the packaged Python support declaration.
3. Product/architecture: the failure was an eager CLI import regression introduced by the static-site module. Fixing the equivalent standard-library call at that seam restores every command without adding installer policy or another abstraction. The reusable failure mode is recorded in the self-repair catalog.
4. Validation: system Python 3.9 CLI import, static-site product smoke, isolated install-local smoke, 9 focused pytest cases, Ruff, py_compile, and standard premerge canary all passed. Canary executed 4/4 selected catalog checks with zero failures, warnings, or manual holds. Boundary scan found only existing synthetic test URLs/path fixtures; no private state or generated artifact is included.
5. Merge decision: approved for authorized self-merge after the GitHub pytest check completes successfully.

PR #2188, opened 2026-07-16
Fix autonomous replan for stalled blocked successors

huangruiteng:
## Summary

- record exact blocked-successor waits as no-spend frontier observations instead of silent repeated reads
- trigger a bounded autonomous replan after two unchanged observations of the same frontier, while preserving the six-poll threshold for ordinary monitors
- carry the stable frontier identity into replan ACKs so the same frontier is deduplicated and a changed frontier can wake again
- require the interaction contract to execute the observation, without allowing delivery or quota spend

## Validation

- `pytest -q tests/control_plane/test_goal_vision_blocked_successor.py` (10 passed)
- full `pytest -q` (462 passed, 2 skipped)
- `ruff check` on all changed Python files
- `examples/autonomous-replan-obligation-smoke.py`
- `examples/control_plane/quota-replan-decision-plane-smoke.py`
- `examples/control_plane/heartbeat-quota-flow-smoke.py`
- `loopx canary premerge --from-git-diff` (16/16 selected checks passed, no manual holds)

## Review focus

This changes agent-facing control-plane routing and is intentionally held for owner review. Please verify that exact blocked successors enter replan after two unchanged frontier observations, ACK dedupe is identity-scoped, and generic monitor-only lanes still require six identical polls. No credentials, private state, local paths, raw logs, or external production actions are included.

### DIFF
--- a/loopx/control_plane/goals/goal_frontier.py
+++ b/loopx/control_plane/goals/goal_frontier.py
@@ -10,6 +10,7 @@ from ..agents.agent_scope import (
 )
 from ..agents.runtime_model import peer_work_key, select_peer_for_work
 from ..work_items.autonomous_replan_ack import (
+    autonomous_replan_ack_matches_frontier,
     autonomous_replan_ack_matches_agent,
     latest_autonomous_replan_ack_for_projection,
 )
@@ -1573,6 +1574,10 @@ def build_goal_frontier_projection_context_from_status(
     if (
         autonomous_replan_is_required(replan_obligation)
         and autonomous_replan_ack_has_frontier_delta(effective_replan_ack)
+        and autonomous_replan_ack_matches_frontier(
+            effective_replan_ack,
+            replan_obligation,
+        )
         and not acceptance_gaps
     ):
         replan_obligation = None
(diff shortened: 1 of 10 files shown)


huangruiteng:
## Owner review map

The behavior is a four-step state machine:

1. `goal_vision_wait` recognizes only a real `exact_blocked_successor` wait and derives a stable identity from agent, selected todo, and `resume_when`.
2. The interaction contract requires one `quota monitor-poll --execute` observation per heartbeat. This write is no-spend and delivery remains forbidden.
3. Two consecutive observations with the same identity produce `blocked_successor_no_progress_repeat` and route before `agent_scope_wait` to `autonomous_replan_required`. The agent may promote a safe in-scope evidence-backed successor; if none exists, it records wait continuation without spend.
4. `refresh-state --autonomous-replan-recorded` automatically copies the observed identity into the ACK. Only a matching ACK suppresses that frontier; a changed todo/resume condition can wake again.

Protected invariants: generic monitor polling still requires six identical observations; user gates, missing vision checkpoints, closed-stage successor gaps, delivery permission, and quota spend rules are unchanged.

Validation: focused tests 10/10; full pytest 462 passed / 2 skipped; quota-replan and heartbeat flow smokes passed; diff-driven premerge canary selected 16 checks and passed all 16; GitHub `pytest` check passed. No failures or manual holds. This is sufficient because it covers the new transition, ACK dedupe, identity change, existing monitor threshold, and the wider decision/heartbeat state machines.

Residual review point: this intentionally changes `agent_scope_wait` from a silent no-op into a mandatory no-spend observation so repeated waiting becomes machine-visible.

PR #2192, opened 2026-07-16
Avoid false Explore regressions for retired parent edges

huangruiteng:
## Summary
- distinguish source-backed Explore rows from projection-derived `parent_*` display edges in the Lark source coverage guard
- allow a retired derived parent edge without weakening fail-closed checks for canonical nodes, explicit edges, or findings
- extend the source-runtime-route smoke with both the allowed derived-edge case and the existing canonical-result regression case

The append-only invariant belongs to canonical Explore source history. Parent edges are mutable display materializations of the latest node topology, so treating them as source history can strand otherwise safe delivery after a node is detached or reparented. This keeps the guard strict at the source boundary while leaving stale display-row reconciliation independent.

## Validation
- `python3 examples/lark-explore-source-runtime-route-smoke.py`
- `python3 -m py_compile loopx/presentation/sinks/lark/explore_source_guard.py examples/lark-explore-source-runtime-route-smoke.py`
- `git diff --check`
- `loopx canary premerge --from-git-diff --format json` (passed; standard tier, public-boundary scan clean, no manual holds)

## Boundary
No credentials, private state, local paths, raw trajectories, or generated logs are included.

### DIFF
--- a/examples/lark-explore-source-runtime-route-smoke.py
+++ b/examples/lark-explore-source-runtime-route-smoke.py
@@ -34,6 +34,9 @@ from loopx.presentation.sinks.lark.explore_results import (  # noqa: E402
 from loopx.presentation.sinks.lark.explore_singleflight import (  # noqa: E402
     explore_feishu_sync_singleflight,
 )
+from loopx.presentation.sinks.lark.explore_source_guard import (  # noqa: E402
+    source_projection_coverage_guard,
+)
 
 
 GOAL_ID = "public-route-fixture"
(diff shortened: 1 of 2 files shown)


huangruiteng:
Independent self-review: no blocking findings.

- Motivation: source-history delivery was stranded when a mutable, projection-derived parent edge retired.
- Solved: the guard now applies the append-only invariant only to source-backed rows while still failing closed for missing canonical nodes, explicit edges, and findings.
- Operator impact: material refresh can continue without treating display-topology churn as source loss; stale display-row cleanup remains independent.
- Main risk: prefix classification could be too broad. The scope is constrained to `goal:edges:parent_*`, the only ID family synthesized by the current parent-edge projector.
- Design: this keeps source invariants in the source guard rather than weakening connector readback or adding project-specific exceptions.

Validation:
- focused source-runtime-route smoke passed, including allowed retired-parent coverage and blocked missing-finding coverage
- changed Python files compiled
- standard `loopx canary premerge --from-git-diff` passed
- GitHub `pytest` passed (1m30s)
- no failures, skips, or manual holds

Merge decision: self-merge approved under standing owner authorization; narrow single-purpose runtime seam with focused durable smoke.

PR #2216, opened 2026-07-16
Retain alternating scheduler host failure pairs

### LINKED ISSUE #2213: Retain alternating scheduler host-update failure pairs
## Problem

The Codex App scheduler persists only one `host_update_failure` record. A normal lifecycle can alternate desired cadence targets against one unchanged host RRULE, for example:

1. due work resets the target to the active-work interval;
2. the host update times out, so that target/host pair is recorded;
3. after bounded work, monitor wait chooses a backoff interval;
4. that host update also times out and overwrites the first failure;
5. the next due turn retries the first known-failing pair.

The exact-pair suppression contract therefore fails across cadence phase changes even though each individual decision is correct.

## Expected contract

- retain a small bounded set of recent `(target_rrule, observed_host_rrule)` failures instead of one scalar record;
- suppress every retained exact pair without ACK or quota spend;
- preserve failure kind/count/timestamp per pair;
- clear or invalidate entries when the observed host changes, the host successfully matches a target, or an explicit scheduler reset policy requires a fresh attempt;
- cap and expire the set so scheduler state stays compact.

## Regression coverage

Add a public-safe replay that covers:

- alternating active-work and monitor-wait targets against one unchanged host RRULE;
- both failed pairs remaining suppressed on the next cycle;
- successful host update clearing the relevant stale failure state;
- observed-host change reopening a bounded attempt;
- cache cap/expiry behavior and compatibility with existing scalar scheduler state.

No benchmark task text, raw logs, credentials, or host-specific identifiers should enter the fixture.

huangruiteng:
## Summary
- replace the single scheduler host-failure slot with a bounded 4-pair compatibility cache
- expire pairs after 24 hours, invalidate old-host pairs on observation changes, and clear the cache after a successful ACK
- replay alternating active-work and monitor-wait failures against one unchanged host RRULE

## Validation
- `uv run pytest -q tests/control_plane` (208 passed)
- scheduler hint compaction, monitor contract, and interaction authority smokes
- `loopx canary premerge --from-git-diff` (18 selected checks, 0 failures, 0 manual holds)

Closes #2213


huangruiteng:
Pre-merge validation packet:

- Changed surfaces: scheduler persisted state, scheduler hint arbitration/backoff, scheduler failure writeback, ACK clearing, public quota docs, self-repair pattern, focused public-safe replay.
- Checks: 208 control-plane tests; scheduler state/ACK CLI smoke; compact hint, monitor cadence, interaction authority, and turn-envelope tests; risk-based premerge canary with 18 selected checks.
- Failures: none.
- Skips: no benchmark jobs or external host updates were launched; the replay uses only synthetic public-safe RRULEs and local state.
- Manual holds: none reported by canary. Runtime behavior still remains owner-reviewed under repository policy.
- Coverage rationale: the replay proves the real 3m/30m -> 15m/30m -> 3m/30m alternation, legacy scalar migration, 4-entry cap, 24-hour expiry, host-change reopening, and successful-ACK clearing, while existing integration smokes protect prior CLI and compact contracts.

PR #2222, opened 2026-07-16
fix(scheduler): settle ACK before backoff progression

huangruiteng:
## Summary

- treat scheduler ACK as settlement of the current host RRULE, not proof that another cadence interval elapsed
- advance unchanged monitor backoff only after one real applied interval, preventing the reproduced 15 -> 30 -> 15 oscillation
- extend focused unit/CLI convergence coverage and document the host contract

## Validation

- `uv run --extra test pytest -q tests/control_plane` (272 passed)
- scheduler state ACK, policy, and hint compaction smokes
- Ruff checks and Python compile checks
- `loopx --format json canary premerge --from-git-diff` (18/18 selected checks passed; public boundary clean; no manual holds)

## Boundary

- no project/private state, credentials, raw gate content, or production actions
- generated local `uv.lock` excluded

## Residual risk

- legacy/partial scheduler state without a trustworthy settlement clock retains historical immediate progression for compatibility; newly ACKed state uses elapsed-time convergence.

### DIFF
--- a/docs/heartbeat-automation-prompt.md
+++ b/docs/heartbeat-automation-prompt.md
@@ -507,7 +507,9 @@ no quota, and end the turn without a scheduler ACK. Otherwise call it only when
 RRULE update, run `loopx` with
 `scheduler_hint.codex_app.ack_hint.cli_args`; current payloads use
 `quota scheduler-ack-current` so LoopX re-reads the latest hint and owns the
-progression/reset state. Attempt the host update at most once per hint and
+progression/reset state. The ACK settles that RRULE; an immediate final guard
+may verify the same target but must not be treated as another elapsed poll.
+Attempt the host update at most once per hint and
 turn. If it fails or times out, do not retry or ACK; run
 `scheduler_hint.codex_app.failure_hint.cli_args` once to persist the failed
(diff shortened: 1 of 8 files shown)


huangruiteng:
Independent review: no blocking findings.

- Motivation: the scheduler treated a successful ACK as if a new host cadence had already elapsed, so a monitor horizon shrink could produce 15 -> 30 -> 15 RRULE churn.
- Solved: progression is now gated by the persisted applied RRULE plus settlement timestamp; immediate reconciliation verifies convergence, while a real elapsed interval still advances backoff. Focused tests reproduce both paths.
- Operator impact: successful host updates/ACKs stop generating same-turn successor updates and tight retry loops.
- Main risk: legacy/partial state lacks a trustworthy clock; compatibility intentionally retains historical progression there. Newly ACKed state follows the convergent contract.
- Design: this is a generic scheduler-state contract, not OpenViking-specific behavior; host truth and active-work behavior remain intact.

Validation:
- control-plane pytest: 272 passed
- scheduler state ACK, policy, and hint compaction smokes passed
- premerge canary: 18/18 selected checks passed
- public/private boundary clean; no failures, skips, or manual holds

Merge decision: self-merge allowed once the remote pytest check is green.

PR #2224, opened 2026-07-16
Mount canonical LoopX source for Turn cases

huangruiteng:
## Summary

- include `loopx-turn-agent-cli` in the existing case-runtime source mount/upload contract
- keep host-local runs on source upload rather than a Docker bind mount
- cover both legacy product mode and Turn in the source-contract smokes

A merged Turn rerun proved the runner source fingerprint was correct but case initialization still had `case_loopx_source_install_requested=false`; the container then fell back to the public installer and failed in `ensure_cli`. Turn is already in `LOOPX_CASE_RUNTIME_ROUTES`, so the source contract should use that canonical set.

## Validation

- `python3 examples/skillsbench-benchmark-run-smoke.py`
- Python compile and repository line-budget smoke
- `loopx canary premerge --from-git-diff --tier standard`: 7/7 checks passed; public boundary scan clean for both files

## Risk and merge hold

The owner has explicitly authorized self-merge for LoopX Turn and benchmark PRs. This does not change scoring, task semantics, submissions, permissions, or launch a benchmark job; it replaces an unreliable installer fallback with the already-defined canonical source upload for a case-runtime route.

### DIFF
--- a/examples/skillsbench-benchmark-run-smoke.py
+++ b/examples/skillsbench-benchmark-run-smoke.py
@@ -7135,34 +7135,38 @@ def test_loopx_source_mount_contract_uses_real_cli_source_not_local_installer() 
         source_dir = Path(tmp)
         (source_dir / "loopx").mkdir()
         (source_dir / "loopx" / "cli.py").write_text("# source fixture\n")
-        args = parse_args(
-            [
-                "--route",
-                "loopx-product-mode",
-                "--loopx-source-dir",
-                str(source_dir),
-            ]
-        )
-        contract = _loopx_source_mount_contract(args)
-        assert contract["requested"] is True
-        assert contract["ready"] is True
-        assert contract["status"] == "ready"
+        for route in ("loopx-product-mode", "loopx-turn-agent-cli"):
+            args = parse_args(
+                [
+                    "--route",
+                    route,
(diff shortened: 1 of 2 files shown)


huangruiteng:
## Self-review

**Findings:** no blocking findings.

**Assumption:** the owner’s standing authorization for LoopX Turn and benchmark PRs satisfies the generic `benchmark_sensitive` hold. GitHub Python Tests are not expected for this PR because the workflow path filter excludes `scripts/**` and `examples/**`.

**Product/architecture judgment:** Turn already belongs to the canonical `LOOPX_CASE_RUNTIME_ROUTES`, but source provisioning used the narrower legacy product-mode family. Reusing the case-runtime set removes that taxonomy drift and makes the merged runner source available inside the case before lifecycle initialization. Host-local runs still use the existing bounded source upload path rather than a bind mount.

**Validation:** full SkillsBench benchmark-run smoke passed, including both product-mode and Turn source-contract cases; Python compile, line-budget, and diff checks passed; risk-based premerge ran 7/7 checks; public/private scan was clean for both changed files.

**Merge decision:** self-merge with admin bypass under explicit owner authorization. No score, task, submission, permission, or public evidence contract changes.

PR #2227, opened 2026-07-16
fix(state): avoid technical gate false positives

huangruiteng:
## Summary
- stop treating the bare word `gate` after `require` as evidence of a user/owner wait
- preserve explicit actor, approval, permission, and decision wait detection
- cover technical benchmark acceptance text that verifies a next-case gate remains open

## Root cause
The user-wait regex accepted `require ... gate` across a 120-character window. A normal benchmark acceptance statement (`require both arms ... leave the next-case gate open`) was therefore projected as an unrepresented user gate even though the matching Agent Todo already existed.

## Validation
- real active-state readback now returns no `state_projection_gap`
- focused state-projection and quota-plan smokes passed
- Full Public Smokes: 593/593 passed, no warnings, skips, failures, or timeouts
- post-rebase premerge canary: 6/6 selected checks passed; no manual holds; public/private boundary clean

### DIFF
--- a/examples/state-projection-gap-smoke.py
+++ b/examples/state-projection-gap-smoke.py
@@ -114,6 +114,15 @@ def test_projection_gap_warning() -> dict:
     )
     assert technical_gate_no_gap is None, technical_gate_no_gap
 
+    benchmark_acceptance_no_gap = state_projection_gap_warning(
+        "## Agent Todo\n\n"
+        "- [ ] Run the second matched LoopX Turn versus Codex Goal pair.\n\n"
+        "## Next Action\n\n"
+        "- Run the second matched pair; require both arms to be official and "
+        "verify committed receipts leave the next-case gate open.\n"
+    )
+    assert benchmark_acceptance_no_gap is None, benchmark_acceptance_no_gap
+
     decision_result_input_no_gap = state_projection_gap_warning(
         "## Agent Todo\n\n"
         "- [ ] Build a read-only projection adapter from compact session facts.\n\n"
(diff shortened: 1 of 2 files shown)


huangruiteng:
Maintainer review and pre-merge evidence

**Findings:** none. The one-token classifier narrowing removes the observed false positive while retaining explicit user/owner, approval, permission, decision, and choice waits.

**Product/architecture judgment:** `state_projection_gap` should represent missing control-plane ownership, not technical prose that happens to mention a gate. The current broad regex made healthy agent work look user-blocked and could preempt normal delivery with unnecessary self-repair. Narrowing the actor/approval predicate is the lightest reusable fix; matching todo text or adding a benchmark-specific exception would be more brittle. Installed-user risk is low and limited to fewer false-positive warnings; true owner/approval waits remain covered by existing fixtures.

**Changed surfaces:** `loopx/state_projection.py` user-wait classification and one durable regression fixture.

**Checks:** focused state-projection smoke, quota-plan smoke, real active-state readback, changed-Python compile, diff checks, 5 catalog canaries, and public/private boundary all passed. Full Public Smokes passed 593/593 with zero warnings, skips, failures, or timeouts.

**Failures/skips:** none. **Manual holds:** none. Premerge gate passed with `self_merge_allowed=true`.

**Coverage rationale:** the fixture reproduces the exact false-positive shape, existing user-wait cases guard the retained positive behavior, quota/status canaries cover the projection consumers, and the full public sweep rules out broad classifier regressions.

**Merge decision:** eligible for maintainer self-merge.

PR #2229, opened 2026-07-16
fix(configure): sync authoritative shared runtime

huangruiteng:
## Summary
- resolve implicit configure-goal sync through the registry-declared authoritative shared runtime
- preserve explicit --runtime-root precedence and expose the selected target plus exact digest readback
- cover CLI and dashboard apply paths with a split-runtime regression and README guidance

## Validation
- LoopX premerge canary: 18/18 selected checks passed
- public boundary scan: clean across 8 changed files
- focused configure-goal, API, planner, and peer-runtime smokes passed

## Boundary
- public repository code, tests, and docs only
- no credentials, private project state, or production actions

### DIFF
--- a/README.md
+++ b/README.md
@@ -422,6 +422,13 @@ templates. Use `loopx configure-goal --help` for the complete settings surface.
 Preview changes without `--execute`; do not enable a feature only because it is
 available.
 
+An applied `configure-goal` change also refreshes the registry-declared shared
+read model. Without `--runtime-root`, LoopX resolves the authoritative shared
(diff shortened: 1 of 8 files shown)


huangruiteng:
Independent review: no blocking findings. The implicit path resolves from the registry-declared source_registry route, explicit --runtime-root remains authoritative, source mutation is preflighted against target writability, and delivery fails closed on exact digest/source/synced_at readback mismatch.

Validation: local premerge canary passed 18/18 with zero warnings; GitHub pytest passed.

PR #2230, opened 2026-07-16
Persist owner-only SkillsBench runner profiles

huangruiteng:
## Summary
- add an owner-only, fail-closed local runner profile for durable SkillsBench source/bridge/validator configuration
- let the existing launcher load the profile while preserving explicit environment overrides
- add a public-safe smoke covering permissions, schema boundaries, precedence, redaction, and launcher integration

## Product judgment
This repairs a real resume gap: runner setup previously lived only in one process environment, so a later heartbeat could not recover it. The partner-facing Turn contract remains unchanged; this is an operator-only launcher input, not a new Turn protocol layer.

## Validation
- `python3 -m py_compile loopx/benchmark_adapters/skillsbench_runner_profile.py examples/skillsbench-runner-profile-smoke.py`
- `bash -n scripts/skillsbench-launch-goal-xhigh.sh`
- `PYTHONPATH=. python3 examples/skillsbench-runner-profile-smoke.py`
- `PYTHONPATH=. python3 examples/skillsbench-benchmark-run-smoke.py`
- `PYTHONPATH=. python3 examples/skillsbench-loopx-turn-agent-cli-smoke.py`
- `PYTHONPATH=. python3 examples/skillsbench-runner-source-fingerprint-smoke.py`
- `loopx canary premerge --from-git-diff --tier standard` (all 7 executed checks passed; benchmark-sensitive manual review hold remains)
- public/private boundary scan clean

No benchmark job, scoring, task semantics, submission, or permission boundary changed. `shellcheck` is unavailable locally.

PR #2232, opened 2026-07-16
Discover the owner-local SkillsBench runner profile

huangruiteng:
## Summary
- give the owner-only SkillsBench runner profile a stable XDG-state default location
- let the existing launcher discover that profile when no explicit path is supplied
- keep explicit environment and explicit profile overrides intact, and preserve fail-closed symlink/permission checks

## Product judgment
PR #2230 persisted profile values but left the profile locator in process-local environment, so a later heartbeat could still lose the runner. This follow-up closes that exact resume gap without adding Turn protocol fields or automatic secret capture. A trusted runner session explicitly runs `capture` once; subsequent launcher calls discover the owner-local profile.

## Validation
- Python compile
- launcher `bash -n`
- runner-profile smoke, including default discovery and redaction
- full SkillsBench runner smoke, including no-profile fallback
- standard LoopX premerge canary: all direct and 7 selected checks passed

No benchmark job, scoring, task semantics, submission, permission boundary, or automatic credential persistence changed. `shellcheck` remains unavailable locally.

### DIFF
--- a/examples/skillsbench-runner-profile-smoke.py
+++ b/examples/skillsbench-runner-profile-smoke.py
@@ -13,6 +13,7 @@ from loopx.benchmark_adapters.skillsbench_runner_profile import (
     SKILLSBENCH_RUNNER_PROFILE_SCHEMA_VERSION,
     SkillsBenchRunnerProfileError,
     capture_skillsbench_runner_profile,
+    default_skillsbench_runner_profile_path,
     load_skillsbench_runner_profile,
     skillsbench_runner_profile_shell_exports,
     skillsbench_runner_profile_summary,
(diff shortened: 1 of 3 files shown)


huangruiteng:
## Maintainer self-review

**Findings:** no blocking code findings.

**Assumption:** profile capture remains an explicit action in a trusted runner environment; discovery must never scrape another process, SSH config, raw logs, or private bridge contents.

**Product/architecture judgment:** #2230 stored the capability but not its locator. This 66-line follow-up makes the locator deterministic under owner-local XDG state while keeping the Turn mental model unchanged. It is smaller and safer than recording an absolute profile path in LoopX state or the automation prompt.

**Changed surfaces:** profile CLI default resolution, existing launcher discovery, focused smoke coverage.

**Validation:** compile, `bash -n`, runner-profile smoke, full SkillsBench runner smoke, standard premerge canary, and public/private boundary scan. All executed checks passed. Missing-profile fallback and profile-value redaction remain covered.

**Failures/skips:** no final validation failures; `shellcheck` unavailable; no benchmark job launched. The unrelated OpenViking duplicate-index warning remains outside this diff.

**Manual hold:** canary flags benchmark-sensitive paths. Under the owner’s explicit self-merge authorization for benchmark/Turn PRs, this is a narrow runner-resume repair with no scoring, task, submission, permission, or launch change.

**Merge decision:** self-merge after required CI passes.

PR #2238, opened 2026-07-16
Fix Codex IDE goal-mode activation

huangruiteng:
## Summary

- add `codex-ide` as a canonical LoopX host surface and route it through the same visible `/goal <task_body>` adapter as Codex CLI TUI
- stop silently treating a missing `start-goal --host-surface` as Codex App; return a read-only exact-host selection gate instead
- upgrade the managed `/loopx` skill to pass the current host explicitly and continue through one bounded validated delivery segment
- extend the deterministic onboarding oracle and one-arm Doubao contract so an IDE packet routed to App fails before any provider call

Official Codex IDE documentation lists `/goal` as the persistent goal command: https://learn.chatgpt.com/docs/developer-commands?surface=ide

## Why

The v0.2.4 and current release paths both defaulted `start-goal` to `codex-app`, while the host catalog represented Codex App and Codex CLI but not Codex IDE. An IDE invocation could therefore receive App heartbeat activation instead of its visible goal-mode contract. This is a latent host-routing mismatch rather than a command removed after v0.2.4.

## Validation

- `pytest -q`: 585 passed, 2 skipped
- post-rebase focused suite: 33 passed
- `ruff check` on all changed Python files: passed
- `loopx canary premerge --from-git-diff`: 16/16 selected checks passed, no holds
- manual CLI projection: explicit `codex-ide` returns `codex_ide_visible_goal_mode`, `set_visible_goal`, and `/goal <task_body>`
- manual omission projection: returns a read-only five-choice host gate with no registry/state/quota writes
- live Doubao one-arm entry qualification: 2/2 repeated candidate calls selected `connect_if_needed`; semantic contract complete, source aligned, zero safety violations, zero tool calls, and no raw packet/response persistence

## Review focus

- confirm changing omitted host from implicit App to a read-only selection gate is the desired fail-closed compatibility behavior
- confirm IDE and CLI should share `set_visible_goal` while keeping distinct activation surface ids
- confirm the regular model qualification only checks the shipped one-arm packet; full-detail recovery remains deterministic-only

## Public/private boundary

No chat transcript, private screenshot, credential, local path, raw model response, or generated run artifact is included. The live key was injected at runtime from an existing ignored local environment and was not copied into this repository or any receipt.


huangruiteng:
Live one-arm model qualification receipt summary:

- surface: candidate default `start-goal --guided` entry projection for `codex-ide`
- repetitions: 2
- decision: `connect_if_needed` in both runs
- source alignment: 2/2
- semantic contract completeness: 2/2
- safety violations: 0
- tool calls: 0
- raw packet persisted: false
- raw model response persisted: false

The deterministic independent oracle separately rejects `requested_host_surface=codex-ide` when activation is routed to Codex App, before invoking the provider. Credentials were runtime-injected from an ignored local environment and were not copied or printed.

PR #2243, opened 2026-07-16
fix(quota): preserve delivery attribution across registry checkouts

huangruiteng:
## Summary
- let accountable refresh-state identify the actual delivery worktree when registry projection runs elsewhere
- persist only credential-free repository/worktree identity and reject canonical peer overrides
- teach the agent prompt and quota/status contracts about the causal override

## Validation
- quota-spend-workspace-causality-smoke: passed
- tests/control_plane/test_quota_slot_accounting.py: 11 passed
- ruff check on changed Python: passed
- LoopX premerge canary: 18/18 passed, public-boundary clean, self-merge allowed

## Commit grouping
1. runtime/CLI and agent-facing guidance
2. durable regression smoke and public contracts

## Excluded
- generated local uv.lock remains untracked
- no private state, credentials, raw trajectories, local paths, or production actions

## Residual risk
Callers that project from a different checkout must opt into --delivery-workspace-path; omission intentionally keeps the existing fail-closed canonical attribution.

### DIFF
--- a/docs/quota-allocation.md
+++ b/docs/quota-allocation.md
@@ -937,6 +937,11 @@ Post-turn accounting protocol:
   account the latest unspent `outcome_progress` delivery run once; a later
   duplicate spend is rejected because the latest run is then the spend event,
   not the delivery run.
+- keep accountable delivery attribution on the worktree that produced it. If
+  `refresh-state` must run from a separate registry checkout, pass
+  `--delivery-workspace-path <delivery-worktree>`; the path is validated locally
+  and omitted from persisted history. Do not point this option at the canonical
+  checkout for peer work.
 - autonomous replans follow the same accountable-outcome rule: spend after a
   concrete successor, blocker, or `outcome_progress`/`primary_goal_outcome`
   writeback, but do not spend for a `surface_only` watch-lane continuation or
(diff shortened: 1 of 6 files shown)


huangruiteng:
## Review

### Findings
No blocking findings.

### Open assumptions
- Cross-checkout callers must explicitly pass `--delivery-workspace-path`; omission intentionally keeps current fail-closed attribution.
- The option proves the same credential-free worktree class/repository identity available when running from that worktree; it does not add new write authority.

- **Motivation:** accountable state projection could overwrite an independently produced delivery with the canonical registry checkout and make quota spend impossible.
- **Solved:** the CLI now carries an explicit causal worktree into the existing `delivery_workspace_v0` snapshot while omitting the local path; the smoke proves canonical projection + independent spend and rejects a canonical peer override.
- **Operator impact:** split-runtime projects can write back through their registry checkout without losing the delivery repository/worktree lineage.
- **Main risk:** callers need one explicit argument when projection and delivery checkouts differ. This is preferable to guessing from prior runs and accidentally blessing unrelated work.
- **Design:** the change extends the existing workspace attribution contract and guard; it does not add a parallel provenance system or weaken repository isolation.

### Validation
- focused causality smoke: passed
- quota slot accounting: 11 passed
- ruff changed Python: passed
- premerge canary: 18/18 passed; public-boundary clean; no manual holds
- GitHub pytest: passed
- generated local `uv.lock` excluded

### Merge decision
Self-merge authorized by the owner and allowed by the validated premerge gate.

PR #2248, opened 2026-07-16
fix(scheduler): retain failed pairs across fallback ack

huangruiteng:
## Summary
- retain unexpired failed target/host pairs when a fallback RRULE is acknowledged on the same host
- clear only the acknowledged target pair or failures invalidated by a genuine host observation change
- document and test fail(3m,15m) -> ack(15m) -> active(3m) suppression

## Validation
- 70 focused scheduler/control-plane tests passed
- quota-scheduler-state-ack-smoke passed
- Ruff check and py_compile passed
- LoopX premerge canary passed 18/18 with no failures, skips, warnings, or manual holds

## Excluded
- generated local uv.lock remains untracked and is not part of this PR

## Risk
Narrow scheduler-state transition only; no schema, CLI, permission, or production action changes.

### DIFF
--- a/docs/quota-allocation.md
+++ b/docs/quota-allocation.md
@@ -845,7 +845,10 @@ so active-work and monitor-wait targets cannot overwrite one another while the
 host RRULE remains unchanged. Later heartbeats expose `apply_needed=false` and
 `state_status=host_update_failure_suppressed` for every retained exact pair.
 A changed host observation invalidates failures recorded against the old host,
-and a successful scheduler ACK clears the cache. The legacy
+and a successful scheduler ACK clears only failures whose target is the
+acknowledged RRULE. A fallback ACK therefore preserves failures for other
+targets observed against the same unchanged host, preventing the next turn
+from retrying a known-bad target/host pair. The legacy
 `host_update_failure` scalar remains as the latest compatibility projection;
 new hosts should consume `host_update_failures`. LoopX never treats an intended
 cadence as an applied cadence.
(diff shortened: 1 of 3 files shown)


huangruiteng:
## Review

### Findings
No blocking findings.

- Assumes a successful fallback ACK is authoritative host observation for that RRULE, while failures for other targets on the same host remain useful until TTL expiry, host change, or exact-target ACK. This matches the existing pair-cache contract.

- Motivation: stop a fallback ACK from erasing a different known-bad target/host pair and causing the next active turn to repeat the same failed host update.
- Solved: the ACK transition now retains only unexpired, same-host, non-acknowledged-target failures; exact-target ACK and genuine host change still clear the relevant records.
- Operator impact: fail(3m,15m) -> ack(15m) -> active(3m) stays suppressed with no repeated host call.
- Main risk: scheduler-state retention could over-suppress after host convergence. Existing exact-target ACK and host-change tests, plus the new fallback regression, cover both invalidation paths.
- Design: reuses the existing bounded/TTL pair cache and state schema; no new compatibility layer, CLI option, or special-case project policy.

### Validation
- 70 focused scheduler/control-plane tests passed.
- quota-scheduler-state-ack-smoke passed.
- Ruff check, py_compile, and git diff checks passed.
- Premerge canary passed 18/18; no failures, skips, warnings, or manual holds.
- Public/private boundary scan clean; local generated uv.lock excluded.

### Merge decision
Validated and self-mergeable under the recorded owner authority; waiting only for GitHub pytest to finish.

PR #2279, opened 2026-07-17
Fix Turn action-signature coverage migration

huangruiteng:
## Summary

- preserve `turn_envelope_action_dimensions_v0` for envelopes without a response plan
- advertise `turn_envelope_action_dimensions_v1` when the blocking user-gate `response_plan` is signed
- carry coverage through CLI differential receipts, allow only the declared v0-to-v1 migration, and keep unversioned digest drift fail-closed
- add a public-safe blocking-gate replay so the real base/head differential exercises the migration

This is the narrow follow-up to the unresolved compatibility finding on #2274.

## Validation

- `34 passed` across Turn envelope, user-gate lane, and CLI differential tests
- `cli-output-base-head-differential-smoke`: 87 base/candidate rows, exactly one expected coverage migration review signal
- `cli-output-budget-regression-smoke`: passed
- `loopx canary premerge --from-git-diff`: passed 18/18 selected checks, 0 failures, 0 manual holds
- Python compile, `git diff --check`, and public/private boundary scan passed

## Boundary

No private state, raw benchmark evidence, credentials, local paths, or generated logs are included. Future coverage versions remain fail-closed until explicitly reviewed.


huangruiteng:
Pre-merge review complete.

Changed surfaces:
- TurnEnvelope action-signature coverage selection and legacy constant compatibility
- CLI semantic receipt and base/head migration arbitration
- public-safe blocking user-gate replay plus focused matrix/unit regressions
- Turn protocol and self-repair guidance

Validation:
- focused pytest: 34 passed
- real base/head CLI differential: 87/87 rows, one expected v0-to-v1 review signal on the blocking-gate row
- GitHub Python Tests: passed
- final `loopx canary premerge --from-git-diff`: passed 18/18 selected checks, including 9 catalog canaries, 8 risk-profile smokes, Python compile, diff hygiene, and public-boundary scan
- failures/skips/manual holds: no final failures, no skips, 0 manual holds

One earlier canary attempt timed out the CLI budget smoke at 120s while unrelated checks were also several times slower. The same final commit passed that smoke directly in 54s and passed the complete canary rerun in 60s, so the timeout was treated as host-load evidence rather than hidden.

Coverage is sufficient because the real replay now exercises the exact blocking state missed by #2274, unchanged envelopes assert v0, blocking envelopes assert v1, and same-version or unknown-version digest drift remains fail-closed. No private state, raw benchmark evidence, credentials, local paths, or generated logs are present.

PR #2280, opened 2026-07-17
fix(scheduler): require explicit runtime ownership

huangruiteng:
## Summary

- remove the library-level Codex App compatibility default so missing scheduler context fails closed
- preserve Codex App cadence through the explicit `codex_app_heartbeat` runtime profile while binding Codex CLI, Claude Code, visible multi-agent, benchmark, and controller paths to their typed host/owner/mode
- keep the heartbeat light: generated App quota guards use `--codex-app`; generated ACK/failure follow-ups use the compact explicit `-A` binding instead of adding prose
- carry the same scheduler context through `scheduler-ack`, `scheduler-ack-current`, and `scheduler-fail-current` recomputation; omit only redundant default surface/state CLI arguments
- migrate durable scheduler oracles and runtime smokes away from implicit App behavior; record the follow-up binding gap as a self-repair pattern

## Runtime closure

A real Codex App heartbeat cadence update exposed a gap after the first implementation: the emitted ACK command recomputed `quota should-run` without the App profile and rejected the writeback. The repair now executes the generated ACK command end to end. On the registered `loopx-meta` runtime, host cadence update plus generated ACK succeeded; immediate readback reported `apply_needed=false`, `ack_needed=false`, `state_status=same_identity`, and `host_action=none`. The cadence transition and ACK consumed no quota spend.

## Validation

- `.venv/bin/python -m pytest -q`: 659 passed, 2 skipped
- CI lint scope: `ruff check tests loopx/canary loopx/control_plane loopx/domain_packs loopx/presentation`
- `.venv/bin/mypy`: success on the configured 12-file strict scope
- scheduler state ACK, registry route, and compact hint smokes execute the generated App-bound follow-up command
- agent-facing CLI base/head differential and output-budget regression smokes passed without raising budgets
- `loopx canary premerge --from-git-diff`: 9/9 catalog canaries, 8/8 risk-profile smokes, Python compile, diff checks, line budget, install/update checks, and public-boundary scan passed
- no private state, local paths, credentials, raw trajectories, generated logs, or benchmark evidence are included

## Review hold

Independent maintainer review is required. The canary result is `manual_review_required` because the migration touches benchmark command-generation paths; this PR must not self-merge. No scoring, task semantics, or launch behavior changes are intended.

PR #2212, opened 2026-07-16
Add provider-aware capability and extension registry

cocolord:
## Summary

- keep capability contracts distinct from extension delivery lifecycle
- compose the existing built-in catalog with explicitly enabled TOML extension manifests
- expose origin, visibility, and provider identity while preserving existing capability order and v0 CLI fields
- fail closed on duplicate ids, incompatible extension API ranges, malformed manifests, and internal-only public reads
- document the boundary without moving existing capability directories or adding an installer

## Validation

- `510 passed, 2 skipped`
- `python3 examples/capability-extension-registry-smoke.py`
- changed-file Ruff check and format check
- wheel build/install readback includes `loopx.capabilities.registry` and `loopx.extensions.manifest`
- `loopx canary premerge --from-git-diff --tier standard`: 13/13 selected checks passed, 0 warnings
- public/private boundary scan: 9 changed files clean

## Holds

- No manual holds from premerge canary.
- Runtime/API change: requesting independent review; no self-merge.

PR #2292, opened 2026-07-18
fix(scheduler): scope failure cache during backoff

huangruiteng:
## Summary

- let an acknowledged human-gate cadence advance after its real interval even when the cache retains failures for other targets
- keep an unacknowledged current target pinned, and preserve exact target/observed-host retry suppression
- cover host-match fallback ACK retention and the persisted CLI 30-to-60 progression

## State-model rationale

A historical failure is evidence about one target/observed-host pair. It must block progression only while the cadence at the persisted progression index is still unsettled. A successful ACK of that cadence is stronger, newer evidence; unrelated cached failures remain available for their own exact-pair circuit breaker but no longer pin the index.

## Validation

- `713 passed, 2 skipped` (`.venv/bin/pytest -q`)
- `9 passed` across user-gate progression, host-failure cache, and scheduler convergence tests
- `quota-scheduler-state-ack-smoke ok`
- Ruff and `loopx check` clean on all three changed paths
- `loopx canary premerge --from-git-diff`: direct checks, compile, public boundary, and 14 selected smokes passed; three selected smokes failed on both this branch and clean current `origin/main`:
  - `interaction-contract-state-machine-smoke.py`: fixture lacks the typed scheduler execution context required since #2280
  - `heartbeat-quota-flow-smoke.py`: same missing execution-context baseline
  - `premerge-validation-gate-smoke.py`: current main imports `tomllib` through the installed-wrapper Python 3.9 probe after #2212

Because the risk gate is not green, this runtime PR is intentionally not self-merged.

### DIFF
--- a/examples/control_plane/quota-scheduler-state-ack-smoke.py
+++ b/examples/control_plane/quota-scheduler-state-ack-smoke.py
@@ -34,6 +34,7 @@ from loopx.control_plane.scheduler.scheduler_hint import (  # noqa: E402
     build_scheduler_hint as _build_scheduler_hint,
 )
 from loopx.control_plane.scheduler.state import (  # noqa: E402
+    SCHEDULER_HOST_UPDATE_FAILURE_SCHEMA_VERSION,
     SCHEDULER_STATE_SCHEMA_VERSION,
     load_scheduler_state,
     write_scheduler_state,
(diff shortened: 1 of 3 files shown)


huangruiteng:
Validation packet

- Changed surfaces: scheduler progression arbitration, human-gate unit regression, persisted scheduler ACK smoke.
- Focused checks: 9 scheduler tests passed; `quota-scheduler-state-ack-smoke ok`; Ruff passed.
- Broad checks: `713 passed, 2 skipped`; `loopx check` passed with 0 errors/warnings.
- Premerge gate: direct diff/compile/public-boundary checks passed. Three selected smokes failed and were reproduced on a clean worktree at current `origin/main` (`f4cc1fa1`): two stale fixtures omit the required scheduler execution context; the installed-wrapper probe reaches Python 3.9 and cannot import `tomllib` after #2212.
- Manual hold: no self-merge while the risk gate is red. The changed-path regressions prove both the positive path (fallback ACK plus elapsed 30-to-60 progression) and the negative invariant (an unacknowledged failed current target remains pinned).

PR #2302, opened 2026-07-18
fix: record reward memory execution evidence

huangruiteng:
## Summary
- distinguish business recall from ingest-verification queries and retain digest-only query evidence
- attach conservative Issue Fix execution evidence with provider-call count, exact readback status, application receipt, and explicit unknowns
- exercise the real OpenViking provider adapter through version/status/search/read in the reward-memory smoke

## Validation
- 32 reward-memory capability tests passed
- reward-memory-recall-application-smoke
- reward-memory-dogfood-smoke
- issue-fix-repository-memory-smoke
- issue-fix-reviewer-notification-sink-smoke
- ruff check and format check
- git diff --check

## Safety
- no exact provider query or provider content is persisted in public evidence
- applied claims require exact readback plus a verified application receipt
- current Open PR file sets were checked and do not overlap

### DIFF
--- a/examples/reward-memory-recall-application-smoke.py
+++ b/examples/reward-memory-recall-application-smoke.py
@@ -22,6 +22,7 @@ from loopx.capabilities.context_providers.openviking import (  # noqa: E402
     OpenVikingContextProvider,
 )
 from loopx.capabilities.issue_fix.reward_memory import (  # noqa: E402
+    _execution_evidence,
     run_issue_fix_patch_planning_reward_memory,
     run_issue_fix_reviewer_artifact_automatic_reward_memory,
     run_issue_fix_reviewer_artifact_reward_memory,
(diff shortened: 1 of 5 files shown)


huangruiteng:
## Self-review

### Findings

No remaining findings on head `3d296ce5`.

During self-review I found one fail-closed gap: an old hand-built session could carry an application receipt while omitting business-query evidence and a positive provider call count. Commit `3d296ce5` fixes that before merge; `memory_applied` now requires matching query evidence, `provider_call_count > 0`, exact readback, and a matching verified receipt.

None. The new evidence fields remain additive under the existing v0 contracts, and incomplete legacy evidence degrades to explicit unknowns instead of an applied claim.

- **Motivation:** separate ingest verification from business recall and stop process proof from being reported as a verified application outcome.
- **Solved:** the shared recall/receipt seam now records digest-only query identity and provider-call/readback facts; the Issue Fix adapter derives conservative result and unknown fields from those receipts.
- **User/operator impact:** reports can distinguish provider not called, called without verified recall, recalled but not applied, and applied with a verified receipt.
- **Main risk:** public evidence compatibility. The change is additive and fail-closed; exact query/provider content remains unexposed, and old incomplete sessions cannot produce an applied claim.
- **Design judgment:** the invariant lives in the shared Reward Memory request/receipt contract with a thin Issue Fix outcome adapter, rather than a one-off reporting patch.

### Validation

- GitHub `pytest`: passed on head `3d296ce5`.
- 32 focused reward-memory tests: passed.
- 4 relevant smoke paths: passed, including the real `OpenVikingContextProvider` version/status/search/read adapter path.
- `loopx canary premerge --from-git-diff --tier standard`: passed; 4 direct checks + 9 selected checks, 0 failures, 0 warnings, 0 manual holds.
- Public/private boundary: clean across all 5 changed files.
- Diff hygiene and changed-file compilation: passed.
- Failures/skips: none in the final validation set.

### Merge decision

Owner-authorized self-merge approved. The diff is single-purpose, canary-gated, public-safe, and the evidence-policy change is conservative/fail-closed; admin-bypass merge follows.

PR #2306, opened 2026-07-18
Fix mixed authority registry projections

huangruiteng:
## Summary
- prefer canonical authority maps when rich source data and legacy compact fields coexist
- refresh co-located compact summary fields after authority registration/import
- recompute material/topic counts and path existence in registration and project-map projections

## Validation
- 750 pytest passed, 2 skipped
- register-authority-source, project-map, import-doc-registry-authority, platform-migration-material-registry, and review-packet CLI smokes passed
- LoopX premerge canary passed: 16 selected checks, 0 failures, 0 manual holds
- ruff check and git diff checks passed

## Boundary
- only public runtime code and durable focused smokes are included
- no local LoopX state, internal material, raw logs, credentials, or generated lockfiles are included

### DIFF
--- a/examples/project/project-map-smoke.py
+++ b/examples/project/project-map-smoke.py
@@ -55,8 +55,16 @@ def write_planned_registry(root: Path) -> Path:
                         "state_file": state_file,
                         "authority_sources": [{"kind": "doc", "role": "primary", "path": "README.md"}],
                         "authority_registry": {
+                            "declared": True,
+                            "required": True,
                             "path": "docs/meta/DOC_REGISTRY.yaml",
+                            "path_exists": False,
                             "read_status": "read",
+                            "default_entry_count": 99,
+                            "default_entries_checked": 99,
+                            "default_entries_present": 99,
+                            "topic_authority_count": 99,
+                            "project_material_count": 99,
                             "default_entry_docs": [
                                 "docs/TODO.md",
                                 "docs/meta/DOC_REGISTRY.yaml",
@@ -197,6 +205,7 @@ def main() -> int:
         assert "planned_adapter_requires_controller_opt_in" not in after["residual_risks"], after
         assert "do not append real run history or grant write-control" in after["recommended_action"], after
         project_map = after["project_map"]
+        assert project_map["authority_registry_path_exists"] is True, project_map
         assert project_map["authority_registry_default_entry_count"] == 2, project_map
         assert project_map["authority_registry_default_entries_present"] == 2, project_map
         assert project_map["topic_authority_count"] == 1, project_map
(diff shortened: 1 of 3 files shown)


huangruiteng:
## Finding

**[P1] Partial mixed registries lose unrelated legacy summary fields.** In `loopx/authority.py:155-158,745-779`, the presence of any one canonical field switches the whole registry away from `authority_registry_from_compact`. A realistic incremental state with legacy `default_entry_count=3`, `topic_authority_count=4`, `required=true`, plus a newly added canonical `project_materials` map is projected as default entries 0, topics 0, required false, and path_exists null. On current main the same state retains 3/4/true/true.

This also affects `register_authority_source`: adding the first `project_materials` map to a compact registry can overwrite unrelated default/topic summaries through `apply_authority_registry_summary`. The new smokes use zero default entries or also add a topic, so they do not cover the partial-migration case.

Suggested fix: merge precedence field by field. Derive default-entry fields only when `default_entry_docs` is present, topic count only when `topic_authority` is present, project-material stats only when `project_materials` is present, and deprecated count only when `deprecated_sources` is present; retain compact values for absent canonical groups. Resolve `path_exists` from the project when available, otherwise preserve the compact value. Add one mixed fixture with nonzero unrelated compact counts and exactly one canonical map.

The motivation is correct: canonical maps must beat stale co-located counts. The all-or-nothing switch is too broad for incremental registration/import, which is precisely where mixed states arise. Field-group precedence is the reusable compatibility contract.

## Validation

- GitHub pytest: passed
- focused `register-authority-source-smoke.py`: passed
- focused `project-map-smoke.py`: passed
- minimal mixed-state replay: reproduces the regression described above

## Merge decision

Request changes / hold until partial mixed-state compatibility is covered.

huangruiteng:
## Finding resolution

The partial mixed-registry blocker is resolved at `f030cca0`.

- Summary fields now start from their co-located compact values.
- `default_entry_docs`, `topic_authority`, `project_materials`, and `deprecated_sources` override only their own summary group when that canonical group is present.
- `required` preserves registry/goal authority.
- `path_exists` is recomputed only when a project path is available; otherwise the compact value is retained.
- The durable registration smoke now covers a project-material-only migration that must retain unrelated default/topic counts, alongside the existing all-canonical stale-count case.

This closes the incremental-migration failure without adding another authority source or a special registration path. Field-group precedence is the reusable compatibility contract: canonical source maps win where they exist, while not-yet-migrated compact groups remain readable. The main residual risk is malformed non-integer legacy counters, which is unchanged from the pre-PR compact parser.

## Validation

- Focused authority/read-map smokes: 5 passed
- Full pytest with owner-local runner profile discovery isolated: 750 passed, 2 skipped
- `loopx canary premerge --from-git-diff`: 16/16 passed, 0 failures, 0 manual holds
- Python compile and `git diff --check`: passed
- Public/private boundary scan: clean

## Merge decision

Owner authorized self-merge. I will squash self-merge after GitHub CI passes on `f030cca0`.

PR #2313, opened 2026-07-18
feat(agents): project bounded material handoffs

huangruiteng:
## Summary

- add `agent_material_handoff_projection_v0`, reducing a full frontier to six summary counts and at most four `{material_id, relation, purpose}` refs
- enrich an existing typed handoff as `handoff_note_v1` without changing the legacy todo handoff producer
- let `agent_management_projection_v0` consume optional prebuilt `status.agent_material_frontiers` packets on its cold path
- prove a successor rebuilds from current goal authority and its own requirements instead of inheriting predecessor receipts, permissions, or authority

## Boundary

This does not add material authoring, receipt writing, source-body caching, permission transfer, dispatcher/runtime behavior, or a new task identity. Without `agent_material_frontiers`, agent-management output remains unchanged.

## Validation

- focused material handoff, existing material frontier, live agent-management, and management protocol smokes: passed
- full pytest in an isolated HOME: 752 passed, 2 skipped
- Ruff and changed-file compile: passed
- new module isolated mypy: passed
- standard premerge: 18/18 passed (9 catalog, 8 risk-profile, 1 public-boundary), 0 warnings, 0 manual holds

A direct mypy invocation through the existing management module still reports the same 10 baseline errors on clean `origin/main`; the changed module introduces no additional mypy finding.

## Merge hold

This is a control-plane protocol extension and should receive independent review before merge.

### DIFF
--- a/docs/reference/protocols/agent-management-projection-v0.md
+++ b/docs/reference/protocols/agent-management-projection-v0.md
@@ -33,6 +33,8 @@ The projection is derived from:
 - quota `agent_lane_next_action`, `interaction_contract`, and scheduler hints;
 - compact run history and the agent-scoped evidence ledger;
 - task graph, handoff, and review-packet projections when present.
+- optional prebuilt `agent_material_frontier_v0` packets on the cold-path
+  `agent_material_frontiers` input.
 
 The projection is stale after any lifecycle event until recomputed. Consumers
 must tolerate missing fields and fall back to existing status/review-packet
@@ -106,6 +108,8 @@ Optional fields:
 - `scheduler_state`;
 - `workspace_ref`;
 - `handoff_refs`;
+- `handoff_note`;
+- `material_frontier`;
 - `stale_claim_hint`;
 - `blocked_on`;
 - `recent_events`;
@@ -155,7 +159,8 @@ model.
 ## Handoff Notes
 
 Inter-agent handoff should appear as a typed note attached to existing todo,
-history, and evidence refs:
+history, and evidence refs. Rows without material context retain the existing
+`handoff_note_v0` shape:
 
 ```json
 {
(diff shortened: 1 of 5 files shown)


huangruiteng:
Pre-merge validation summary:

- **Changed surfaces:** bounded agent material handoff projector, optional agent-management cold-path join, protocol docs, and one durable end-to-end smoke.
- **Direct checks:** committed/staged/unstaged diff checks and changed-file compile passed.
- **Focused checks:** new handoff projection, existing frontier, live management, and management protocol smokes passed; Ruff passed; isolated mypy for the new module passed.
- **Full suite:** 752 passed, 2 skipped in an isolated HOME.
- **Catalog canaries:** 9/9 passed.
- **Risk-profile smokes:** 8/8 passed across core control-plane, docs/project content, and canary-runner profiles.
- **Public/private boundary:** 1/1 passed; no material bodies, receipts, permissions, authority payloads, private paths, credentials, or raw evidence are forwarded.
- **Failures/skips:** no validation failure; two existing pytest skips. Direct mypy through management reproduces the same 10 errors on clean origin/main and adds no changed-line error.
- **Manual holds:** none from the premerge gate.
- **Merge decision:** hold for independent review because this extends a control-plane protocol. Coverage is sufficient because it proves bounded truncation, legacy no-input compatibility, successor authority/receipt isolation, adjacent live projection behavior, and aggregate control-plane/public-boundary regressions.

huangruiteng:
## Post-merge independent review

### Findings

- **[P1] The optional material-frontier cold path can join another goal's frontier into the current goal's agent row.** `loopx/control_plane/agents/management_projection.py:192` indexes prebuilt frontiers only by `agent_id`, and the join at `:540` never checks `frontier.goal_id` against `status.goal_filter`, the current todo's `goal_id`, or the row's goal set. A minimal public fixture with a `goal-a` todo/row and a `goal-b` frontier for the same agent emits `goal-b-private-plan` under the `goal-a` projection. The same keying also makes multiple frontiers for one agent overwrite each other by input order. This breaks the goal-owned authority boundary and can misdirect successor reconstruction even though the refs are bounded/public-safe.

  Suggested repair: key frontiers by `(goal_id, agent_id)`; select only an exact current-todo/goal-filter match; fail closed or omit enrichment when the row spans goals without a unique match; add cross-goal and same-agent/multi-frontier regression cases to `agent-material-handoff-projection-smoke.py`.

### Open question / assumption

The current status producer may usually call this projection with a goal filter, but the public builder accepts unfiltered/multi-goal payloads and the new cold-path input is optional/prebuilt. The isolation rule therefore needs to live in this join rather than depend on caller discipline.

The motivation and overall seam are sound: a bounded, read-only material projection is preferable to copying source bodies, receipts, permissions, or authority into a handoff, and the implementation does not create a dispatcher or second task runtime. The bounded-ref and successor receipt-isolation behavior is covered. However, goal ownership is part of the core contract, so joining only on agent identity is not sufficient for a reusable multi-goal control-plane read model.

### Validation

- Reproduced the cross-goal join on merge commit `897f6cff` and confirmed the same code remains on current `origin/main`.
- Passed: `agent-material-handoff-projection-smoke.py`, `agent-material-frontier-smoke.py`, `agent-management-live-status-smoke.py`, `agent-management-projection-contract-smoke.py`, `status-collection-readmodel-smoke.py`.
- Passed changed-file `py_compile` and `git diff --check`.
- GitHub `pytest` check is green. Two initially attempted smoke filenames did not exist; the canonical live-status and projection-contract smoke names above were run instead.

### Merge decision

The PR is already merged. Treat the bounded handoff seam as accepted, but hold use of the optional `agent_material_frontiers` management cold path for multi-goal payloads until the goal-scoped join regression is fixed.

PR #2316, opened 2026-07-18
Fix agent-scoped operator gate isolation

huangruiteng:
## Summary
- preserve an inherited operator gate only when the current agent still has an open blocking `user_gate`
- keep current-agent `user_action` reminders visible without blocking an independent runnable agent todo
- add a focused multi-agent regression covering reminder delivery, gate isolation, selected work, and the final interaction contract

## Validation
- `python3 examples/control_plane/quota-agent-scoped-user-gate-smoke.py`
- `python3 examples/control_plane/control-plane-risk-characterization-smoke.py`
- `python3 examples/control_plane/primary-action-resolution-parity-smoke.py`
- `python3 examples/control_plane/quota-contract-smoke.py`
- `python3 -m py_compile loopx/control_plane/agents/agent_scope.py examples/control_plane/quota-agent-scoped-user-gate-smoke.py`
- public-safe compact replay against current multi-agent state: peer decision becomes `run`, state becomes `eligible`, reminder remains `NOTIFY`, and agent delivery remains allowed
- `loopx canary premerge --from-git-diff`: selected control-plane, scheduler, gate, and boundary checks passed except the existing `peer-agent-runtime-v1-smoke.py` baseline failure; the same `agent-onboard-host-loop-activation-smoke.py` failure reproduces unchanged at `origin/main`

## Boundary
- no credentials, private state, raw benchmark evidence, local paths, or generated logs are included
- no gate authority, cross-agent mutation authority, scheduler policy, or benchmark behavior is widened

### DIFF
--- a/examples/control_plane/quota-agent-scoped-user-gate-smoke.py
+++ b/examples/control_plane/quota-agent-scoped-user-gate-smoke.py
@@ -230,6 +230,47 @@ def other_agent_gate_with_non_due_monitor_payload() -> dict:
     )
 
 
+def other_agent_gate_with_current_user_action_payload() -> dict:
+    primary_login_gate = todo_item(
+        todo_id="todo_primary_runner_login",
+        text="Log in to the primary agent runner before resuming its evaluation.",
+        role="user",
+        task_class="user_gate",
+        action_kind="runner_login",
+        blocks_agent="codex-main-control",
+    )
+    value_user_action = todo_item(
+        todo_id="todo_value_review_material",
+        text="Review the value-lane handoff material when convenient.",
+        role="user",
+        task_class="user_action",
+        action_kind="review_handoff_material",
+        claimed_by="codex-value-explorer",
+        priority="P1",
+    )
+    value_agent_todo = todo_item(
+        todo_id="todo_value_review_pr",
+        text="[P1] Review the value-lane projection PR.",
+        claimed_by="codex-value-explorer",
+        action_kind="review_value_projection",
+        priority="P1",
+    )
+    return status_fixture_payload(
+        status="active_state_user_todo",
+        waiting_on="controller",
+        severity="action",
+        recommended_action="Primary runner login is pending.",
+        quota_state="operator_gate",
+        quota_extra={"reason": "open user gate"},
+        user_todos=todo_summary_items(
+            [value_user_action, primary_login_gate],
+            source_section="User Todo",
+        ),
+        agent_todos=todo_summary(value_agent_todo, source_section="Agent Todo"),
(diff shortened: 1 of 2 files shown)


huangruiteng:
Independent LoopX lane review (codex-value-explorer): **no blockers found**.

Reviewed the operator-gate isolation path and regression coverage:

- The runtime change is limited to `_agent_scoped_user_todo_override`: an inherited `operator_gate` is now preserved only when the current agent's scoped summary still contains an actual open `user_gate`.
- A current-agent `user_action` remains visible through the user notification channel (`NOTIFY`) while an independent runnable agent todo remains selected and delivery stays allowed.
- Existing assertions still cover the target agent's blocking gate, explicit global gates, unscoped-gate projection repair, and quiet non-due monitor lanes.
- Gate decision authority and cross-agent todo lifecycle mutation authority are unchanged; this diff does not touch either mutation surface.

Validation:
- focused quota/gate, risk characterization, primary-action parity, quota contract, and todo CLI smokes passed
- full pytest: `752 passed, 2 skipped`
- Ruff and changed-file compile passed
- standard risk-based premerge command exited successfully; 18 checks were selected, with the known inherited architecture-budget advisory remaining non-blocking
- GitHub `pytest` check passed

The local default Python is unsupported 3.9; CLI/full-suite validation was rerun under Python 3.12, matching `requires-python >=3.11`.

clementineCU:
The regression looks more reviewable if the replay fixture records the isolation boundary explicitly.

Suggested receipt fields for the gate case:
- parent agent id and current agent id
- inherited operator_gate id, if any
- whether the current scoped summary still has an open user_gate
- selected runnable todo
- user_action reminder emitted as NOTIFY
- peer-agent delivery allowed/blocked
- authority that would make the gate block again

The important negative case is: inherited gate exists, current agent has no open user_gate, peer todo is runnable, so the peer must not be blocked. If that flips later, the canary should fail before the scheduler treats a stale human gate as live authority.

huangruiteng:
Closeout note on the later replay-receipt suggestion: the durable smoke already makes the isolation boundary explicit without adding a new production receipt schema. It fixes the current lane as `VALUE_AGENT`, records the inherited gate via `blocks_agent=codex-main-control`, asserts the scoped summary has no current-agent user gate but retains the other-agent gate id, asserts the selected runnable todo id, preserves `user_action` as `NOTIFY`, and requires peer delivery to remain allowed. Existing companion cases assert the authority conditions that make the gate block again for the target agent and for global/unscoped gates.

I am keeping this as test-fixture evidence rather than widening the runtime payload solely for a review aid.

PR #2318, opened 2026-07-18
fix(quota): preserve monitor poll scheduler context

huangruiteng:
## Summary
- include quota monitor-poll in the shared scheduler runtime-context parser
- preserve the resolved context across both pre-write and post-write should-run decisions
- add a real CLI regression proving --codex-app stays applicable before and after monitor writeback

## Validation
- monitor-poll-writeback smoke passed under Python 3.12
- quota-spend-agent-identity smoke passed
- heartbeat-quota-flow smoke passed
- standard diff-based premerge passed: 18/18 selected checks, no warnings or manual holds, public-boundary clean
- changed CLI/smoke Ruff checks and Python compile passed
- full pytest: 752 passed, 2 skipped, 1 inherited SkillsBench launcher failure; the exact failing test reproduces on clean origin/main under the same host environment
- optional control-plane-integrated canary also reproduces its unrelated stale multi-agent todo --agent-id failure on clean origin/main

## Scope
No scheduler cadence policy, quota spending, todo semantics, private state, credentials, benchmark task behavior, or external actions changed.

### DIFF
--- a/examples/control_plane/monitor-poll-writeback-smoke.py
+++ b/examples/control_plane/monitor-poll-writeback-smoke.py
@@ -795,6 +795,30 @@ def assert_cli_help_names_capability_sensitive_commands() -> None:
         assert command in option_help, (command, option_help)
 
 
+def assert_cli_monitor_poll_preserves_codex_app_scheduler_context() -> None:
+    with tempfile.TemporaryDirectory(prefix="loopx-monitor-poll-codex-app-") as tmp:
+        registry_path, _state_file = write_fixture(Path(tmp))
+        payload = run_cli(
+            registry_path,
+            "quota",
+            "monitor-poll",
+            "--goal-id",
+            GOAL_ID,
+            "--agent-id",
+            AGENT_ID,
+            "--codex-app",
+            "--todo-id",
+            TODO_ID,
+            "--result-hash",
+            "old",
+        )
+
+        for decision in (payload["before"], payload["after"]):
+            scheduler_hint = decision["scheduler_hint"]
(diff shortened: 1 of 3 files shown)


huangruiteng:
Post-merge independent review

**Findings**
- No behavioral blocker found. `monitor-poll` now resolves the same typed scheduler context as `should-run` and passes the identical resolution into both the before and after decisions.
- P2 maintainability follow-up: `loopx/quota.py` places `scheduler_execution_context` on the same line as the preceding parameter and call argument. This does not change behavior, but it makes a high-risk wrapper harder to audit. The follow-up should normalize the signature/calls and add explicit monitor-poll CLI parity cases for outer-controller and invalid-context routes; the current new end-to-end case names Codex App only.

**Product/architecture judgment**
- Motivation: prevent a Codex App monitor writeback from losing host authority between its before/after quota decisions and falsely entering scheduler-context repair.
- Solved: yes. The CLI derives context for `monitor-poll`, the wrapper forwards it to both decisions, and the actual CLI smoke proves App applicability survives. No-spend/writeback semantics are untouched.
- Main risk: future wrapper drift across runtime profiles; the typed resolver remains the correct shared contract.

**Validation**
- GitHub `pytest` check passed on the PR head.
- `tests/control_plane/test_scheduler_execution_context.py`: 76 passed.
- `examples/control_plane/monitor-poll-writeback-smoke.py`: passed, including real `--codex-app` before/after assertions and handler propagation.

**Decision**: post-merge review passes; record the formatting and cross-profile wrapper coverage as a non-blocking P2 successor.

PR #1787, opened 2026-07-10
feat(control-plane): hard cut agent hierarchy to peer runtime

huangruiteng:
## Summary

- make `peer_v1` the only live registered-agent runtime model; authority comes from claims/leases, task boundaries, capabilities, typed continuation, and repository policy
- replace persistent main/side/controller routing with deterministic task-scoped coordination, symmetric workspace guards, and explicit `review_handoff`
- keep host-launched child workers separate from durable registered peers; dormant agents and closed/blocked/deferred todos cannot become task coordinators
- migrate legacy registries through one stable migration id: host updates are idempotent, registry completion is atomic, repeated ack is a no-op, and stale legacy rewrites never wake the user with the same migration again
- upgrade advisory profiles to `agent_profile_v1`, remove identity-level reviewer/workspace authority, and hard cut canonical status/quota/prompt surfaces
- align README, skills, protocols, fixtures, examples, generated showcases, and agent onboarding with the shipped peer runtime

## Commit grouping

- `d8f57a5c` / `7ecd0839`: peer runtime and complete runtime/API hard cut
- `a17f24cf` / `9de87b8b`: protocol and public documentation cutover
- `699df43b` / `7e890c7f`: migration, continuation, workspace, and boundary regression coverage
- `8e482bb7`: latest-main integration plus exactly-once migration, participant-only coordinator selection, README, and residual hierarchy cleanup
- `ddc040f8`: final clean merge from current `origin/main`

## Validation

- focused peer runtime, migration, workspace, continuation, task orchestration, quota/replan, README, and hard-cut boundary smokes passed
- full-public runner: 533 checks executed; 532 passed, 0 timeouts, 0 tracked side effects
- the sole failure, `examples/codex-cli-long-run-benchmark-smoke.py`, was reproduced unchanged on detached latest `origin/main`: both treatment and baseline score `0.875` because overhead cancels writeback quality; it is not introduced by this PR
- `python3 -m compileall -q loopx`
- `python3 examples/control_plane/repo-python-line-budget-smoke.py`: 995 tracked Python files, 0 over-budget violations
- `python3 -m loopx.cli canary premerge --from-git-diff`: 17/17 automated checks passed; public boundary passed; 0 automated failures
- `loopx check` on README/docs/examples: 0 errors, 0 warnings
- `git diff --check` and private/credential/path scan passed

The canary correctly reports one `benchmark_sensitive` manual hold. This is not an automated failure; it prevents self-merge and requires maintainer review.

## Merge hold

[...]

PR #2330, opened 2026-07-18
fix(todos): validate runtime model at completion boundary

huangruiteng:
## Summary
- validate the registry-declared agent runtime model before todo completion writes any state
- keep the hard-cut `peer_v1` runtime contract while accepting the typed `legacy_hierarchy` migration input
- remove the now-unconsumed `CompletionPolicy.agent_model` projection and align the continuation smoke with its actual claim/exclusion contract
- replace the broad base-green suite with a focused base-red/head-green atomic rejection test

## Semantic differential
- merge base: 2 focused failures (`agent_model` remained projected; unknown `hierarchy_v2` completed successfully)
- candidate: 2 focused tests passed; unknown model is rejected before state write and the state file remains unchanged

## Validation
- `pytest -q tests/control_plane`: 412 passed
- risk-based premerge: 18 selected, 18 executed, 0 failures, 0 manual holds
- peer runtime aggregate smoke: passed
- Ruff, Python compile, diff hygiene, and public/private boundary scan: passed

## Advisory
- the current `origin/main` maintainability ratchet reports one inherited stale exception for `build_scheduler_hint.hint`; premerge classified it as advisory and unrelated to this diff

## Merge decision
Owner authorized self-merge after the review finding was fixed. The semantic differential and full risk-based gate now pass.


huangruiteng:
Independent quality review: **resolved**.

### Correction to the original base comparison

The first comparison ran a candidate test file by absolute path. Pytest selected the candidate repository as its root and imported candidate sources, so the reported `4 passed` result did **not** prove merge-base parity. After physically placing the focused test in merge-base worktree `2a1239d8`, the semantic differential is:

- merge base: 2 failed
  - `CompletionPolicy.agent_model` was still projected even though production never consumed it;
  - unknown `coordination.agent_model=hierarchy_v2` completed successfully instead of failing before write.
- candidate: 2 passed.

### Resolution

The final implementation preserves the part of #2330 that fixes a real boundary defect: `agent_runtime_model_for_goal(...)` validates the registry model before completion writes state. It removes the migration-era, unconsumed `CompletionPolicy.agent_model` return field and the stale smoke assertion that treated that projection as behavior.

This retains the peer-runtime hard-cut safety introduced in #1787 without making completion policy a second owner of runtime identity.

### Validation

- focused semantic test: base 2 failed, candidate 2 passed
- `tests/control_plane`: 412 passed
- risk-based premerge: 18/18 passed, 0 manual holds
- peer runtime aggregate, Ruff, compile, diff hygiene, and public-boundary scan: passed
- one current-main maintainability stale-exception result was classified as inherited advisory

Detailed follow-up: https://github.com/huangruiteng/loopx/pull/2330#issuecomment-5013862927

Squash merged as `3019772fe7c6e1cfacb6ccc9d9660c96943550bf`.

huangruiteng:
Addressed the earlier review hold.

The original `agent_model` field came from the peer-runtime hard cut in #1787, where it served as a migration-era contract marker. The production completion path never consumed the field, but this PR did expose a real boundary gap: an unknown registry model could reach todo completion without validation.

The final diff therefore separates those concerns:
- keeps `agent_runtime_model_for_goal(...)` at the completion input boundary, so unknown models fail before state mutation;
- removes the unconsumed `CompletionPolicy.agent_model` projection;
- removes the stale smoke assertion that treated the projection as behavior;
- adds a focused atomic-rejection test.

The previous base comparison used an absolute candidate test path, which let pytest import candidate sources. Re-running the identical test after physically placing it in the merge-base worktree gives the intended semantic differential: base 2 failed, candidate 2 passed.

Validation: 412 control-plane tests passed; risk-based premerge selected/executed 18 checks with 0 failures and 0 manual holds; peer runtime aggregate, Ruff, compile, diff hygiene, and public-boundary scan passed. One maintainability stale-exception result from current main was classified as inherited advisory.

Owner authorization for self-merge is recorded in the task. With the review finding fixed and the gate reporting `self_merge_allowed=true`, I am proceeding with squash merge.

PR #2334, opened 2026-07-18
fix(agents): isolate material frontiers by goal

huangruiteng:
## Summary

- key prebuilt agent material frontiers by `(goal_id, agent_id)` instead of agent identity alone
- resolve the cold-path join from an explicit goal filter, the current todo goal, or one unambiguous row goal; omit enrichment when a multi-goal row cannot be resolved safely
- extend the durable handoff smoke with cross-goal, same-agent/multi-frontier, current-todo, and ambiguous-row regressions, and document the goal-isolation rule

## Validation

- `git diff --check`
- Ruff on both changed Python files
- adjacent frontier, handoff, live-management, projection-contract, and status-collection smokes
- isolated-HOME full suite: `791 passed, 2 skipped`
- `loopx canary premerge --from-git-diff`: 18/18 passed, including 9 catalog canaries, 8 risk-profile smokes, and the public/private boundary scan
- focused mypy remains non-green with the same 10 pre-existing errors documented during #2313 review; the changed join adds no new reported line

The first premerge attempt exposed an environment issue: nested `scripts/loopx` selected macOS `/usr/bin/python3` 3.9 and correctly failed the Python 3.11+ runtime guard. Re-running with `LOOPX_PYTHON` bound to the validated Python 3.13 venv passed the complete peer-runtime smoke and the full premerge gate.

## Boundary

Only public control-plane code, its durable smoke, and the protocol reference are included. No local state, raw material bodies, receipts, permissions, private paths, credentials, or generated logs are committed.

## Residual risk

The management projection still aggregates one agent row across goals. This change deliberately does not redesign that read model; it only makes optional material enrichment fail closed unless the row has a unique goal context.

### DIFF
--- a/docs/reference/protocols/agent-management-projection-v0.md
+++ b/docs/reference/protocols/agent-management-projection-v0.md
@@ -229,6 +229,12 @@ from current goal authority and its own agent-scoped requirements and receipts.
 The optional cold-path input does not create an agent row, claim, or task by
 itself.
 
+The cold-path join is scoped by `(goal_id, agent_id)`, not agent identity alone.
+An explicit status goal filter takes precedence, followed by the current todo's
+goal and then a single unambiguous row goal. When a multi-goal agent row has no
+unique goal context, LoopX omits the material enrichment instead of selecting a
+frontier by input order.
+
 ## Stale Claim Hint
 
 `stale_claim_hint` is an observability warning, not an automatic reclaim rule.
--- a/examples/control_plane/agent-material-handoff-projection-smoke.py
+++ b/examples/control_plane/agent-material-handoff-projection-smoke.py
@@ -25,6 +25,7 @@ from loopx.control_plane.agents.material_handoff import (  # noqa: E402
 
 
 GOAL_ID = "material-handoff-fixture"
+OTHER_GOAL_ID = "other-material-handoff-fixture"
 SOURCE_AGENT = "agent-builder"
 SUCCESSOR_AGENT = "agent-reviewer"
 SOURCE_TODO = "todo_material_source"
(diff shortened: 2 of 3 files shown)


huangruiteng:
Independent review at exact head `8ac7536ce5b366d8838a7ddc8b1d6f71d5084825`.

**Findings**
- No blocking correctness, compatibility, or public-boundary findings.
- The state rule is sound: material enrichment is keyed by exact `(goal_id, agent_id)`; an explicit goal filter is accepted only when consistent with the row/current todo, otherwise the selected current todo disambiguates, then a single row goal may disambiguate. Multi-goal rows with no current todo fail closed by omitting enrichment.
- The change does not make the aggregated agent row authoritative and does not add write authority; it only prevents optional material data from crossing goal boundaries.

**Assumptions and residual risk**
- The existing management projection still aggregates one agent row across goals. That broader read-model redesign is intentionally outside this PR; the new join prevents it from leaking a frontier while the row remains ambiguous.
- Review and merge are scoped to the immutable head above.

**Product judgment**
- This is a cohesive, right-sized correction at the existing read-model join. The regression checks semantic ownership and ambiguity behavior rather than snapshotting incidental output.

**Validation**
- Base/head qualification: the candidate regression fails on base `0bacf01821dcf7eee0b9871489db677b7d3b65db` by attaching `other-goal-only` to the current goal; it passes on head.
- Current-main integration: clean no-commit merge onto `a54fdcf511a0747c4556390a91acef5ad7815dbd`; focused smoke passed.
- Six adjacent material-frontier, handoff, management, projection-contract, and status read-model smokes: passed.
- Isolated-HOME current-main integration suite on Python 3.13: `791 passed, 2 skipped`.
- Ruff on both changed Python files: passed.
- Standard integration premerge: 4 direct checks plus 18/18 catalog/risk/boundary checks passed; 0 failures, 0 warnings, 0 manual holds; public/private boundary clean.
- GitHub Actions `pytest`: passed.

**Merge decision**
- Approved for exact-head squash merge. The integration premerge gate reports `self_merge_allowed=true`.

PR #2335, opened 2026-07-18
Make successor routing metadata atomic

huangruiteng:
## Summary
- let `todo complete/supersede --next-agent-todo` carry successor `task_repository` and repeatable `required_capabilities` in the same mutation
- preserve those fields across materialized state, event writeback, replay, backfill, and structured projection
- document and regress the excluded-implementer to independent-review handoff

## Validation
- focused materialized lifecycle smoke: passed
- focused event projection/replay smoke: passed
- Ruff on touched Python: passed
- isolated-HOME full pytest: 791 passed, 2 skipped
- `loopx canary premerge --from-git-diff --tier standard`: 18/18 passed, 0 warnings, 0 manual holds

The first non-isolated full-suite run inherited an owner-local SkillsBench runner profile and failed one validator-absence expectation; rerunning that test and the full suite with an isolated HOME passed. No benchmark launch, scoring, permission, scheduler, or merge-authority behavior changes.

### DIFF
--- a/docs/project-agent-todo-contract.md
+++ b/docs/project-agent-todo-contract.md
@@ -473,7 +473,11 @@ loopx todo complete \
 Without `--self-merged`, no implicit review route is created. For independent
 review, use `--next-action-kind review --next-continuation-policy
 independent_handoff`; add `--next-excluded-agent <author>` only when the review
-must remain open to multiple peers while excluding the author.
+must remain open to multiple peers while excluding the author. When that
+successor belongs to a specific repository or needs execution capabilities,
+set `--next-task-repository <git:host/path>` and repeat
+`--next-required-capability <capability>` in the same completion command. The
+successor is then fully routable before its executor exclusions take effect.
 
 Use `todo update` for lower-level, non-terminal status changes:
 
--- a/examples/control_plane/todo-lifecycle-cli-smoke.py
+++ b/examples/control_plane/todo-lifecycle-cli-smoke.py
@@ -40,6 +40,10 @@ from todo_lifecycle_fixtures import (  # noqa: E402
 )
 
 
+SUCCESSOR_REPOSITORY = "git:github.com/huangruiteng/loopx"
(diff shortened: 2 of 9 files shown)


huangruiteng:
Independent review at exact head `b331d8fed85ab68f585157c7924630c29edb3133`.

**Findings**
- No blocking correctness, compatibility, or public-boundary findings.
- The new routing fields are carried through both materialized-state creation and event-projected successor creation, then normalized through replay/backfill and structured projection. Executor exclusion still applies only after the successor has enough repository/capability metadata to be runnable.
- Residual test gap: the committed regression directly covers `complete` on materialized and event-projected paths, while `supersede` uses the same new contract without a dedicated committed assertion. I ran an independent supersede CLI probe covering repository, repeated capabilities, claim, and excluded-agent metadata; it passed. This is non-blocking because the implementation delegates to the already-covered `add_todo_to_lines` path.

**Assumptions**
- Review and merge are scoped to the immutable head above.
- No benchmark launch, scoring, scheduler, permission, or merge-authority semantics are intended to change.

**Product judgment**
- This is the right-sized fix for the atomic handoff gap: it adds only the two routing inputs required by an immediately active successor and preserves the existing todo lifecycle authority model.

**Validation**
- Base/head qualification: both new regression smokes fail on base `074000933ce7b6e25610c21988a034548439b33e` because the CLI lacks the routing arguments, and pass on head.
- Focused materialized lifecycle smoke: passed.
- Focused event projection/replay smoke: passed.
- Independent supersede atomic-routing probe: passed.
- Isolated-HOME full suite: `791 passed, 2 skipped`.
- Ruff on all touched Python files: passed.
- Standard premerge: 4 direct checks plus 18/18 catalog/risk/boundary checks passed; 0 failures, 0 warnings, 0 manual holds; public/private boundary clean.
- GitHub Actions `pytest`: passed.

**Merge decision**
- Approved for exact-head squash merge. The standard premerge gate reports `self_merge_allowed=true`.

PR #2340, opened 2026-07-19
refactor(capabilities): bind connector profiles to outcome owners

huangruiteng:
## Summary

- freeze `value-connectors` as the compatibility CLI/protocol facade and project an explicit outcome owner plus migration state for every GitHub, content, social, Agent-Reach, finance, Botmail, and community profile
- move the proven public GitHub probe/reply-monitor implementation and protocol ownership to `issue-fix`; keep existing CLI commands and packet schemas unchanged
- split generic install checks from the GitHub provider, remove the obsolete internal Python import path, and document the ownership model

## Validation

- `uvx --from ruff==0.12.11 ruff check ...` (changed Python surfaces): passed
- `examples/value-connectors-github-public-probe-smoke.py`: passed
- `examples/capability-extension-registry-smoke.py`: passed
- `examples/capability-extension-placement-doc-smoke.py`: passed
- `loopx canary premerge --from-git-diff`: passed, 4 direct checks plus 18/18 selected checks, public-boundary scan clean, no warnings or manual holds

## Failures / skips

- no final failures or skipped required checks
- two earlier premerge attempts hit host interpreter selection and then exhausted local temporary disk during installer smokes; the supported interpreter was bound, stale test-only temp directories were removed, both installer smokes passed independently, and the exact final HEAD passed the complete gate

## Scope rationale

The facade remains because its CLI and packet schemas are existing caller contracts. This PR migrates only the proven GitHub implementation; other profiles receive explicit owners without speculative provider modules. No old Python import shim is retained because repository callers are migrated and the import path was not a supported public contract.

### DIFF
--- a/docs/capabilities/issue-fix/README.md
+++ b/docs/capabilities/issue-fix/README.md
@@ -450,6 +450,30 @@ sync the management surface, record residual risk, and choose one of:
 One merged PR proves a delivery slice. Repeating the loop on independent issues
 tests whether the system is a durable employee rather than a scripted demo.
 
+## Public GitHub Signal Provider
+
+Issue-fix owns the body-free public GitHub probe and reply-monitor provider in
+`loopx.capabilities.issue_fix.github_public`. Existing CLI callers keep using
+the compatibility commands:
+
+```bash
+loopx value-connectors github-public-probe \
+  --url https://github.com/owner/repo/issues/1 \
+  --fetch-metadata \
+  --format json
+
+loopx value-connectors github-reply-monitor \
+  --issue-url https://github.com/owner/repo/issues/1 \
+  --after-comment-url https://github.com/owner/repo/issues/1#issuecomment-123 \
+  --fetch-metadata \
+  --format json
+```
+
+The CLI name and packet schemas remain stable. Only implementation ownership
+moved: probes and monitor signals now evolve with the issue-to-PR outcome they
+feed, while connector installation and generic approval planning remain in the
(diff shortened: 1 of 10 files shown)


huangruiteng:
## Self-review

**Findings:** no blocking findings.

**Assumptions:** `loopx value-connectors ...` commands and packet schemas are the supported compatibility contract; the old internal Python module path is not. `finance-value-discovery` remains a mapped future outcome capability, so its `provider_module` is intentionally null rather than speculative.

**Product / architecture judgment:** the change solves the broad-bucket problem without adding another framework. The facade keeps shared install/source-map/plan behavior, every profile gets an explicit outcome owner and honest binding state, and only the proven GitHub provider moves to `issue-fix`. This keeps future profile moves local and reversible. Main risk is accidental caller breakage from the module move; repository-wide import search, CLI parity smoke, package install checks, and catalog ownership assertions cover that risk.

**Validation:** Ruff passed on changed Python surfaces; focused value-connectors, capability-registry, and placement smokes passed; exact-head risk-based premerge passed 4 direct checks plus 18/18 selected checks; public-boundary scan was clean; merge-tree against current `origin/main` is clean. Earlier installer retries exposed only host Python selection and temporary-disk exhaustion; after binding Python 3.13 and cleaning stale test-only temp data, the exact head passed the complete gate.

**Merge decision:** self-merge is authorized by the owner and locally ready; hold only until required GitHub pytest is green on exact head `6adc1ba6`.

PR #2346, opened 2026-07-19
Add composed Doubao 2.1 control-plane qualification

huangruiteng:
## Summary
- add three production-generated control-plane composition scenarios to the actual-default one-arm portfolio
- qualify required-vision replan, non-blocking user notice plus ready successor, and capability fallback into monitor-schedule repair
- preserve exact vision trigger kinds in the compact TurnEnvelope and align primary action with the selected monitor repair
- expand the regular bounded portfolio from 9 scenarios / 18 calls to 12 scenarios / 24 calls and update the release receipt contract

## Why
Single-field fixtures do not exercise the decisions that strand long-running lanes. These scenarios intentionally combine individually reasonable but competing signals and judge the intended state semantics before checking model alignment.

The new qualification exposed two real contract issues:
1. the portfolio oracle treated every user action as a blocking gate instead of following the typed response plan;
2. capability fallback selected monitor repair but still described the capability-blocked advancement as the primary action.

## Validation
- `pytest`: 32 passed across the portfolio, release receipt, scoped user-gate lane progress, and vision-blocked successor suites
- focused control-plane smokes: monitor scheduler, quota replan decision plane, and interaction state machine passed
- `ruff check`: passed for all changed Python surfaces
- `loopx canary premerge --from-git-diff` under the project runtime: 18 selected, 18 passed, 0 manual holds
- Doubao 2.1 live portfolio: not run because `ARK_API_KEY` was unavailable to the clean candidate process; no provider calls were attempted

## Qualification policy
The regular route remains the actual-default one-arm packet with two repeats per scenario. No retired dual-arm behavior is restored. Live Doubao remains a low-frequency local/manual release gate rather than ordinary CI.


huangruiteng:
Validation update for head 0c2b1e8a: GitHub pytest passed. Local coverage remains 32 focused tests, three control-plane smokes, Ruff, the public/private boundary scan, and 18/18 project-runtime premerge checks with zero manual holds. The changed surfaces are TurnEnvelope vision projection, primary-action fallback selection, the actual-default model-behavior portfolio, release receipts, and bilingual qualification docs. Doubao 2.1 live remains intentionally unclaimed because ARK_API_KEY was unavailable to the clean candidate process; no provider call was attempted. This coverage is sufficient for review of the deterministic/product contract, while the 12-scenario x 2 live receipt remains a separate local/manual release gate.

PR #2349, opened 2026-07-19
fix(control-plane): isolate replan stalls by agent

huangruiteng:
## Summary
- scope autonomous replan run-history windows to the newest attributable agent lane
- preserve unscoped goal-level runs while excluding interleaved peer records
- add a regression for two identical blocked-successor waits separated by another peer monitor

## Why
A multi-agent goal could record two unchanged waits for one agent, but an interleaved peer monitor changed the global two-run window. The detector then saw different monitor targets and skipped the required bounded replan.

## Validation
- `23 passed` across `test_goal_vision_blocked_successor.py` and `test_quota_slot_accounting.py`
- `uvx ruff check` on changed Python files
- `loopx canary premerge --from-git-diff`: 16/16 selected checks passed, 0 manual holds
- replayed the active multi-agent run history and observed the expected agent-owned `autonomous_replan_required` decision

## Boundary
Only public control-plane code and a focused regression test are included. Local state, raw run records, credentials, absolute paths, and private evidence are excluded.

### DIFF
--- a/loopx/control_plane/work_items/autonomous_replan_obligation.py
+++ b/loopx/control_plane/work_items/autonomous_replan_obligation.py
@@ -57,6 +57,44 @@ def _single_public_agent_id(items: list[dict[str, Any]]) -> str | None:
     return next(iter(agent_ids)) if len(agent_ids) == 1 else None
 
 
+def _run_history_agent_id(run: dict[str, Any]) -> str | None:
+    agent_id = str(run.get("agent_id") or "").strip()
+    if agent_id:
+        return agent_id
+    monitor_target = run_history_monitor_target(run)
+    if isinstance(monitor_target, dict):
+        return str(monitor_target.get("agent_id") or "").strip() or None
+    return None
+
+
+def _latest_agent_run_history(
+    latest_runs: list[dict[str, Any]] | None,
+    *,
+    neutral_classifications: set[str],
+) -> list[dict[str, Any]]:
+    """Keep the newest attributable agent lane while preserving goal-level runs."""
+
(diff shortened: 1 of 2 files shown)


huangruiteng:
## Review

**Findings:** no correctness, boundary, or test-quality blockers found in the final diff.

**Assumption:** run history is newest-first, and an unscoped run remains goal-level evidence that may interrupt every agent lane. Explicitly attributed peer runs must not interrupt or satisfy another lane's stall threshold.

**Product / architecture judgment:** the bug was in the reduction boundary, not monitor writeback. The previous reducer treated a shared multi-agent history as one stall window, so an unrelated peer monitor changed the target set and suppressed a required bounded replan. Scoping the reducer to the newest attributable lane preserves the existing single-obligation projection, keeps goal-level unscoped events visible, and avoids adding a second ledger or per-command special case. The regression uses a structurally valid peer monitor event and asserts ownership plus frontier identity, rather than snapshotting incidental output.

**Validation:**
- 23 focused pytest cases passed (goal_vision_blocked_successor + quota_slot_accounting)
- Ruff passed on both changed files
- premerge canary passed 16/16 selected checks, including autonomous-replan, monitor writeback, refresh correctness, and core control-plane risk profiles
- active multi-agent history replay produced the expected agent-owned autonomous_replan_required
- public/private scan clean; no local state, credentials, raw runs, or absolute paths included

**Merge decision:** approved after an independent reviewer. I am not self-merging this runtime control-plane behavior change solely on local green checks; GitHub currently reports REVIEW_REQUIRED and Python Tests is still running.

PR #2359, opened 2026-07-19
Fix SkillsBench meaningful-operation test doubles

huangruiteng:
## Summary
- allow the SkillsBench Turn test doubles to execute non-meaningful baseline setup and cleanup
- count only the independent post-agent validation as a meaningful operation
- retain an explicit assertion that exactly one meaningful operation is observed

## Validation
- `uv run --extra test python -m pytest -q -n 2 --cov=loopx --cov-report=term --cov-fail-under=19.6` (888 passed, 2 skipped, 45.38% coverage)
- `uvx ruff check tests/test_skillsbench_turn_runtime.py`
- `loopx canary premerge --from-git-diff` (4/4 catalog canaries passed; no manual holds)

### DIFF
--- a/tests/test_skillsbench_turn_runtime.py
+++ b/tests/test_skillsbench_turn_runtime.py
@@ -75,8 +75,8 @@ def test_satisfied_pre_agent_postcondition_runs_but_does_not_claim_readiness(
         ) -> dict[str, Any]:
             if allow_nonzero:
                 return {"ok": True, "exit_code": 0, "elapsed_ms": 1}
-            assert meaningful is True
-            self.meaningful_operation_count += 1
+            if meaningful:
+                self.meaningful_operation_count += 1
             return {"ok": True, "exit_code": 0, "stdout": "", "elapsed_ms": 1}
 
     def fake_turn_once(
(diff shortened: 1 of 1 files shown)


huangruiteng:
Pre-merge qualification at `29f65caf`:

- Changed surface: one SkillsBench Turn runtime test file; no runtime or scoring behavior changed.
- Exact regression: `tests/test_skillsbench_turn_runtime.py` now passes 7/7; the test doubles allow baseline file setup/cleanup without counting it as meaningful work and still assert exactly one meaningful validation operation.
- Full CI-equivalent fast suite: 888 passed, 2 skipped; coverage 45.38% (threshold 19.6%).
- Ruff: passed.
- `loopx canary premerge --from-git-diff`: 4/4 selected catalog canaries passed; no failures or manual holds.
- Public/private boundary: only public test logic changed; no credentials, local paths, raw benchmark evidence, or generated artifacts are included.

Coverage is sufficient because this is a test-double correction for the exact two failures already reproduced on `origin/main`; product runtime and benchmark semantics are unchanged.

PR #2364, opened 2026-07-19
Add first-class OpenCode host support

elegracer:
## Summary

- register OpenCode as a first-class LoopX host surface across onboarding, guided start, command catalogs, and scheduler profile selection
- install static OpenCode command facades through the default `all` surface while keeping the executable goal bridge, runtime, and dependencies behind explicit `--with-goal-bridge` opt-in
- preflight OpenCode config and package state before any OpenCode write, then provision bridge files before command files so blocked bridge installs cannot leave commands pointing at a missing runtime
- gate idle continuation, timer wakes, disposal, and validated terminal completion through `loopx quota should-run`
- document exact install/uninstall boundaries in the adapter README and add OpenCode to the repository Quick Start

## Validation

- `python3 -m pytest -q`: 900 passed, 2 skipped
- `node --test tests/opencode_goal_bridge_runtime.test.mjs`: 10 passed
- `python3 examples/slash-command-install-smoke.py`: passed
- `python3 examples/install-local-smoke.py`: passed
- CLI output base/head differential and budget regression smokes: passed
- changed-file Ruff checks: passed
- `python3 -m loopx.cli canary premerge --from-git-diff --tier standard`: passed, 18 selected checks, public boundary clean, 0 manual holds

## Install boundary

Default and `--surface all` installs write only static OpenCode command files. Enable or remove the executable bridge explicitly with:

```bash
loopx slash-commands --install --surface opencode --with-goal-bridge
loopx slash-commands --uninstall --surface opencode --with-goal-bridge
```

Bridge uninstall preserves shared `package.json` dependencies. A static-only uninstall omits `--with-goal-bridge` and leaves bridge files untouched.

## Runtime notes

- The global OpenCode config must not also register a goal plugin directly; preflight detects JSON/JSONC conflicts and fails closed before writes.
- OpenCode installs config-directory dependencies with Bun at startup, so restart OpenCode after bridge installation.
- `opencode run` is one-shot and cannot own timers after process exit; recurring operation uses the visible TUI or persistent OpenCode server.
- This changes runtime and installer behavior, so it remains open for normal review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,6 +185,10 @@ Choose your surface:
   pasteable gate.
 - **Claude Code**: best when Claude Code's native `/loop` should drive each tick.
   Install the opt-in adapter, run `/loopx <task>`, then `/loop`.
+- **OpenCode**: install the static command facade with the normal command
+  installer. For recurring quota-gated goals, explicitly run `loopx
+  slash-commands --install --surface opencode --with-goal-bridge`, restart
+  OpenCode, then run `/loopx <task>`.
 - **Manual shell / other agents**: best when you want LoopX state without a
   supported runtime bridge. Install from the no-clone installer, then run
   `loopx doctor` and `loopx bootstrap`.
@@ -192,8 +196,10 @@ Choose your surface:
 Command registration is host-specific, but the state path is not. Codex
 surfaces may expose LoopX through `$loopx` or `/skills` command facades before
 native `/loopx` exists; Claude Code can expose `/loopx <task>` after its opt-in
-adapter is installed. If a host command is missing, run `loopx slash-commands`
-for the current catalog or start the same agent-safe path from a shell with
+adapter is installed; OpenCode can expose `/loopx <task>` through its static
+command facade, while its executable goal bridge remains explicit opt-in. If a
+host command is missing, run `loopx slash-commands` for the current catalog or
+start the same agent-safe path from a shell with
 `loopx start-goal --guided --project . --goal-text "<task>"`. Host and plugin
 integrations that need the lower-level handoff packet can still use
 `loopx bootstrap-command-pack --project . --goal-text "<task>"`. Full routing and
--- a/docs/guides/getting-started.md
+++ b/docs/guides/getting-started.md
@@ -123,8 +123,8 @@ integrations that need the lower-level handoff packet can use
 manager or PR review commands, use `loopx slash-commands` to print the current
 canonical command list and fallback CLI shapes.
 
-Use `codex-app`, `codex-ide-plugin`, or `codex-cli-tui` for the corresponding
-Codex host. Select `codex-ide-plugin` only when LoopX is running through the
+Use `codex-app`, `codex-ide-plugin`, `codex-cli-tui`, or `opencode` for the
+corresponding host. Select `codex-ide-plugin` only when LoopX is running through the
 installed IDE plugin; using Codex beside an editor does not make the host an IDE
 plugin. If the exact host is not known, omit `--host-surface` once: LoopX
 returns a read-only selection gate with exact rerun commands and does not write
--- a/docs/guides/getting-started.md
+++ b/docs/guides/getting-started.md
@@ -79,6 +79,10 @@ can discover user-installed skills:
 - Claude Code: lightweight user skills under `~/.claude/skills/loopx*`, so the
   command family can appear as Claude Code slash commands without enabling the
   opt-in MCP/hook adapter.
+- OpenCode: static command files under `~/.config/opencode/commands/` expose
+  native `/loopx` slash commands after restart. The executable goal bridge
+  (timer-based idle continuation gated by LoopX quota) requires an explicit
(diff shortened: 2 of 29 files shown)


huangruiteng:
Request changes on the installation boundary.

Today the default `all` surface expands to `codex`, `claude-code`, and `opencode` in `loopx/slash_command_install.py`. The first two install static command/skill facades, while OpenCode additionally installs an executable host plugin, a 662-line runtime, persistent session/timer behavior, and package dependency declarations. That is a materially stronger side effect than the existing meaning of the default command installer.

Please separate the two layers:

- The default `all` path may install static command/skill facades for all three hosts.
- The OpenCode goal bridge should require an explicit opt-in, for example `--surface opencode --with-goal-bridge`. A reliable existing-OpenCode detection plus an explicit confirmation surface would also be acceptable, but merely adding OpenCode to `all` is not enough.
- Bridge preflight should complete before any OpenCode files are written, so a direct-plugin conflict or invalid config cannot leave command files installed without the bridge they reference.

Please also tighten `loopx/opencode_goal_mode/README.md` around this boundary. It should concisely distinguish:

1. the static OpenCode command facade installed by the default `all` path;
2. the explicitly enabled goal bridge and the additional plugin/runtime/dependency side effects;
3. the exact opt-in install and uninstall commands;
4. the runtime contract in one compact flow: OpenCode idle event -> `quota should-run` -> continue, wait, pause, or validated completion.

Avoid presenting bridge activation as an implicit consequence of ordinary command installation.

Acceptance criterion: a default install on a machine that does not use OpenCode must not activate or provision the OpenCode goal runtime, while an explicit bridge install still produces the full tested OpenCode integration and the README describes that distinction without ambiguity.

Reviewed at `4f854eee4bd5162b89c1e95b1a97ad0bd336f07b`.

huangruiteng:
仓库首页 readme 的 quickstart 也可以写上对 opencode 的支持

elegracer:
Addressed both review comments in the existing PR head.

Changed surfaces:
- Default `all` now installs the static OpenCode command facade only.
- `--with-goal-bridge` is required for plugin/runtime/dependency installation and removal.
- Bridge preflight validates `opencode.json`, `opencode.jsonc`, and `package.json` before any OpenCode write. Bridge dependencies/runtime/plugin are written before command files.
- Static-only and bridge uninstall behavior now have separate regression coverage.
- The adapter README documents both layers, exact install/uninstall commands, and the compact idle -> quota -> continue/wait/pause/completion flow.
- The repository Quick Start now lists OpenCode and the explicit bridge command.

Validation:
- `python3 -m pytest -q`: 900 passed, 2 pre-existing skips
- Node bridge tests: 10 passed
- slash-command and local-install smokes: passed
- CLI output differential and budget smokes: passed
- standard premerge canary: 18 selected checks passed, public boundary clean, 0 failures, 0 manual holds

Coverage includes the requested installation boundary, invalid/direct-plugin/package preflight failures, static-only default behavior, both uninstall modes, in-flight disposal, timer isolation, coalescing, user intervention, and validated completion.

huangruiteng (approved):
**Findings**

No remaining blockers on head `355f244b`.

I found and fixed one install-path correctness bug before approval: `slash-commands --install` printed an `ok=false` blocked payload but still exited 0, so `install-local.sh` could report the OpenCode bridge as installed after preflight rejected invalid config. Commit `355f244b` now returns nonzero for blocked install payloads and adds real CLI plus installer regression coverage. Invalid OpenCode config preserves the static command facade while writing no plugin, runtime, or package dependencies.

**Open questions / assumptions**

None. The goal bridge remains explicit opt-in; default `all` installs only the static Codex, Claude Code, and OpenCode command facades.

The PR now cleanly separates host-neutral discovery from executable host integration: static facades are lightweight defaults, while OpenCode idle-event handling, quota gating, timer wakes, and completion validation live behind the explicit bridge. Preflight is fail-closed before bridge writes, and the CLI exit contract now lets every shell caller observe that failure. This solves the intended first-class OpenCode workflow without silently provisioning a runtime on machines that only need command discovery.

**Validation**

- `23 passed`: slash-command installer, OpenCode bridge, and host-loop activation pytest set.
- `10 passed`: real Node tests for the OpenCode goal bridge runtime.
- `examples/slash-command-install-smoke.py`: passed, including blocked CLI exit and zero-write assertions.
- `examples/install-local-smoke.py`: passed, including blocked bridge reporting and facade-only persistence.
- `loopx canary premerge --from-git-diff`: passed 18/18 on the exact head; 9 catalog canaries, 8 risk-profile smokes, public-boundary scan, Python compile, and diff checks all passed; 0 failures, 0 skips, 0 manual holds. The installer smokes were run with explicit Python 3.13 because macOS system Python is 3.9 and LoopX requires 3.11+.

**Merge decision**

Approved for self-merge. Coverage exercises the actual CLI, packaged/local installer, OpenCode runtime, and public boundary, which is sufficient for the changed install/runtime surfaces.

PR #2367, opened 2026-07-19
feat(opencode): add quota-gated host support

elegracer:
## Summary

- add a first-class OpenCode host activation packet and `loopx_goal_activate` bridge
- gate idle continuation, timer wakes, and terminal completion through LoopX quota
- install OpenCode commands, bridge runtime, and pinned dependencies through an explicit opt-in surface
- document the lifecycle and add Python, Node.js, installer, output-budget, and host-parity coverage

## Validation

- `python3 -m pytest -q`: 899 passed, 2 skipped
- `node --test tests/opencode_goal_bridge_runtime.test.mjs`: 10 passed
- changed-file Ruff checks: passed
- `python3 examples/install-local-smoke.py`: passed
- CLI output base/head differential and budget regression smokes: passed
- `loopx canary premerge --from-git-diff --tier standard`: passed, 0 manual holds
- default-public smoke sweep: all product checks passed; the local pytest-facade invocation hit an unusable pyenv `uvx` shim that reproduces on `origin/main`, and the same facade passed with a valid uvx environment

## Install boundary

OpenCode remains opt-in. Default and `--surface all` installs continue to provision only the existing Codex and Claude Code surfaces; use `--surface opencode` or `LOOPX_INSTALL_OPENCODE=1` explicitly.

### DIFF
--- a/docs/guides/getting-started.md
+++ b/docs/guides/getting-started.md
@@ -123,8 +123,8 @@ integrations that need the lower-level handoff packet can use
 manager or PR review commands, use `loopx slash-commands` to print the current
 canonical command list and fallback CLI shapes.
 
-Use `codex-app`, `codex-ide-plugin`, or `codex-cli-tui` for the corresponding
-Codex host. Select `codex-ide-plugin` only when LoopX is running through the
+Use `codex-app`, `codex-ide-plugin`, `codex-cli-tui`, or `opencode` for the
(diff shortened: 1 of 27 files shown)


elegracer:
Opened before the owner clarified that this work must remain off the upstream PR queue. Closing immediately; no upstream review is requested.

PR #2371, opened 2026-07-19
feat(control-plane): retain bounded path deltas

### LINKED ISSUE #2353: A user-side idea: a lightweight “look back and retain” step for bounded agent loops
#### A user-side methodology note

Hello, and thank you for the thoughtful work on LoopX.

I am not a software engineer. I am a long-term user of LingTai and a few other AI-agent systems, and I have also participated in some parts of LingTai’s maintenance from the user/contributor side. I may misunderstand some implementation details, so I am sharing this as a conceptual, user-side perspective rather than an engineering proposal. Please forgive any mistakes in the way I describe this idea.

When I came across LoopX and read its public documentation, I thought the project was excellent—careful, ambitious, and genuinely inspiring. I would like to take the liberty of sharing one small observation that grew out of my own experience using LingTai. I hope it might offer a little inspiration.

This is not a request for a specific feature or an engineering prescription. I do not know whether the idea belongs in a schema, a control-plane contract, a prompt pattern, a review surface, or nowhere at all. I only hope to share a user-side conceptual model.

### The user-side question

Over time, my experience with LingTai and other AI agents gradually led me to a personal human–AI collaboration method called the Yuanjiang Work Method (圆酱工作法). While reading LoopX’s ideas about long-running loops, replanning, evidence, and continuation, I wondered whether one part of this method might be relevant.

The core question is not simply whether an agent should continue working or generate another plan. It is:

> **Before entering the next loop, has the system truly looked back at what reality changed, retained what was earned, and released what is no longer valid?**

I understand that LoopX already contains mechanisms for stalled-turn detection, autonomous replanning, periodic review, evidence writeback, and todo adjustment. I am not proposing to replace or correct them. I only want to share one modest user-side way of thinking about what a completed loop might mean.

### What I have observed

The practical problems behind this question are:

1. a concrete task gradually expands into broader branches simply because the agent can keep generating;
2. long context accumulates old branches, temporary assumptions, and noise, so the agent drifts away from the original objective and acceptance criteria;
3. humans must repeat explanations and review work that the system completed but did not retain well;
4. the system either carries too much history forward or restarts without preserving the verified gains from the previous loop.

### The rolling-iron-hoop metaphor

A metaphor that has helped me is **rolling an iron hoop**. The system should not be locked to one rigid path, but it also should not roll forward without limit. Perhaps it can move for a bounded distance, pause, look back at the original objective and actual evidence, preserve what is real, let obsolete or disproven material leave the active working context while retaining a compact audit trace of what was rejected and why, and only begin another loop when there is a genuine trigger.

### The Yuanjiang Work Method

The method can be summarized as `claim → act → mirror → reflect → close/retain`. Its core principle is: **let reality have the right to revise the thinker**.

It also uses two short rules:

- **照无改令** — the mirror does not silently change the mandate;
- **省必留差** — reflection must preserve the actual delta.

In practical terms:

- The mirror step reveals the difference between the original claim and reality; it does not silently rewrite the goal.
- The reflection step records the actual delta, including an honest `no_change` when the evidence does not justify a change.
- The closing step preserves verified gains, rejected hypotheses and their reasons, unresolved questions, boundaries, and re-entry conditions. Its purpose is to reduce repeated human and agent work, not merely to produce a longer summary.

> **If a closing step does not reduce the human’s need to re-explain, re-check, or reconstruct the previous loop, then the loop has not truly been closed.**

### Dynamic progress without unrestricted self-evolution

I think of the loop as multidimensional. Goals, evidence, methods, agent roles, artifacts, context, and memory may each have their own smaller loops, while all of them remain anchored to the same human need and acceptance criteria. The intended outcome is **dynamic progress under human goals and real evidence**, not unrestricted self-evolution.

### A possible look-back before or during replanning

I wonder whether a very lightweight reflection before, or as part of, an autonomous replan or a new advancement obligation might help when there is repeated no-progress or scope expansion. Such a reflection might distinguish among:

- drift;
- an external blocker;
- no new evidence;
- repeated work;
- genuine completion;
- a real reason to replan.

### What I mean by a path delta

One part I could not find explicitly in the inspected documents is what I call a **path delta**: a compact account of what the previous path assumed, what reality changed, and what the new path keeps, changes, or stops.

If the outcome is `replan`, it may help to compare the old path and the new path: what assumptions the old path relied on; what reality changed; what the new path keeps, changes, and stops; why those differences answer the observed gap; who confirmed the change; and what evidence or test would validate—or later falsify—the new path. Without that comparison, a replan may only rename the same action, lose useful parts of the old path, or leave the human to ask again what actually changed.

This observation is limited to the documents I inspected, not a claim about the whole repository.

### Possible outcomes

A look-back might allow several honest outcomes—`continue`, `replan`, `wait`, `no_change`, `ask_human`, or `stop`—rather than treating replan as the only response to stalled progress.

### Relationship to #1802

I noticed that closed issue #1802 addresses the need to avoid an indefinitely quiet monitor-only successor while acceptance remains open. My thought may be adjacent rather than a replacement: before creating a new advancement or replan obligation, could the system first look back, retain the verified grain, and make the reason for another loop explicit?

### Closing

I hope this metaphor offers a little inspiration. If it does not fit LoopX’s architecture or goals, please feel no obligation to act on it. Thank you for reading.

huangruiteng:
## Summary
- add an optional bounded `goal_path_delta_v0` inside the existing agent-vision JSON packet
- preserve prior assumptions, observed reality, retained/changed/stopped path decisions, unresolved questions, re-entry conditions, and evidence refs through run history and shared runtime projection
- keep the normal Codex App automation path compact: no new inline CLI flags, heartbeat text, planner, or state machine
- document the contract and add a focused write/read-path smoke

## Design judgment
The existing vision patch already carries direction, acceptance, and replan trigger, but it did not preserve an explicit old-path-to-new-path comparison. The new object reuses `--agent-vision-json`, the existing 1,200-character vision budget, outer `agent_id` attribution, and current run-history projection. It requires an old assumption, observed reality, and at least one retained/changed/stopped disposition, while allowing honest `no_change`, `wait`, `ask_human`, and `stop` outcomes.

## Validation
- `loopx canary premerge --from-git-diff --timeout-seconds 240 --no-progress`: passed; 4 direct checks plus 18/18 selected checks, 0 failures, 0 warnings, 0 manual holds; public boundary scan clean
- `python examples/project/goal-vision-path-delta-smoke.py`
- `python examples/project/goal-vision-replan-contract-smoke.py`
- `python examples/project/goal-vision-refresh-state-budget-smoke.py`
- `pytest -q tests/control_plane/test_goal_vision_succession.py tests/control_plane/test_goal_vision_blocked_successor.py`: 28 passed
- focused Ruff and compile checks passed

## Boundary and residual risk
No private state, raw logs, credentials, local paths, or generated evidence are included. Existing inline vision writes and packets without `path_delta` retain their prior behavior. The new object is optional and JSON-only to avoid increasing routine agent-to-CLI argument complexity.

Fixes #2353


huangruiteng:
## Self-review

**Findings:** No blocking correctness, compatibility, or boundary findings after reviewing the final six-file diff. The optional object is rejected at the existing JSON write boundary when its outcome, old/new comparison, disposition, item counts, text budgets, or public-safety checks are invalid. Packets without `path_delta` keep the prior shape and behavior.

**Open questions / assumptions:** The contract intentionally does not infer path semantics from prose and does not add inline flags. Agents that need the look-back use the existing machine-generated `--agent-vision-json` packet; ordinary Codex App automation remains unchanged. Evidence refs remain public-safe pointers rather than copied artifacts.

**Product / architecture judgment:** Motivation is valid: current vision state preserves direction, acceptance, and replan triggers but not an explicit old-path-to-new-path comparison. The implementation closes that gap in the existing per-agent vision bounded context, shares its 1,200-character budget and run-history projection, and adds no planner, prompt expansion, state machine, or permission path. This is a reusable control-plane contract rather than issue-specific routing.

**Validation:**
- final `loopx canary premerge --from-git-diff --timeout-seconds 240 --no-progress`: 4 direct checks and 18/18 selected checks passed; 0 failures, 0 warnings, 0 manual holds; public boundary scan clean
- focused path-delta, goal-vision contract, and refresh-state budget smokes passed
- 28 focused goal-vision pytest cases passed
- focused Ruff, compile, and diff checks passed
- no private state, credentials, raw logs, local paths, or generated evidence included

**Coverage judgment:** The tests cover normalization failures, shared budget enforcement, compact run-history retention, shared-runtime retention, and existing vision succession/replan behavior. This is sufficient for the optional additive packet shape; the required GitHub pytest check remains the final merge gate.

**Merge decision:** Approved for admin self-merge after the required pytest check passes.

PR #2376, opened 2026-07-19
Qualify N-Turn resumed Codex CLI flow

huangruiteng:
## Summary
- generalize the disposable Codex CLI Turn qualification from a two-step branch to `--turn-count N`
- add an optional caller-stable `--turn-instance-id` so a new logical Turn with the same semantic action gets a distinct transaction key while same-id retries remain idempotent
- independently validate every marker transition, keep one opaque Codex CLI session, spend exactly once per committed Turn, and replay the final journal with zero effects

## Changed surfaces
- Turn transaction identity and `turn plan` / `turn run-once` CLI plumbing
- experimental README and Codex CLI Turn quickstart
- disposable public N-Turn qualification example
- focused transaction, session-resume, quickstart, and N=3 regression coverage

## Validation
- deterministic N=3 qualification: `start_new -> resume -> resume`, 3 validators passed, 3 commits/spends, final replay had zero effects
- deterministic default N=1 qualification passed
- focused Turn tests: 58 passed
- full pytest suite: 918 passed, 2 skipped
- `loopx canary premerge --from-git-diff --tier standard`: 4 direct checks and 17 selected checks passed, 0 failures/warnings
- `loopx check`: 9 changed public files scanned, 0 errors/warnings

## Failures, skips, and holds
- the earlier head qualified two real Codex CLI calls; the current N=3 refinement was validated with the deterministic host and did not spend three additional model calls
- the two full-suite skips are existing environment-dependent skips
- no private state, raw prompts/transcripts, session ids, credentials, benchmark artifacts, generated lockfiles, or temporary paths are included
- hold merge for independent exact-head review; this PR is not self-merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -487,19 +487,23 @@ From a repository checkout with Codex CLI logged in, run this disposable,
 non-benchmark example:
 
 ```bash
-python3 examples/loopx-turn-codex-cli-e2e-smoke.py --real-codex-cli
+python3 examples/loopx-turn-codex-cli-e2e-smoke.py \
+  --real-codex-cli \
+  --two-turn-resume
 ```
 
-It creates a temporary goal and workspace, asks Codex CLI to write one marker,
-checks the marker with an independent validator, commits exactly one Turn and
-one quota spend, then replays the same transaction to prove it has no duplicate
-effects. The temporary state is deleted and is never synced to your global
-LoopX registry. Success reports `status=committed`,
-`validation_status=passed`, and `quota_slot_spend_count=1`.
+It creates a temporary goal and workspace, advances one marker through two
+separately validated Turns, and resumes the same opaque Codex CLI session for
+the second interaction. Each Turn commits once and spends once; the example
+then replays the second transaction to prove it has no duplicate effects. The
+temporary state is deleted and is never synced to your global LoopX registry.
+Success reports `status=committed`, `validation_status=passed`,
+`second_status=committed`, `second_validation_status=passed`,
+`session_resumed=true`, and `quota_slot_spend_count=2`.
 
 Run `codex --version` first. If the configured default model requires a newer
 Codex CLI, update Codex or add `--codex-model <qualified-model>`. The
-[one-Turn Codex CLI guide](docs/product/loopx-turn-codex-cli-quickstart.md)
+[Codex CLI Turn guide](docs/product/loopx-turn-codex-cli-quickstart.md)
 shows the command for a connected project and explains the validator boundary.
 
 ### Explore Graph And Harness (Supported, Optional, Default-Off)
--- a/README.md
+++ b/README.md
@@ -481,7 +481,7 @@ the run can claim progress. See
 [Auto-research command path](docs/guides/auto-research-command-path.md) for the
 full stop, attach, retry, and evidence boundary.
 
-### Experimental: Try One Governed Turn Locally
+### Experimental: Try Governed Turns Locally
 
 From a repository checkout with Codex CLI logged in, run this disposable,
 non-benchmark example:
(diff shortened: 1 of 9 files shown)


huangruiteng:
No blocking findings at exact head efd3e1bfd6a994278a9ad4e39b931822eca570c4.

Open question / assumption:
- I did not repeat the real-model Codex CLI run because that would consume external model compute. This review independently exercised the same Turn and Codex CLI adapter path with the deterministic fixture and relied on the PR public validation receipt plus green GitHub CI for the real-host qualification claim.

- Motivation: move the local experimental Turn entry from a one-transaction demo to a concrete continuation proof without making LoopX own the raw session.
- Solved: the opt-in path creates two separately validated and committed transactions, derives start_new then resume from the existing opaque session binding, spends once per transaction, and replays the second transaction with zero host/state/quota effects.
- Operator impact: users can now exercise continuation and idempotency in one disposable command while the session id remains outside the public result.
- Main risk: this is qualification/example code rather than new runtime semantics; keeping the command and README placement experimental is appropriate.
- Design: this reuses the existing Turn plan, session binding, validator, receipt, and journal contracts. It does not add a parallel state machine or expose the session handle.

Independent validation:
- GitHub Python Tests: passed.
- uv Python 3.12: pytest -q tests/test_loopx_turn_codex_cli.py -> 8 passed.
- Deterministic one-Turn and --two-turn-resume qualifications: passed; second_session_action=resume, quota_slot_spend_count=2, replay effects all false.
- quickstart smoke: passed.
- Ruff and changed-file py_compile: passed.
- loopx canary premerge --from-git-diff: 14/14 selected checks passed.
- git diff --check and public/private boundary scan: passed.
- No tracked reviewer changes; generated local validation artifacts are not part of the PR.

Decision: approved after independent reviewer; merge remains held for the authorized maintainer path.

huangruiteng:
Design finding at exact head `efd3e1bf`:

[P1] Keep the public qualification contract N-Turn rather than encoding a two-step special case. The production path is already generic because each transaction calls `turn run-once` and the opaque session binding selects `start_new` for the first transaction and `resume` thereafter. However, the qualification surface currently hard-codes `MARKER_VALUES[0/1]`, `--two-turn-resume`, `first_*`/`second_*` result fields, a one-off second branch, and quota count `1 | 2`. That makes the example users copy less general than the runtime it is meant to prove.

Requested refinement before merge:
- expose `--turn-count N` with a safe default of 1;
- generate/advance numbered marker states instead of `step-one`/`step-two`;
- execute one loop of separately validated and committed Turn transactions;
- report a bounded `turns[]` summary, asserting transaction 1 uses `start_new` and transactions 2..N use `resume`;
- require exactly N quota spends;
- replay the final transaction key and require zero effects;
- exercise at least N=3 deterministically so the smoke proves the control flow is not two-step-specific.

This should remain qualification/example code, not a new production N-turn orchestrator. The goal is to simplify the smoke around the already-generic `run-once` contract.

huangruiteng:
Refinement at exact head `a2ee513cd9465a9cb22ec3693c95a477200482b4` addresses the two-step hard-coding concern.

The important semantic distinction is now explicit:

- same semantic action + same `turn_instance_id` = the same logical Turn, so retry/replay keeps zero duplicate effects;
- same semantic action + a new `turn_instance_id` = a new logical Turn, while the Codex CLI session remains bound to the same goal/agent/todo and resumes opaquely;
- the experimental driver owns only the bounded `for 1..N` qualification loop; production remains a composable `turn run-once` primitive rather than gaining a second run-N scheduler.

Current-head evidence:

- deterministic `--turn-count 3`: `start_new, resume, resume`; every validator passed; exactly 3 durable commits and 3 quota spends; final journal replay had zero effects;
- default one-Turn path passed;
- focused tests: 58 passed;
- full suite: 918 passed, 2 existing environment skips;
- premerge: 4 direct + 17 selected checks passed, 0 failures/warnings/manual holds;
- public/private boundary: 9 changed files clean.

The earlier real Codex CLI two-call receipt is still useful session-resume evidence, but I am not reusing it as proof of the new N=3 head. No merge performed; exact-head review remains required.

PR #2381, opened 2026-07-20
fix(replan): keep user reminders nonblocking

huangruiteng:
## Summary
- derive frontier ownership from typed open `user_gate` items instead of every open user todo
- keep legacy or partial summaries fail-closed when typed gate projection is absent
- cover the combined non-blocking user action, stalled monitor, and blocked vision-successor path while preserving blocking gate precedence

## Root cause
`OPEN_USER_TODO` used the summary-wide `open_count`, so a `user_action` reminder marked non-blocking by the interaction contract preempted `MONITOR_NO_CHANGE_STREAK`. After two unchanged polls, quota still returned `monitor_quiet_skip` instead of requiring a bounded replan.

## Validation
- `62 passed`: goal-frontier rules, monitor replan, blocked vision successor, user-gate lane, public-safe replay, and decision-scope consistency
- `loopx canary premerge --from-git-diff`: 17/17 selected checks passed, 0 manual holds
- `git diff --check` passed

## Boundary
No private state, local paths, raw logs, credentials, benchmark task text, or generated artifacts are included.

### DIFF
--- a/loopx/control_plane/goals/goal_frontier.py
+++ b/loopx/control_plane/goals/goal_frontier.py
@@ -892,6 +892,20 @@ def _open_todo_count(summary: dict[str, Any] | None) -> int:
     return safe_non_negative_int(summary.get("open_count"))
 
 
+def _blocking_user_open_count(summary: dict[str, Any] | None) -> int:
+    if not isinstance(summary, dict):
+        return 0
+    gate_items = summary.get("gate_open_items")
+    if isinstance(gate_items, list):
+        return sum(
+            1
+            for item in gate_items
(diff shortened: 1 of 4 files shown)


huangruiteng:
Self-review qualification: no blocking findings. The change narrows frontier ownership to typed blocking user gates; non-blocking user actions remain visible through the user notification channel and cannot suppress a two-poll monitor replan. Legacy summaries without typed gate projection remain fail-closed. The integrated regression also includes a blocked vision successor and a blocking-gate control. Validation: 62 adjacent tests, Ruff, compile, and premerge 17/17 passed with no manual holds.

PR #2391, opened 2026-07-20
Add finance extension sample and starter scaffold

cocolord:
## Summary

- add `loopx-finance-value-discovery` as an optional standalone extension, without registering a Finance capability
- add `loopx extension init` starter scaffolding and managed `loopx extension run` for zero-permission standalone extensions
- preserve the public-safe PayPal de-beta method sample
- remove Finance execution from `value-connectors`; legacy selectors return migration tombstones
- keep effectful providers behind capability/domain commands with a minimal request-bound execution envelope
- terminate the entire provider process tree on timeout or stdout/stderr overflow

## Placement

- extension id: `loopx-finance-value-discovery`
- capability id: none
- provider: optional and extension-delivered
- runtime protocol: `finance_value_discovery_extension_v0`

Finance is independently installable and lifecycle-managed. LoopX core does not claim a provider-neutral Finance routing contract.

## Extension Management

`loopx extension init <extension-id>` previews by default and writes a six-file starter only with `--execute`. Generated extensions are independently buildable packages with a side-effect-free doctor and zero declared permissions.

`loopx extension run <extension-id>` resolves the enabled, doctor-ready active revision and provides bounded JSON stdin/stdout execution only for runtime-only standalone extensions. Direct run rejects:

- any `[[provides]]` or `[[implements]]` provider;
- any declared permission, including `external_write` and `openviking_context_write`;
- caller replacement of manifest-bound entrypoint, args, protocol, timeout, or permissions.

Capability providers remain reachable only through capability or domain commands that own policy checks, active-revision resolution, and managed dispatch.

## Minimal Execution Envelope

The earlier security-shaped authority decision was removed. Effectful dispatch now uses `loopx_extension_execution_envelope_v0`, containing only fields the provider independently revalidates before its first external write:

- exact operation/action;
- structured effect scope;
- extension id and resolved manifest revision;
- digest of the exact provider request, excluding the envelope.

Issuer, capability, protocol, permission, issue/expiry timestamps, and content-derived decision id are not duplicated in the envelope. Capability/protocol/permission remain authoritative in manifest and binding resolution. The envelope is request binding inside the same-host synchronous trust boundary, not issuer authentication or a security token.

The bundled OpenViking periodic-report provider is the reference path. Its capability command rejects caller-supplied envelopes, resolves the enabled doctor-ready revision, applies domain activation policy, creates the envelope, and dispatches through the bounded runtime. The provider rejects altered requests, wider scope, wrong action, wrong revision, and unknown fields before writes.

## Process-Tree Termination

Every managed provider starts in a dedicated POSIX session/process group. Timeout and stdout/stderr overflow send `SIGTERM` to the group, wait for the bounded grace period, then send `SIGKILL` to any remaining group. Windows uses a new process group plus `taskkill /T` and forced tree termination after the grace period.

Regression tests start real descendants that delay and then write effect markers. Shared-helper timeout/overflow tests and public `extension run` timeout/stdout/stderr tests wait beyond the intended write time and prove no marker appears after LoopX reports failure.

## Compatibility

- extension lifecycle state and existing lifecycle commands remain compatible
- `extension init`, runtime-only manifests, and standalone `extension run` are additive
- legacy `finance_market_snapshot` selectors remain self-describing migration tombstones
- agent-driven upgrades receive ordered inspect, install, register, and run steps
- machine clients hard-coding old Finance mapping values must adopt `value_connector_extension_migration_v0`
- LoopX-wheel-only users still need an explicit provider source; this PR does not silently install executable code

## Validation

- full suite: `1055 passed, 2 skipped`
- complete extension + registry suite: `91 passed`
- focused envelope/runtime/OpenViking suite: `53 passed`; the additional shared-helper overflow regression also passed in the complete suite
- Ruff, Python compile, and `git diff --check`: passed
- standard premerge: 18 selected checks passed, 0 failures, 0 warnings, 0 manual holds
- premerge covered local/package install, packaged CLI, update, extension registry, placement, Python compile, and public/private boundary checks

## Boundaries

No network collection, investment advice, price target, trade execution, account or portfolio access, continuous watch, credentials, private paths, or raw provider payloads are introduced. The PayPal packet is historical method evidence, not a current company view.


cocolord:
Validation note after fork publication:

- Changed surfaces: optional extension lifecycle/provider, value-connectors compatibility dispatch and source-map, public method docs/sample, focused semantic/runtime tests.
- Focused checks: 38 passed; both finance extension and existing value-connectors smokes passed; temporary package install, doctor, and installed sample execution passed.
- Risk gate: `loopx canary premerge --from-git-diff` passed 18/18 checks, including Python compile and public/private boundary; 0 failures, 0 warnings, 0 manual holds.
- Full-suite residual: 362 tests passed with no failures before the unrelated existing scheduler host-binding test stalled; that run was interrupted after 4m29s.
- Coverage is sufficient for draft review because the changed lifecycle, catalog, compatibility, semantic reducer, package install, and public-boundary paths are each exercised directly. No merge is requested from this comment.

cocolord:
Owner correction applied in `f75d80fe`:

- moved the co-located provider from `extensions/finance-value-discovery/` to `extensions/loopx-finance-value-discovery/`; the directory now matches manifest id `loopx-finance-value-discovery`
- updated install, example, smoke, and test paths
- added a generic regression requiring every top-level co-located extension directory to equal its manifest provider id
- reran 39 focused tests, both smokes, clean package install/doctor/sample execution, and 18/18 premerge checks with public-boundary pass

This corrects extension delivery identity only; reducer behavior and the non-advice PayPal result are unchanged.

cocolord:
Owner correction applied in `15aeec4d`: the extension no longer creates or implements a `finance-value-discovery` capability. Its manifest is runtime-only, the compatibility facade resolves `loopx-finance-value-discovery` directly, and focused tests assert that loading the manifest leaves the capability catalog unchanged.

Local validation: 44 focused tests; three extension/placement smokes; isolated install + doctor + PayPal readback; premerge 9/9 catalog canaries, 8/8 risk smokes, public boundary clean, 0 manual holds.

cocolord:
Post-rebase validation at `fbe2a31a`: this supersedes earlier comments that referenced a `value-connectors` Finance compatibility delegate. The final branch keeps `loopx-finance-value-discovery` standalone: it registers no capability and exposes no `value-connectors` Finance route. Validation after rebasing onto `3f9895e8`: 29 focused tests passed; three focused smokes passed; CI-equivalent Ruff passed; `loopx canary premerge --from-git-diff` passed 18/18 with 0 failures and 0 manual holds; GitHub `pytest` passed.

huangruiteng (changes requested):
## Blocking findings

1. **[P1] Enforce operation authority before invoking a standalone provider.** The new generic `extension run --execute` path goes directly from the CLI into `run_standalone_extension` ([CLI](https://github.com/huangruiteng/loopx/blob/3d77f1c46b11fcb2ff3bc5a1fd1c2dbd643e5a01/loopx/cli_commands/extension.py#L139-L150), [runtime](https://github.com/huangruiteng/loopx/blob/3d77f1c46b11fcb2ff3bc5a1fd1c2dbd643e5a01/loopx/extensions/runtime.py#L735-L769)). The only permission decision is whether runtime-required strings appear in the manifest; no goal boundary or effect authorization is supplied or checked. A minimal standalone provider declaring and requiring `external_write` can therefore create an external-effect marker and return `succeeded`. That contradicts the documented contract that declaration does not grant authority and existing goal/user/external-write gates still decide execution ([docs](https://github.com/huangruiteng/loopx/blob/3d77f1c46b11fcb2ff3bc5a1fd1c2dbd643e5a01/docs/reference/extensions.md#L355-L358)). Please either thread an explicit typed authority decision into the generic runner and fail closed, or reject effectful permissions here and route them through a domain command that owns those gates. Add a regression proving declaration alone cannot authorize an effectful run.

2. **[P1] Enforce the advertised byte limits while reading/running, not after unbounded work.** `_load_json_object` reads all stdin/file content before the 1 MB request check ([CLI](https://github.com/huangruiteng/loopx/blob/3d77f1c46b11fcb2ff3bc5a1fd1c2dbd643e5a01/loopx/cli_commands/extension.py#L153-L165)), while provider stdout/stderr are written to unlimited temporary files and measured only after the subprocess exits ([runtime](https://github.com/huangruiteng/loopx/blob/3d77f1c46b11fcb2ff3bc5a1fd1c2dbd643e5a01/loopx/extensions/runtime.py#L799-L834)). In a minimal repro, a provider wrote 2 MB, flushed, then executed code after the nominal cap; LoopX only returned `response_too_large` after completion. A buggy provider can therefore consume arbitrary memory/disk until timeout even though the public contract says requests and responses are bounded. Please cap reads before JSON parsing, stream/cap stdout and stderr, terminate the provider as soon as either cap is crossed, and cover stdin, file input, stdout, and stderr limits.

Validation performed: `git diff --check` passed; LoopX public-boundary scan passed with 0 errors/warnings; 42 focused extension/finance tests passed. One generated-starter E2E could not complete locally because the local Python runtime failed while creating its child venv; the GitHub pytest check is green.

cocolord:
Addressed both requested blockers in `be007e3c` / `ef59bb1d` / `1514d1e7`:

- Authority: generic standalone execution now rejects the authority-bound `external_write` permission before provider invocation and directs effectful operations to a domain command with explicit goal/external-write gates.
- Byte limits: file and stdin requests are capped before JSON parsing; provider stdout and stderr are drained concurrently, and the provider is terminated as soon as either stream exceeds 1 MB. Timeout termination remains fail-closed.
- Regression coverage: declaration alone cannot invoke an effectful provider; oversized file/stdin input is rejected before runtime resolution; stdout/stderr overflow and timeout terminate before the provider writes its completion marker.

Validation on the pushed head: 50 focused extension tests passed; finance-value-discovery, capability-placement, and value-connectors-finance smokes passed; Ruff and diff checks passed; `loopx canary premerge --from-git-diff` passed 18/18 with 0 failures and 0 manual holds; an explicit five-file public/private boundary readback passed with 0 errors/warnings. No checks were skipped. This covers the changed CLI/runtime path, all four requested byte-limit surfaces, timeout preservation, public docs, and the public/private boundary.

GitHub pytest is queued. I am leaving the PR unmerged for maintainer re-review and approval.

cocolord:
@huangruiteng Both blocking findings from the change-request review are addressed on the rebased head `3c16c12d`.

1. **Operation authority:** generic `extension run` now rejects a provider that merely declares the authority-bound `external_write` permission, before provider invocation. The regression uses an invocation marker and proves declaration alone cannot execute the provider. Effectful providers must enter through a domain command with explicit goal/user/external-write gates.
2. **In-flight byte limits:** file and stdin reads stop at `MAX_EXTENSION_REQUEST_BYTES + 1` before JSON parsing. Provider stdout/stderr are concurrently drained into capped buffers; crossing either cap terminates the process immediately. Regressions cover file input, stdin, stdout, stderr, timeout, and post-cap marker non-execution.

Validation on the exact staged tree:

- `1019 passed, 2 skipped` full pytest
- `61 passed` extension suite; `49 passed` focused authority/runtime/starter/Finance set
- 20 PR Python files pass Ruff lint and format
- built LoopX wheel -> generated six-file starter -> extension wheel -> install/doctor/run/readback: passed
- Finance wheel -> install/doctor/run/PayPal readback: passed
- premerge: 4 direct + 18 selected checks, 0 failures/warnings/holds
- public boundary: 10/10 staged files scanned, 0 hits, 0 skips

Please re-review the two previously blocking surfaces. I have not merged or dismissed the existing review state.

cocolord:
已按 owner 方向统一 Extension 管理入口：`loopx extension run` 现在接受 standalone、`[[provides]]` 和 `[[implements]]` Extension；capability/domain command 仅在需要额外策略或权限时作为上层入口，并复用同一托管 runtime。新增真实 bundled OpenViking provider 的 install/doctor/run/readback 集成覆盖，证明这不是 Finance 特例。已 rebase 最新 main；本地验证：full suite 1021 passed / 2 skipped，extension suite 62 passed，standard premerge 0 failures / 0 warnings / 0 holds。请基于新 head `3b262ea8` 重新 review。

cocolord:
纠正上一版 direct-run 边界：统一的是 Extension 生命周期管理，不是所有 Provider 的直接执行。`loopx extension run` 现在仅接受 runtime-only standalone Extension；`[[provides]]` 与 `[[implements]]` Provider 必须经 capability/domain command。新增真实 bundled Lark install/doctor 后 direct-run 拒绝测试，并保留 Finance standalone managed-run 覆盖。验证：focused 50 passed，extensions 62 passed，full suite 1021 passed / 2 skipped，standard premerge 0 failures / 0 warnings / 0 holds。请基于新 head `c7229e84` review。

cocolord:
补齐截图 review 的两个 blocker，见 `73495f2c`：\n\n1. **Authority**：standalone direct-run 现在要求 manifest `permissions=[]` 且 runtime `required_permissions=[]`。任何声明权限都会在 Provider 启动前拒绝；回归分别使用 `external_write` 与 `openviking_context_write`，并验证 effect marker 未创建。`[[provides]]` / `[[implements]]` 仍只能经 capability/domain command。Finance 与 starter 已改为零权限纯执行。\n2. **Legacy plan migration**：`value-connectors plan --connector-id finance_market_snapshot` 不再生成普通 plan，直接返回顶层 `value_connector_extension_migration_v0`，与 `source-map` / `install-check` 使用同一 inspect-install-register-run 契约。\n\n验证：focused 52 passed，extensions 64 passed，full suite 1023 passed / 2 skipped，standard premerge 0 failures / 0 warnings / 0 holds，public boundary passed。请基于 head `73495f2c` 重新 review。

cocolord:
Addressed the authority blocker on head `7b26dfbf`.

The provider no longer treats raw `available_capabilities` as operation authority. Effectful dispatch now follows this path:

1. the capability/domain command resolves an enabled, doctor-ready active extension revision;
2. it issues `loopx_extension_authority_decision_v0`, bound to capability, protocol, permission, action, structured scope, exact request digest, extension id/revision, and a maximum five-minute lifetime;
3. the decision is validated before bounded subprocess dispatch;
4. the provider validates the same decision and bound runtime identity again before any write.

The validator uses an exact field set and fails closed for caller-supplied, expired, tampered, replayed-against-another-request, wider-scope, wrong-action, wrong-revision, or unknown-field decisions. Public `extension run` continues to reject `[[provides]]`, `[[implements]]`, and every permissioned extension, so the capability path cannot be bypassed.

Validation:
- `1040 passed, 2 skipped` full suite
- `80 passed` extension suite
- `54 passed` focused authority/OpenViking/runtime suite
- Ruff and diff checks passed
- standard premerge passed all 18 selected checks with no failures, warnings, or manual holds

Trust boundary: this is a same-host LoopX control-plane decision and bounded executor, not an OS sandbox or a substitute for OpenViking/service credentials.

huangruiteng:
Maintainer direction on latest head `7b26dfbf`: **please substantially rework this before the next review**.

The latest commit addresses the earlier raw-capability-string concern in a thoughtful way, and the focused suite is green. However, it introduces a much larger authority protocol than the current same-host, synchronous execution path needs, while the subprocess boundary still permits effects to continue after LoopX reports a timeout. Please simplify the control contract first rather than adding more fields and tests around the current design.

## 1. Reduce the authority design to the enforceable contract

`loopx_extension_authority_decision_v0` currently carries issuer, capability, protocol, permission, action, scope, extension identity/revision, request digest, issue/expiry timestamps, and a content-derived decision id. This is a lot of protocol and validation surface for one synchronous in-process-to-subprocess handoff.

The important distinction is:

- binding an invocation to the exact request/revision/scope is useful and enforceable;
- proving that an authorized issuer created it is not provided by a content-derived hash;
- a five-minute expiry has little demonstrated value when the command immediately launches the provider;
- duplicating capability/protocol/permission fields already checked by manifest resolution creates multiple sources of truth unless the provider independently needs each field.

The current docs correctly say this is not an OS sandbox. Within that stated trust boundary, please model the mechanism as a **minimal typed execution envelope**, not a security-token-like authority object.

Suggested target:

1. The capability/domain command remains the only entrypoint for an effectful provider. It owns policy checks, active-revision resolution, and managed dispatch.
2. Public `extension run` continues to reject every permissioned extension and every `[[provides]]` / `[[implements]]` provider before invocation.
3. The dispatcher creates the smallest envelope the provider must revalidate before effects. Likely fields are:
   - schema version;
   - operation/action;
   - structured effect scope;
   - extension id and resolved revision;
   - digest of the exact provider request without the envelope.
4. Keep capability/protocol/permission in one authoritative place. Include them in the envelope only when the provider has a concrete independent validation rule for them.
5. Remove issuer, issued-at, expiry, and content-derived decision id unless there is a documented asynchronous forwarding/replay threat that needs them. If the intent is genuine issuer authenticity, specify and implement an actual authenticity mechanism; an unkeyed content hash must not be presented as one.
6. Keep provider-side request/revision/scope validation before the first external write, with negative tests for altered request, wider scope, wrong action, and wrong revision.

Please also reduce the PR's scope. A clean decomposition would be:

- this PR: Finance standalone extension, migration tombstones, starter scaffold, zero-permission managed runtime, and the runtime hardening required by that path;
- a separate authority-envelope PR only if a reusable effectful-provider contract is still required after simplifying the threat model.

It is also acceptable to keep the minimal envelope here if the resulting diff remains cohesive. What should not remain is a generic security-shaped framework whose extra fields do not change what the same-host runtime can enforce.

## 2. Fix the process-tree termination blocker

[`run_capped_process`](https://github.com/huangruiteng/loopx/blob/7b26dfbf23ddc9c95984379cd96bf2e276c120c1/loopx/extensions/process_runtime.py#L22-L37) still calls `terminate()` / `kill()` only on the direct `Popen` child. The provider is not placed in a dedicated process group at [process creation](https://github.com/huangruiteng/loopx/blob/7b26dfbf23ddc9c95984379cd96bf2e276c120c1/loopx/extensions/process_runtime.py#L50-L57).

I reproduced this on the exact latest head:

1. the provider starts a child that waits and then writes an effect marker;
2. the provider itself sleeps beyond the one-second LoopX timeout;
3. LoopX returns `failure_kind=timeout`, and the provider exits with `-15`;
4. the descendant remains alive and writes the marker after LoopX has already reported failure.

That violates the runtime contract: an operator can observe a failed/stopped extension while its side effect is still executing in the background. The same helper is used for stdout/stderr overflow termination, so the gap applies there as well.

On POSIX, the expected shape is:

- start each provider in a new session/process group (`start_new_session=True`);
- on timeout or either output-limit event, send `SIGTERM` to the provider's process group;
- wait for the bounded grace period;
- send `SIGKILL` to the group if any part remains;
- tolerate normal exit and `ProcessLookupError` races without targeting unrelated processes.

Use the repository's supported-platform policy for Windows; if Windows is supported here, use an equivalent process-tree primitive rather than silently reverting to direct-child-only cleanup.

Add a regression that starts a real descendant with a delayed marker, triggers timeout, waits beyond the descendant's intended write time, and proves the marker never appears. Cover the shared helper plus at least one public runtime path. Add equivalent descendant coverage for output overflow, because both abort paths promise that execution has stopped. Ensure failed tests clean up any leaked process.

The existing timeout test only sleeps and writes in the direct provider process, so it cannot catch this class of leak.

## Validation observed

- GitHub `pytest`: passed on `7b26dfbf`.
- Focused authority/runtime/OpenViking suite: `54 passed`.
- `git diff --check`: passed.
- Independent descendant-process reproduction: failed as described above; a post-timeout marker was written.

## Merge decision

**Request changes / hold.** Please simplify the authority layer and fix process-tree termination before asking for another merge review. This should be a design reset, not another additive patch on top of the current decision schema.

cocolord:
Addressed the design-reset and process-tree blockers on rebased head `b298e4c2`.

### 1. Authority protocol reduced to an enforceable execution envelope

The previous `loopx_extension_authority_decision_v0` implementation is removed. The replacement `loopx_extension_execution_envelope_v0` contains only:

- action;
- structured effect scope;
- extension id and resolved revision;
- digest of the exact provider request without the envelope.

Issuer, capability, protocol, permission, issued/expiry timestamps, and content-derived decision id were deleted. Capability/protocol/permission remain single-source in manifest/binding resolution. Docs now explicitly state that the envelope provides request/revision/scope binding, not issuer authenticity or token security.

Public `extension run` still rejects every permissioned, `[[provides]]`, and `[[implements]]` extension before invocation. OpenViking remains behind the periodic-report domain command. Provider-side negative tests cover altered request, wider scope, wrong action, wrong revision, and unknown fields before any write.

### 2. Provider descendants are terminated

`run_capped_process` now starts each POSIX provider with `start_new_session=True`. Timeout or stdout/stderr overflow sends `SIGTERM` to the process group, waits the bounded grace period, then sends `SIGKILL` to the remaining group. The Windows path uses a new process group and `taskkill /T`, with forced tree termination after the grace period.

New regressions launch real descendants with delayed effect markers and wait beyond their intended write time. Coverage includes:

- shared helper: timeout and stdout overflow;
- public `extension run`: timeout, stdout overflow, and stderr overflow.

No marker appears after LoopX reports failure.

### Validation

- full pytest: `1055 passed, 2 skipped`
- extension + capability registry: `91 passed`
- focused envelope/runtime/OpenViking set: `53 passed` before the additional helper-overflow case; that case passed in the complete suite
- standard premerge: 18/18 selected checks, 0 failures, warnings, or manual holds
- Ruff, compile, diff hygiene, and public/private boundary: passed

I have not dismissed the existing change request or merged the PR. Please re-review the two blocking surfaces on `b298e4c2`.

huangruiteng (approved):
Approved on head `b298e4c2`.

The two blocking surfaces are resolved:

- The security-shaped authority decision was removed. The replacement execution envelope contains only the action, effect scope, bound extension id/revision, and exact request digest; it explicitly does not claim issuer authentication.
- Managed provider aborts now terminate the POSIX process group on timeout and stdout/stderr overflow. Independent descendant-marker reproductions no longer produce post-failure effects.

Validation:

- GitHub pytest: passed
- focused envelope/runtime/OpenViking/process suite: 54 passed
- standard premerge: 18/18 selected checks, 0 failures, warnings, or holds
- public/private boundary: clean across 38 changed files
- local extension and registry suite: 91 passed; one additional generated-starter test was environment-blocked by the local Python 3.13 ensurepip/dyld setup, while GitHub's full pytest passed

Architecture judgment: Finance is correctly standalone and does not create a core capability; zero-permission direct run and capability-owned effectful dispatch now have a coherent boundary.

Merge decision: approved.

PR #2405, opened 2026-07-20
feat(turn): retain material replan path deltas

### LINKED ISSUE #2353: A user-side idea: a lightweight “look back and retain” step for bounded agent loops

huangruiteng:
## Summary
- classify LoopX Turn results as routine unchanged continuation or a material replan
- require material replans to carry a bounded, agent-scoped `goal_path_delta_v0`, then write that packet through the existing vision checkpoint boundary
- keep routine continuation light with the existing `vision_unchanged_reason` path
- add one real `turn run-once` acceptance smoke that proves material retention and routine no-delta behavior

## Validation
- `86 passed` across Turn executor/Codex CLI/driver and goal-vision succession suites
- `python examples/loopx-turn-path-delta-acceptance-smoke.py`
- focused Ruff passed
- premerge canary passed: 4 direct checks, 4 catalog checks, 8 risk-profile checks, public boundary clean, no warnings or manual holds

## Boundary
No heartbeat expansion, new planner, raw host evidence, private state, credentials, or benchmark execution. Existing generic hosts that only provide `vision_unchanged_reason` retain compatibility.

Fixes #2353

### DIFF
--- a/examples/loopx-turn-path-delta-acceptance-smoke.py
+++ b/examples/loopx-turn-path-delta-acceptance-smoke.py
@@ -1,0 +1,272 @@
+#!/usr/bin/env python3
+"""Prove material Turn replans retain a path delta while routine replans stay light."""
+
+from __future__ import annotations
+
+import contextlib
+import io
+import json
+import sys
+import tempfile
+from pathlib import Path
+from typing import Any
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(REPO_ROOT))
+
+from loopx.cli import main as cli_main  # noqa: E402
+
+
+GOAL_ID = "loopx-turn-path-delta-fixture"
+AGENT_ID = "codex-path-delta-fixture"
+TODO_ID = "todo_pathdelta001"
+MARKER = "docs/path-delta-route.txt"
+
+
+def _write_fixture(root: Path) -> tuple[Path, Path, Path, Path]:
+    project = root / "project"
+    runtime = root / "runtime"
+    workspace = root / "workspace"
+    runtime.mkdir(parents=True)
+    (workspace / "docs").mkdir(parents=True)
+
+    state = project / ".codex" / "goals" / GOAL_ID / "ACTIVE_GOAL_STATE.md"
+    state.parent.mkdir(parents=True)
+    state.write_text(
+        "\n".join(
+            [
+                "---",
+                "status: active",
+                "updated_at: 2026-01-01T00:00:00+00:00",
+                "---",
+                "",
+                "# LoopX Turn Path Delta Fixture",
+                "",
+                "## Next Action",
+                "",
+                "Validate the original fixture route.",
+                "",
+                "## Agent Todo",
+                "",
+                "- [ ] [P0] Validate one material route change and one routine continuation.",
+                (
+                    f"  <!-- loopx:todo todo_id={TODO_ID} status=open "
+                    "task_class=advancement_task action_kind=path_delta_acceptance "
+                    f"claimed_by={AGENT_ID} priority=P0 -->"
+                ),
+                "",
+            ]
+        ),
+        encoding="utf-8",
+    )
+    registry = project / ".loopx" / "registry.json"
+    registry.parent.mkdir(parents=True)
+    registry.write_text(
+        json.dumps(
+            {
+                "schema_version": 1,
+                "common_runtime_root": str(runtime),
+                "goals": [
+                    {
+                        "id": GOAL_ID,
+                        "domain": "loopx-turn-public-fixture",
+                        "status": "active",
+                        "repo": str(project),
+                        "state_file": str(state.relative_to(project)),
+                        "adapter": {"kind": "fixture_v0", "status": "connected-delivery"},
+                        "quota": {"compute": 10.0, "window_hours": 24},
+                        "coordination": {
+                            "agent_model": "peer_v1",
+                            "registered_agents": [AGENT_ID],
+                            "agent_profiles": {
+                                AGENT_ID: {
+                                    "schema_version": "agent_profile_v1",
+                                    "profile_role": "fixture",
+                                    "scope": "public qualification",
+                                }
+                            },
+                            "write_scope": ["docs/**"],
+                        },
+                    }
+                ],
+            },
+            indent=2,
+            sort_keys=True,
+        )
+        + "\n",
+        encoding="utf-8",
+    )
+    return project, runtime, workspace, registry
+
+
+def _write_host(root: Path) -> Path:
+    host = root / "path-delta-host.py"
+    host.write_text(
+        f"""
+import json
+import pathlib
+import sys
+
+request = json.load(sys.stdin)
+mode = sys.argv[1]
+marker = pathlib.Path({MARKER!r})
+marker.write_text(mode, encoding="utf-8")
+material_packet = {{
+    "agent_id": {AGENT_ID!r},
+    "state": "active",
+    "vision_patch": {{
+        "direction": "Validate the revised public fixture route.",
+        "acceptance_summary": "The revised route and its routine continuation both pass independent validation.",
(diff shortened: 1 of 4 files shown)


huangruiteng:
## Self-review and qualification

**Findings:** No blocking correctness or boundary findings in the final four-file diff. `material_replan` is accepted only for `replan_required`, requires a normalized agent-scoped vision packet with `goal_path_delta_v0` and `outcome=replan`, and rejects contradictory `vision_unchanged_reason`. Routine results keep the existing unchanged-reason path; older generic hosts that omit the two new fields are inferred as unchanged for compatibility.

**Changed surfaces:** LoopX Turn host result validation, Codex CLI structured result schema/prompt, `turn run-once` refresh-state writeback, and one public acceptance smoke.

**Checks:** 86 focused pytest cases passed; the new material/routine real-callsite smoke passed; focused Ruff passed; exact-head standard premerge canary passed 4 direct, 4 catalog, and 8 risk-profile checks with the public boundary clean, 0 failures, 0 warnings, and 0 manual holds.

**Coverage judgment:** The existing suites cover transaction validation, host compatibility, resumable writeback, and vision succession. The new smoke exercises the missing production callsite end to end and asserts both sides of the policy: material replans retain the delta, routine replans do not emit a new vision packet.

**Merge posture:** This changes runtime Turn result semantics, so despite green canary self-merge eligibility it should receive independent review before merge.

huangruiteng (commented):
## Independent review

**Findings:** No blocking correctness, compatibility, or public-boundary findings at exact head `df1ba8ad6f4f3cd72ccca066ca5216edd3f52e97`.

**Product and architecture judgment:** This closes a real Turn provenance gap. A host could previously return `replan_required`, but the committed Turn record retained only an unchanged-vision reason and therefore lost the material route change that justified the replan. The patch keeps routine continuation light while making a material replan explicit, agent-scoped, bounded, and durable through the existing goal-vision checkpoint. It reuses the canonical vision normalizer instead of introducing a parallel planner or state sink.

**Changed surfaces:** Turn host-result validation; Codex CLI structured result schema and prompt; `turn run-once` refresh-state writeback; one real-callsite material/routine acceptance smoke.

**Correctness checks:** Material deltas are limited to `replan_required`, must normalize against the Turn envelope's `goal_id` and `agent_id`, require `goal_path_delta_v0` semantics with `outcome=replan`, and cannot coexist with `vision_unchanged_reason`. Routine material results require an unchanged reason and reject an embedded vision packet. Existing generic hosts that omit the new fields still infer the prior unchanged contract.

**Validation:** Reproduced the stated 86 focused pytest cases across Turn executor/Codex CLI/driver and goal-vision succession; the new `loopx-turn-path-delta-acceptance-smoke.py` passed; focused Ruff passed. Exact-diff premerge canary passed 4 direct checks, 4 catalog checks, 8 risk-profile checks, and the public-boundary scan with 0 failures, 0 warnings, 0 skips, and 0 manual holds. GitHub `pytest` is also green.

**Residual risk:** Whether a model classifies a borderline route change as material remains model-dependent, but malformed or contradictory output cannot reach writeback. That is an acceptable residual behavior-quality risk for this deterministic runtime contract.

**Merge decision:** Ready to merge after this independent review.

PR #2420, opened 2026-07-21
Complete standalone extension init contract

cocolord:
## Summary

- generate versioned request and response JSON Schemas from `loopx extension init`
- make the starter provider fail closed on missing, mismatched, or unsupported request fields
- expose `starter_kind=standalone` and the managed `loopx extension run` entrypoint in the init receipt
- extend lifecycle coverage through install, list/readback, doctor, disable/enable, and managed invocation
- document why `provides` and `implements` require capability-specific integration profiles instead of inferred generic authority

## Validation

- extension and capability-extension registry suite: 93 passed
- scaffold focused suite: 10 passed
- Ruff: passed
- `loopx canary premerge --from-git-diff`: 10/10 selected checks passed, 0 failures, warnings, or holds

## Scope

This completes the safe standalone onboarding path. It intentionally does not generate `[[provides]]` or `[[implements]]` providers until a real capability integration profile defines the caller contract, resolver, policy checks, action/scope mapping, and execution-envelope adapter.


huangruiteng:
Maintainer follow-up: I fixed the schema contract on the contributor branch and re-reviewed the PR at `b95c38a0` against current `main`.

No remaining PR-local blocker found.

### Fixes applied

- The request schema now rejects whitespace-only `message` values, matching the provider's `.strip()` behavior.
- The response schema is now a strict success/failure `oneOf` contract:
  - success requires `ok: true`, `request_schema_version`, and `result`;
  - failure requires `ok: false` and `error`;
  - success-only and failure-only fields cannot be mixed.
- Scaffold tests now validate actual provider success and failure output with the generated Draft 2020-12 schemas, plus the previously false-valid edge cases.
- `jsonschema` is declared in the test extra used by those tests.

### Validation

- `54 passed, 1 deselected` across scaffold/runtime/co-located/finance extension tests. The one local deselection is the nested-venv lifecycle test because the uv-managed macOS Python cannot load its copied `libpython` dylib during `venv.EnvBuilder`; this occurs before LoopX code executes.
- Independently exercised the generated wheel through `extension install`, `list`, `doctor`, `disable`, `enable`, and managed `run`; the success response validates and the false-valid request/response cases are rejected.
- Ruff, `py_compile`, and `git diff --check` passed.
- `loopx canary premerge --from-git-diff`: `merge_gate_passed=true`, `self_merge_allowed=true`, 8/8 risk-profile smokes passed, public-boundary scan passed, no manual holds.
- GitHub Actions executed the complete suite: `1073 passed, 2 skipped`. Its sole failure is the inherited maintainability ratchet for `loopx.control_plane.quota.goal_boundary:goal_boundary` (`73 > 66` decision points, `107 > 96` statements). This PR does not change that file; current `main` at `9820e937` fails the same check after #2437, and the canary classifies it as `advisory_inherited_failure`.

### Product/architecture judgment

The PR now delivers the intended standalone-extension contract without inventing a capability: generated packages have a lifecycle-valid managed entrypoint and schemas that agree with runtime behavior in both directions. The change remains narrowly scoped to scaffold generation, its public documentation, and contract tests.

With the owner's explicit authorization, I am admin-merging this PR despite the unrelated base-branch ratchet failure. The inherited `goal_boundary` maintainability regression should be repaired separately rather than bundled into this extension PR.

PR #2425, opened 2026-07-21
Fix missing benchmark score countability

huangruiteng:
## Summary

- honor explicit `official_score_attempt_countable=false` before using a bool-only score fallback
- only let an allowed bool fallback override ledger attempt countability
- cover both the missing-score regression and the valid completed bool-only zero fallback

## Motivation

A compact benchmark result can contain `official_task_score.passed=false` while the official score is missing because verifier infrastructure did not produce a countable score. The ledger previously converted that shape to a countable `0.0`, which could pollute aggregates and matched comparisons.

This change keeps the official benchmark result untouched and tightens only the ledger ingestion contract.

## Validation

- `PYTHONPATH=. uv run --no-project --with pytest python -m pytest -q tests/test_benchmark_ledger_countability.py` (2 passed)
- `PYTHONPATH=. uv run --no-project --with pytest python examples/benchmark-run-ledger-smoke.py` (passed)
- `loopx canary premerge --from-git-diff --format json` (compile and 6 selected canaries passed; benchmark-sensitive manual hold remains)
- public/private boundary scan clean; no raw task text, logs, trajectories, verifier output, credentials, or local paths added

## Risk

Low and localized to benchmark ledger ingestion. Completed countable bool-only scores retain their existing fallback behavior; explicitly uncountable missing scores remain missing and cannot enter aggregate scoring.

### DIFF
--- a/loopx/benchmark_ledger.py
+++ b/loopx/benchmark_ledger.py
@@ -2158,7 +2158,7 @@ def build_benchmark_run_ledger_entry(
             if isinstance(marker.get("attempt_accounting"), dict)
             else {}
         )
-    has_official_bool_score = official_score_bool_fallback_used(benchmark_run)
+    has_official_bool_score = bool_fallback_used
     if attempt_accounting:
         for source_field, entry_field in (
             ("lifecycle_phase", "attempt_lifecycle_phase"),
--- a/loopx/benchmark_ledger_countability.py
+++ b/loopx/benchmark_ledger_countability.py
@@ -126,6 +126,17 @@ def official_score_bool_fallback_allowed(run: dict[str, Any]) -> bool:
     if not _fallback_candidate_present(run):
         return False
 
+    accounting = (
+        run.get("attempt_accounting")
+        if isinstance(run.get("attempt_accounting"), dict)
+        else {}
+    )
+    attempt_countable = run.get("official_score_attempt_countable")
+    if attempt_countable is None:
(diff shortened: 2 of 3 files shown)


huangruiteng:
No actionable findings after reviewing the three-file diff.

- Motivation: prevent verifier-infrastructure score absence from becoming a synthetic official zero in the benchmark ledger.
- Solved: explicit attempt countability is now authoritative, while completed countable bool-only results preserve compatibility.
- Operator impact: matched comparisons and aggregates can no longer silently treat a missing score as model failure.
- Main risk: overly broad fallback suppression. The paired regression tests cover both sides of that boundary.
- Design: the rule stays in the existing countability module and reuses the already-computed allowed-fallback decision in entry construction; no new adapter abstraction is introduced.

Validation:
- focused pytest: 2 passed
- full benchmark run-ledger smoke: passed
- premerge canary: changed-file compile plus 6 selected maintainability, adapter, artifact-boundary, ledger, and case-analysis checks passed
- failures/skips: the first manual smoke invocation used unsupported system Python 3.9 and was rerun successfully under the controlled uv Python; no product test failure remains
- boundary: no raw task text, logs, trajectories, verifier output, credentials, local paths, uploads, or submissions

Manual hold: premerge correctly classifies this as benchmark-sensitive. Owner authorization for benchmark helper/runtime self-merge is active; the change does not alter verifier scoring, task semantics, permissions, or leaderboard behavior. Pending GitHub pytest, this is suitable for admin self-merge.

PR #2441, opened 2026-07-22
Validate SkillsBench Turn content changes

huangruiteng:
## Summary
- compare durable task-file content before each SkillsBench bridge write
- commit Turn progress only for a successful new or changed durable task file
- keep private paths and contents out of persisted receipts; expose bounded counts only
- preserve task-declared independent postconditions as the primary validation contract

## Why
The earlier fallback accepted any successful durable write, so rewriting identical content could incorrectly commit a Turn. This makes the fallback content-aware while failing closed when the private pre-write probe cannot establish a change.

## Validation
- `uv run pytest -q tests/test_skillsbench_turn_runtime.py tests/test_skillsbench_turn_launcher.py tests/test_skillsbench_docker_command_file_bridge.py` (44 passed)
- `uv run ruff check ...` on the five changed files (passed)
- `uv run python examples/skillsbench-benchmark-run-smoke.py` (passed)
- `uv run python examples/skillsbench-agent-operation-snapshot-smoke.py` (passed)
- `uv run loopx canary premerge --from-git-diff` (6/6 catalog canaries passed; manual `benchmark_sensitive` hold remains)
- `ruff format --check` reports the same three pre-existing files on exact `origin/main`; no broad formatting churn included
- `examples/skillsbench-loopx-turn-agent-cli-smoke.py` remains a known pre-existing failure on exact main (`SkillsBenchTurnBridgeError` during turn planning)

## Boundaries
- does not run, upload, score, or submit a benchmark
- does not consume official verifier feedback
- does not persist raw request content, task paths, task contents, logs, trajectories, credentials, or host paths
- requires explicit maintainer review because the premerge canary classifies this as benchmark-sensitive

### DIFF
--- a/examples/skillsbench-loopx-turn-agent-cli-smoke.py
+++ b/examples/skillsbench-loopx-turn-agent-cli-smoke.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import json
 import shlex
+import subprocess
 import sys
 import tempfile
 from pathlib import Path
@@ -19,7 +20,15 @@ from loopx.benchmark_adapters.skillsbench import skillsbench_route_contract  # n
 from loopx.benchmark_adapters.skillsbench_acp_failure_policy import (  # noqa: E402
     recoverable_codex_turn_failure_message,
 )
+from loopx.benchmark_adapters.skillsbench_acp_relay import (  # noqa: E402
+    CodexExecConfig,
+    SkillsBenchLocalAcpRelay,
+)
+from loopx.benchmark_adapters.skillsbench_bridge_summary import (  # noqa: E402
+    bridge_summary_task_progress_receipt,
+)
 from loopx.benchmark_adapters.skillsbench_turn_runtime import (  # noqa: E402
+    SkillsBenchTurnAgentResult,
     SkillsBenchTurnRuntimeConfig,
     build_skillsbench_loopx_turn_trace,
     run_skillsbench_loopx_turn,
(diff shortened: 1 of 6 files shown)


huangruiteng:
Premerge review packet:

- Changed surfaces: SkillsBench instrumented ACP bridge, bounded bridge progress receipt, Turn runtime validation fallback, public validation aggregation, and focused runtime tests.
- Focused checks: 44 pytest cases passed; Ruff lint passed; benchmark-run and agent-operation-snapshot smokes passed.
- Risk canary: 6/6 selected catalog canaries passed, Python compile and all diff checks passed.
- Known skip/failure: `skillsbench-loopx-turn-agent-cli-smoke.py` still fails with the same turn-planning bridge error on exact `origin/main`; this change does not own that pre-existing path.
- Manual hold: `benchmark_sensitive`; no self-merge and no benchmark launch from this PR.
- Coverage rationale: tests exercise new-file writes, changed-content writes, identical rewrites, failed writes, raw-material fail-closed behavior, receipt reduction, runtime commit eligibility, and route aggregation while public smokes cover adapter/evidence boundaries.

huangruiteng:
Added committed proof `cf8ff97d` for the repaired fallback:

- the durable public smoke now drives a real `run_loopx_turn_once` transaction through an instrumented command-file bridge
- its task-declared command validator intentionally fails, so only `verified_task_content_change` can validate progress
- observed receipt: execution `committed`, transaction validation `progress`, content-change count `1`, official feedback used `false`
- model invoked `false`, remote job launched `false`, raw material recorded `false`
- negative paths still prove validation failure, recoverable host failure, missing validator, and turn-plan failure do not spend

Validation after this commit: 44 focused pytest cases passed; the enhanced Turn smoke passed; Ruff lint passed; premerge selected 6/6 catalog canaries plus the public-boundary check, all passed. The existing `benchmark_sensitive` manual review hold remains.

PR #2450, opened 2026-07-22
Add controlled SkillsBench apt source fallback

huangruiteng:
## Summary
- add a bounded `mirror|primary` apt source mode for staged SkillsBench Dockerfiles
- keep mirror mode as the default; primary preserves task-owned sources while retaining apt retry hardening
- project the selected mode through runner config and setup-only public receipts

## Validation
- 51 focused pytest cases passed
- `skillsbench-failure-attribution-smoke` passed
- `skillsbench-benchmark-run-smoke` passed
- Python compile, shell syntax, and diff checks passed
- premerge canary: 6/6 catalog canaries and public-boundary scan passed

## Hold / scope
- premerge reports the expected `benchmark_sensitive` manual hold
- owner has explicitly authorized self-merge for bounded benchmark helper/runtime changes
- no scoring, task semantics, submission, upload, or verifier-feedback behavior changes
- no raw benchmark evidence, private state, credentials, local paths, or generated logs included

PR #2451, opened 2026-07-22
Honor SkillsBench primary apt source mode

huangruiteng:
## Summary
- gate Ubuntu and Debian mirror patch application on the selected apt source mode
- add a pytest-collected primary-source staging regression
- wire the existing primary-source check into the durable SkillsBench smoke entrypoint

## Validation
- 52 focused pytest cases passed
- `skillsbench-benchmark-run-smoke` passed
- 6/6 catalog canaries and the public-boundary scan passed

## Self-repair evidence
- an exact-main setup-only receipt showed `dockerfile_apt_source_mode=primary` while both mirror patches were still applied
- the invalid probe was stopped before agent/verifier execution and is not benchmark evidence
- no scoring, task semantics, submission, upload, permission, or verifier-feedback behavior changes
- no raw benchmark evidence, private state, credentials, local paths, or generated logs included

### DIFF
--- a/examples/skillsbench-benchmark-run-smoke.py
+++ b/examples/skillsbench-benchmark-run-smoke.py
@@ -14862,6 +14862,7 @@ if __name__ == "__main__":
     test_skillsbench_codex_acp_post_success_finalization_route()
     test_skillsbench_docker_task_staging_adds_app_skills_mount()
     test_skillsbench_docker_task_staging_adds_debian_apt_mirror_patch()
+    test_skillsbench_docker_task_staging_can_keep_primary_apt_sources()
     test_skillsbench_no_skill_route_removes_staged_task_skills()
     test_skillsbench_docker_task_staging_adds_apt_retry_patch()
     test_skillsbench_docker_task_staging_apt_retry_is_nonroot_safe()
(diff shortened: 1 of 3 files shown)


huangruiteng:
No actionable findings.

The prior exact-main setup-only probe exposed a real implementation/test gap: `primary` was projected, but mirror patch application was still unconditional. The probe was stopped before agent or verifier execution and is not treated as benchmark evidence.

Product/architecture judgment: this follow-up repairs the behavior at the existing staging boundary without expanding the public contract. It also moves the decisive regression into pytest's collected path and restores the existing durable smoke call, which directly addresses why PR #2450 validation missed the bug.

Validation:
- 52 focused pytest cases passed.
- `skillsbench-benchmark-run-smoke` passed with the primary-source case now executed.
- Python compile and diff checks passed.
- Premerge selected 6 catalog canaries; all passed.
- No checks were skipped or failed.

Manual hold: premerge reports the expected `benchmark_sensitive` hold. Owner authorization covers bounded benchmark helper/runtime self-merges. This three-file repair is single-purpose and does not change scoring, task semantics, submissions, permissions, or verifier feedback. Approved for admin-bypass squash merge once GitHub pytest passes.

PR #2472, opened 2026-07-23
fix(benchmark): preserve setup preflight repair route

huangruiteng:
## Summary

- preserve the typed SkillsBench setup-preflight attribution when a pre-agent runner setup diagnostic becomes the public failure class
- derive the repair route from that typed attribution without changing the stored failure class, scoring, countability, or runner behavior
- cover the apt fail-fast decision in the focused task-source preflight smoke

## Validation

- `examples/skillsbench-task-source-preflight-smoke.py`: pass on candidate; exact `origin/main` baseline fails at the repaired decision
- `loopx canary premerge --from-git-diff`: all 4 direct checks and 7 selected checks pass; public-boundary scan clean
- Full Public shard 5: pass at exact head `be5f3e43` under Python 3.11.15 with the workflow-equivalent configuration; 100/100 checks executed, 0 failures, 0 timeouts, 0 git-required skips, and 0 warnings
- the earlier 93/100 and 97/100 attempts were interrupted by host `ENOSPC`; the completed capacity-recovery run supersedes that validation hold
- no raw task text, logs, trajectories, verifier output, credentials, or local paths are committed

## Review hold

The canary classifies this as `benchmark_sensitive`, so this PR requires explicit maintainer review and must not be self-merged. The validation hold is cleared; the maintainer-review hold remains.

### DIFF
--- a/examples/skillsbench-task-source-preflight-smoke.py
+++ b/examples/skillsbench-task-source-preflight-smoke.py
@@ -421,7 +421,13 @@ def test_reverse_tunnel_app_goal_defaults_apt_bootstrap_fail_fast() -> None:
         ), case
         latest_run = _latest_decision_run(case)
         assert latest_run["failure_class"] == (
+            "skillsbench_runner_setup_blocked_before_agent_rounds"
+        )
+        assert latest_run["score_failure_attribution"] == (
             "skillsbench_docker_apt_setup_risk_preflight_blocked"
+        )
+        assert latest_run["repair_class"] == (
+            "skillsbench_setup_preflight_selection"
         )
         assert (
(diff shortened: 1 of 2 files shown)


huangruiteng:
Validation update for exact head `be5f3e43`:

- GitHub `pytest` is green.
- The focused task-source preflight smoke and the risk-based premerge canary remain green.
- After host capacity recovered, the workflow-equivalent Full Public shard 5 run passed under Python 3.11.15: 100/100 executed, 0 failures, 0 timeouts, 0 git-required skips, 0 warnings.
- The earlier `ENOSPC` attempts were host-capacity failures and are superseded by this complete run.

The `benchmark_sensitive` hold remains. Requesting explicit maintainer review; this PR must not be self-merged.

PR #2478, opened 2026-07-23
fix(quota): expire replan ACK at newer stalls

huangruiteng:
## Summary
- preserve the durable replan ACK timestamp in the compact projection
- let an ACK suppress only same-frontier stall evidence observed at or before that ACK
- add regression coverage for newer same-frontier stalls after an older watch-lane continuation

## Validation
- `uv run --extra test python -m pytest -q tests/control_plane/test_goal_vision_blocked_successor.py tests/control_plane/test_monitor_replan_agent_scope.py tests/control_plane/test_goal_frontier_replan_rules.py` (43 passed)
- `uv run --extra test ruff check loopx/control_plane/work_items/autonomous_replan_ack.py tests/control_plane/test_goal_vision_blocked_successor.py`
- autonomous replan obligation/read-model smokes and both no-progress regression contracts
- `loopx canary premerge --from-git-diff` (4 direct checks and 16 risk-selected checks passed; no warnings or manual holds)
- live public-safe quota projection changed from `monitor_quiet_skip` to `autonomous_replan_required` for newer post-ACK stalls

## Boundaries
- no benchmark jobs launched
- no scoring, runner, task semantics, permission, or submission behavior changed
- no private state, local paths, raw evidence, credentials, or generated logs included
- no authoritative check failures or skips; `uv` supplied the repository test environment

## Merge decision
The change is a narrow control-plane reducer fix with focused regression coverage. The standard premerge gate reports `self_merge_allowed=true`; coverage exercises quota/status consistency and the known no-progress contract.

### DIFF
--- a/loopx/control_plane/work_items/autonomous_replan_ack.py
+++ b/loopx/control_plane/work_items/autonomous_replan_ack.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 from typing import Any
 
+from ..runtime.time import parse_timestamp
+
 
 AUTONOMOUS_REPLAN_ACK_MATERIAL_RUN_WINDOW = 20
 
@@ -44,6 +46,9 @@ def compact_autonomous_replan_ack(run: dict[str, Any] | None) -> dict[str, Any] 
     frontier_identity = str(ack.get("frontier_identity") or "").strip()
     if frontier_identity:
         result["frontier_identity"] = frontier_identity
+    generated_at = str(run.get("generated_at") or "").strip()
+    if generated_at:
+        result["generated_at"] = generated_at
     return result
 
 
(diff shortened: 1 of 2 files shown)


huangruiteng:
No actionable findings.

**Open questions / assumptions**
- The compatibility fallback intentionally keeps legacy ACKs without timestamps behaving as before; only newly compacted ACKs gain temporal coverage.

- Motivation: quota was treating a same-frontier replan ACK as permanent, even when status had observed two newer durable no-progress events.
- Solved: the compact ACK now carries its existing run timestamp, and frontier matching compares that high-water mark with the newest obligation trigger.
- Operator impact: heartbeat receipts after an ACK can trigger a fresh bounded replan instead of remaining indefinitely quiet.
- Main risk: this is quota/status hot-path behavior, but the change is limited to ACK deduplication and preserves the legacy fallback when timestamps are absent.
- Design: a temporal high-water mark belongs in the reusable ACK reducer and is preferable to heartbeat- or benchmark-specific special cases.

**Validation**
- 43 focused quota/status tests passed.
- Ruff checks passed for both changed files.
- Autonomous replan obligation/read-model smokes and both no-progress regression contracts passed.
- Standard premerge canary passed 4 direct checks and 16 selected checks with no warnings, failures, skips, or manual holds.
- Live public-safe state reproduced the correction: newer post-ACK stalls project `autonomous_replan_required`.

**Merge decision**
- Self-merge is appropriate: the PR is single-purpose, public/private boundary clean, does not launch benchmark work or change scoring/permissions, and the premerge gate reports `self_merge_allowed=true`.

PR #2486, opened 2026-07-23
fix(onboarding): emit scoped refresh-state commands

huangruiteng:
## Summary
- generate bootstrap `refresh-state` commands through the existing shared renderer
- include the selected agent identity and explicit `agent_lane` scope in agent-bound startup guidance
- keep `health_check` derived/output-only and guard against exposing it as a caller argument

## Root cause
The executable CLI contract was already correct: multi-agent refreshes require a registered `--agent-id`, and an agent-bound refresh is scoped to `agent_lane`. The bootstrap command pack bypassed the shared renderer and hand-built only `refresh-state --goal-id ...`, while the Codex CLI bootstrap relied on an implicit scope default. This PR fixes the guidance generators instead of adding a redundant or forgeable `--health-check` input.

## Validation
- `git diff --check`
- Ruff on all four changed files
- changed-file Python 3.11 compile
- `examples/bootstrap-command-pack-smoke.py`
- `examples/control_plane/agent-onboard-host-loop-activation-smoke.py`
- `examples/codex-cli-bootstrap-message-smoke.py`
- 11 focused onboarding model-behavior tests
- `loopx canary premerge --from-git-diff --no-progress`: passed; 3/3 selected checks, public/private boundary clean, no warnings or manual holds

## Scope
No CLI option or state semantics changed. The generated command is also parsed by the real LoopX argparse parser in the focused smoke.

### DIFF
--- a/examples/bootstrap-command-pack-smoke.py
+++ b/examples/bootstrap-command-pack-smoke.py
@@ -238,6 +238,10 @@ def test_goal_text_invocation_plans_ranked_todos_before_activation() -> None:
         assert "--repo-path <approved-repo>" in reviewer_request_template
         assert "--execute" in reviewer_request_template
         assert "--agent-id codex-test-agent" in str(commands["goal_start_quota_should_run"])
+        refresh_command = str(commands["goal_start_refresh_state"])
+        assert "--agent-id codex-test-agent" in refresh_command
+        assert "--progress-scope agent_lane" in refresh_command
+        assert "--health-check" not in refresh_command
         assert "Same-priority items use that write order as the tie-breaker" in str(payload["message"])
         assert "preview the issue-fix route before todo writeback" in str(payload["message"])
         assert "PR lifecycle monitor" in str(payload["message"])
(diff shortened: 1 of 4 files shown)


huangruiteng:
Post-merge independent audit: **no findings**.

Reviewed the exact PR head `c29cc041a8e58b96db58959c8566e848083700cc` and confirmed merge commit `3225147b55995aaf81a5dc573e1831017481566d` is contained in current `origin/main` at `61c4789110f3545b6ce0171b6aad6d75f1cebd98`.

Product and architecture judgment:

- The generated `refresh-state` guidance now preserves agent identity and explicitly scopes progress to `agent_lane`.
- Reusing `render_refresh_state_command` removes a real CLI/guidance divergence without introducing a second contract.
- Omitting `--health-check` is correct because it is an output field, not a caller argument.
- The change is right-sized: two narrow production call-site updates plus focused parser and behavior coverage.

Validation:

- `git diff --check` passed for the PR range.
- `ruff check` passed on all four changed files.
- `bootstrap-command-pack-smoke.py`, `agent-onboard-host-loop-activation-smoke.py`, and `codex-cli-bootstrap-message-smoke.py` passed at the exact PR head.
- The same three smokes passed on current `origin/main`.
- 11 focused onboarding model-behavior tests passed at both the exact PR head and current `origin/main`.
- Production modules compiled under Python 3.11.

One initial local smoke invocation used system Python 3.9 and failed at import because this repository requires Python 3.11 or newer; rerunning with the supported interpreter passed. No product failure or follow-up hold remains.

PR #2493, opened 2026-07-24
Fix Codex CLI paste message line breaks

xielixing:
## Summary

- Keep the README Codex CLI setup prompt as one source line so copying it into the TUI submits one message.
- Apply the same README-only wording to the English and Chinese quick-start sections.

## Issue Or Task

- Closes #
- Contributor task ID:

## Validation

- [x] README paste-block single-line check
- [x] `git diff --check`
- [x] `python -m loopx.cli check` (passed with one existing active-state warning)

## Boundary Checklist

- [x] I did not commit `.loopx/`, `.codex/goals/`, live `ACTIVE_GOAL_STATE.md`, credentials, private benchmark traces, verifier output, raw agent sessions, internal document links, or local machine paths.
- [x] I kept the change scoped to the README Codex CLI install/paste section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,26 +287,11 @@ cd /path/to/your-project
 codex
 ```
 
-Then paste one setup message:
+Then paste one setup message. The paste block is intentionally one source line
+so Codex CLI receives one TUI message; do not add manual line breaks:
 
 ```text
-Connect this repo to LoopX from this visible Codex CLI TUI. Do not clone the
-LoopX repository for ordinary use. If `loopx` is not on PATH, install or repair
-it with the official no-clone installer:
-curl -fsSL https://raw.githubusercontent.com/huangruiteng/loopx/main/scripts/install-from-github.sh | bash
-export PATH="$HOME/.local/bin:$PATH"
-
-Then run `loopx doctor`. Work only from this project root: if LoopX state
-already exists, reuse it and do not create or overwrite a goal or the active objective; if the project
-is not connected, prefer `loopx connect`, and use `loopx bootstrap` only when
-project state clearly needs initialization. Ensure `.loopx/`, `.codex/goals/`,
-and `.local/` are ignored. Keep me in this TUI, do not use hidden headless
-execution. Then stop and report the project connection status, current user
-gate, top agent todo, and next safe action. After that I will start work with
-`$loopx <complex task>` or the `loopx` skill from `/skills`. When that task
-writes LoopX todos, generate the thin task body and set this visible TUI to
-`/goal <task_body>`; if you cannot mutate `/goal`, show me the exact text to
-paste instead of saying the loop is active.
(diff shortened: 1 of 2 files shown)


huangruiteng (changes requested):
Request changes on head `899751ab`.

## Findings

1. **[P1] The latest head no longer implements the runtime behavior or regression coverage claimed by the PR.** The current diff only changes `README.md` and `README.zh-CN.md`, but the PR summary still says that `--message-only` output is made paste-safe and that smoke coverage prevents multiline regressions. In the current tree, `render_codex_cli_bootstrap_message_text()` still emits a large multiline message (`loopx/project_prompt.py:560-680`), and the smoke only checks that CLI output equals that payload (`examples/codex-cli-bootstrap-message-smoke.py:206-220`); it does not require a single physical line. I reproduced this with the current `--message-only` command. Therefore users of the canonical generated bootstrap message still see the behavior this PR says it fixes.

   Please choose one honest scope:
   - If this is intentionally a README-only compatibility workaround, retitle and rewrite the PR description accordingly, link the affected host issue/version, and do not claim runtime or smoke coverage.
   - If the PR is meant to fix generated bootstrap messages, restore a narrow renderer-boundary change plus a negative regression test that fails when the copy payload contains an internal newline. Keep the readable structured source and derive the single-line copy form from it; do not maintain a second hand-written bootstrap/install policy.

2. **[P2] The documentation states a broad transport rule that the evidence only supports as a scoped host workaround.** `README.md:290-294` and `README.zh-CN.md:78-82` say that one source line makes Codex CLI receive one TUI message. The relevant upstream report appears to be [openai/codex#31416](https://github.com/openai/codex/issues/31416), which is a Windows/Codex CLI 0.142.5 multiline-paste bug and may depend on a CJK first line. Please describe this as a compatibility workaround for the affected paste path, not as a general source-line/message-framing contract. This also avoids turning a very long, difficult-to-review line into the canonical format without explaining the trade-off.

## Open Questions

- Was the force-push to a two-file, docs-only diff intentional?
- Is this specifically targeting `openai/codex#31416`, or is there a separate LoopX reproduction on another Codex CLI version/platform?

The user problem is real: a broken multiline paste can split the bootstrap instruction before LoopX is connected. A one-line copy payload is a reasonable bounded workaround. The current head only applies that workaround to two README blocks, while the canonical generated payload remains multiline and the PR metadata says otherwise. The clean contract is one authoritative structured bootstrap message with an explicit, tested paste-safe rendering at the final copy boundary; a docs-only workaround is also acceptable if it is named and scoped honestly.

## Validation

- `git diff --check origin/main...HEAD`: passed.
- `uv run --python 3.12 python examples/codex-cli-bootstrap-message-smoke.py`: passed, but does not cover the claimed no-newline invariant.
- `loopx check --scan-path README.md --scan-path README.zh-CN.md`: passed, 9 checks, no errors or warnings.
- Manual `codex-cli-bootstrap-message --message-only` check: still emits internal newlines.
- Synthetic merge-tree against current `origin/main`: clean.
- Remote status checks: none reported.

## Merge Decision

**Request changes.** The README workaround itself is small, but the latest head, PR description, and claimed validation do not describe the same shipped behavior.

PR #2494, opened 2026-07-24
Fix agent-lane projection smoke parity

huangruiteng:
## Summary

- clear stale next-action projection warnings when agent-lane scoping removes the corresponding latest-run recommendation
- migrate projection smokes to assert current-agent run isolation
- provide explicit scheduler execution context in direct quota fixtures

## Validation

- `python3 examples/control_plane/status-markdown-smoke.py`
- `python3 examples/control_plane/status-neutral-run-window-smoke.py`
- `python3 examples/project/next-action-projection-contract-smoke.py`
- `pytest -q tests/control_plane/test_monitor_replan_agent_scope.py` (8 passed)
- `ruff check` on all changed Python files
- `loopx canary premerge --from-git-diff` (18/18 passed; no manual holds)
- full-public shard 4: 100/100 passed
- full-public shard 3: 98/100 passed in parallel; the two installer checks passed when rerun serially with explicit Python 3.12 (the parallel run used macOS Python 3.9 for fresh-clone and hit the 120-second timeout for install-local)

## Review

The runtime change stays in the existing status projection synchronization helper. It does not add a compatibility layer or a new abstraction.

### DIFF
--- a/examples/control_plane/status-neutral-run-window-smoke.py
+++ b/examples/control_plane/status-neutral-run-window-smoke.py
@@ -13,6 +13,9 @@ REPO_ROOT = Path(__file__).resolve().parents[2]
 sys.path.insert(0, str(REPO_ROOT))
 
 from loopx.diagnose import collect_diagnosis  # noqa: E402
+from loopx.control_plane.scheduler.execution_context import (  # noqa: E402
+    GENERIC_CLI_OUTER_CONTROLLER_SCHEDULER_CONTEXT,
+)
 from loopx.quota import (  # noqa: E402
     QUOTA_MONITOR_POLL_CLASSIFICATION,
     QUOTA_SLOT_SPENT_CLASSIFICATION,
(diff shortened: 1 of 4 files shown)


huangruiteng:
Self-review: approved for the narrow CI repair after checks complete.

Changed surfaces: status agent-lane projection synchronization and three public contract smokes. The helper now removes a warning when the scoped guard no longer supplies the recommendation that warning describes; it does not change goal-level state or add a compatibility path.

Validation: focused projection smokes passed, the eight agent-scope unit tests passed, changed files passed Ruff and compile checks, and the standard premerge gate passed 18/18 with no manual holds. Full-public shard 4 passed 100/100. Shard 3 passed 98/100 in parallel; its fresh-clone and install-local checks then passed serially with an explicit supported Python. Dashboard browser validation was skipped because local npm dependencies were unavailable; this change does not touch dashboard code.

Coverage is sufficient because it exercises both sides of the invariant: an own agent-lane run is retained and rendered, while a missing own-lane run removes both the cross-lane recommendation and its stale warning.

PR #2509, opened 2026-07-24
Fix release smoke archive cache race

huangruiteng:
## Summary

- stop the three release archive smoke helpers from recursively re-traversing directories already selected by their filtered `rglob` walk
- keep `__pycache__` and transient `.pyc` files out of generated release fixtures
- add a parameterized regression test covering all three helpers

## Root cause

[Full Public Smokes run 30068072844](https://github.com/huangruiteng/loopx/actions/runs/30068072844) failed on main because `tarfile.add()` recursively entered a nested `__pycache__` directory after the outer walk had filtered it. A temporary `.pyc` file disappeared between directory enumeration and `lstat`, raising `FileNotFoundError`. The smoke-fleet-health failure was downstream of that shard failure.

## Validation

- `pytest -q tests/test_release_smoke_archive_helpers.py`: 3 passed
- Ruff on all four changed files: passed
- no-clone release verification smoke: passed
- packaged install smoke: passed
- Codex CLI TUI bootstrap bundle smoke: passed
- standard premerge gate: 18/18 passed, 0 manual holds, public boundary clean
- Full Public shard 500: the repaired no-clone smoke passed; one unrelated concurrency smoke hit the local 120-second limit and passed when rerun standalone

No runtime product behavior, release metadata, credentials, private state, or generated logs are included.

### DIFF
--- a/examples/codex-cli-packaged-install-smoke.py
+++ b/examples/codex-cli-packaged-install-smoke.py
@@ -19,7 +19,11 @@ def add_tree(tar: tarfile.TarFile, root: Path, name: str) -> None:
         for child in sorted(path.rglob("*")):
             if ".git" in child.parts or "__pycache__" in child.parts:
                 continue
-            tar.add(child, arcname=str(Path("loopx-main") / child.relative_to(root)))
+            tar.add(
+                child,
+                arcname=str(Path("loopx-main") / child.relative_to(root)),
+                recursive=False,
+            )
     else:
         tar.add(path, arcname=str(Path("loopx-main") / name))
 
--- a/examples/codex-cli-tui-bootstrap-smoke-bundle-smoke.py
+++ b/examples/codex-cli-tui-bootstrap-smoke-bundle-smoke.py
@@ -22,7 +22,11 @@ def add_tree(tar: tarfile.TarFile, root: Path, name: str) -> None:
         for child in sorted(path.rglob("*")):
             if ".git" in child.parts or "__pycache__" in child.parts:
                 continue
-            tar.add(child, arcname=str(Path("loopx-main") / child.relative_to(root)))
+            tar.add(
+                child,
+                arcname=str(Path("loopx-main") / child.relative_to(root)),
+                recursive=False,
+            )
     else:
         tar.add(path, arcname=str(Path("loopx-main") / name))
 
--- a/examples/release/codex-cli-no-clone-release-verification-smoke.py
+++ b/examples/release/codex-cli-no-clone-release-verification-smoke.py
@@ -64,7 +64,11 @@ def add_tree(tar: tarfile.TarFile, root: Path, name: str) -> None:
         for child in sorted(path.rglob("*")):
             if ".git" in child.parts or "__pycache__" in child.parts:
                 continue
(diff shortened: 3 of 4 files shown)


huangruiteng:
Exact-head validation receipt for `2095a6d5`:

- Changed surfaces: three public release/install smoke archive helpers and one focused pytest regression.
- Root contract: the filtered `rglob` walk is the only recursive traversal; each `tar.add` call is non-recursive, so nested `__pycache__` entries and transient `.pyc` files cannot re-enter the fixture.
- Focused pytest: 3/3 passed across all three helpers.
- Ruff: passed on all four changed files.
- Direct smokes: packaged install, Codex CLI TUI bootstrap bundle, and no-clone release verification all passed.
- Standard premerge gate: 18/18 passed; 0 warnings, 0 failures, 0 manual holds; public/private boundary clean.
- GitHub Python Tests: passed at exact head: https://github.com/huangruiteng/loopx/actions/runs/30070435163
- Full Public shard 500 local qualification executed 100 checks. The repaired no-clone smoke passed. One unrelated release-promotion concurrency smoke exceeded the local 120-second per-check limit; it passed when rerun standalone, so this is recorded as a local timing limitation rather than hidden as green.

This coverage is sufficient for the narrow test-fixture change because it proves the cache exclusion invariant directly, exercises every duplicated helper, validates the affected release/install routes, and runs the repository risk-based install/release/public-boundary gate. No product runtime, release metadata, permission boundary, or production action changes are included.

huangruiteng:
Post-merge independent audit at exact PR head `2095a6d5`: **no blocking findings**.

## Findings

None. Setting `recursive=False` on each `tar.add()` makes the already-filtered `rglob()` walk the sole recursive traversal. A parent directory can no longer re-enter an excluded nested `__pycache__`, while directory entries and allowed descendants are still archived individually.

- The fix intentionally addresses concurrent Python-cache churn, not arbitrary source files disappearing during archive creation. That scope matches the reproduced failure and PR contract.
- The PR merged before the independent review completed. This comment is therefore a post-merge correctness audit, not merge authorization.

This is a narrow root-cause fix to release-test fixture construction, with no product runtime or release-metadata change. The three helpers retain four lines of equivalent code, but consolidating them would add import/package coupling to standalone packaged-install, TUI-bundle, and no-clone smoke entrypoints. The small duplication is justified here.

The 39-line parameterized regression is sufficient: it exercises all three helpers, proves an allowed nested file remains present, and proves every nested Python cache entry is absent. I also replayed the same fixture against each pre-fix helper; all three old implementations leaked `__pycache__` into the archive, so the test is a genuine regression guard rather than characterization derived only from the new output.

## Validation

- `git diff --check` against PR base: passed
- focused Python 3.12 pytest: 3/3 passed
- Ruff on all four changed files: passed
- pre-fix mutation replay: all three old helpers reproduced the cache leak
- remote exact-head Python Tests check: passed

I did not rerun the three heavier end-to-end smoke routes during this audit. Their exact-head results and the one disclosed local timeout are preserved in the validation receipt above; the focused regression, mutation proof, static scope review, and remote CI are sufficient for this 54-line test-helper-only change.

## Merge decision

Already merged as `436b03c6`. Post-merge audit passes; no corrective follow-up is needed.

PR #2528, opened 2026-07-24
test(extensions): support relocatable Python venvs

huangruiteng:
## Summary

- create generated-extension test venvs with symlinks on POSIX, matching `python -m venv` behavior
- preserve copied executables on Windows
- keep the change test-only; LoopX extension runtime behavior is unchanged

## Why

Relocatable macOS CPython resolves `libpython` relative to the executable. `venv.EnvBuilder` defaults to copying the executable, so the nested test venv aborts before exercising LoopX. Symlinking on POSIX preserves the base runtime path and makes the lifecycle test portable.

## Validation

- reproduced the pre-change `@rpath/libpython3.13.dylib` failure
- exact managed-lifecycle test passes on relocatable CPython 3.13
- POSIX symlink venv boots on relocatable CPython 3.12
- full `tests/extensions/test_extension_scaffold.py`: 10 passed
- Ruff and LoopX public-boundary check passed
- standard premerge gate passed; self-merge allowed

### DIFF
--- a/tests/extensions/test_extension_scaffold.py
+++ b/tests/extensions/test_extension_scaffold.py
@@ -187,7 +187,12 @@ def test_generated_starter_installs_and_runs_through_managed_lifecycle(
         execute=True,
     )
     environment = tmp_path / "venv"
-    venv.EnvBuilder(with_pip=True, system_site_packages=True).create(environment)
+    # Match `python -m venv` so relocatable POSIX runtimes keep their libpython path.
+    venv.EnvBuilder(
+        with_pip=True,
+        system_site_packages=True,
+        symlinks=os.name != "nt",
+    ).create(environment)
     binary_dir = environment / ("Scripts" if os.name == "nt" else "bin")
     python = binary_dir / ("python.exe" if os.name == "nt" else "python")
     wheel_dir = tmp_path / "wheelhouse"


huangruiteng:
Self-review and validation receipt (exact head `2d0435d2`):

- Changed surface: test fixture only (`tests/extensions/test_extension_scaffold.py`); no LoopX runtime/API behavior changed.
- Root cause: relocatable macOS CPython links `@rpath/libpython`; copied nested-venv executables resolve against the child venv and abort before the extension lifecycle runs.
- Repair: use POSIX symlinks, matching `python -m venv`; retain copied executables on Windows.
- Validation: reproduced the pre-change CPython 3.13 failure; exact test passes after the change; relocatable CPython 3.12 symlink venv boots; extension scaffold test file passes 10/10; Ruff passes; LoopX public-boundary check passes; standard premerge gate passes with no holds and `self_merge_allowed=true`.
- Skips/failures: none in the risk-matched validation. GitHub pytest is still pending and will be required before merge.

Coverage is sufficient because the failure occurs in test environment construction, the patch changes only that constructor, and the existing managed-lifecycle test still exercises build, install, doctor, enable/disable, and run end to end.

huangruiteng:
Final exact-head review: no findings.

The required GitHub `pytest` check passed. This remains a single-purpose,
test-only portability correction with no LoopX runtime, permission, or public
evidence-policy change. The existing managed lifecycle test preserves the
end-to-end product contract, while the fixture now follows the platform venv
default on POSIX. Self-merging under the repository's narrow-cleanup policy.

PR #2530, opened 2026-07-24
fix(quota): project explicit agent blockers

huangruiteng:
## Summary

- preserve typed blocked `blocker` todos in the canonical todo read model even when they fall outside the generic item window
- project only current-agent blockers with an explicit reason into the quota hot path
- treat that blocker as an auditable vision wait so a recorded blocker delta clears repeated autonomous replan without closing the active vision
- keep other-agent and malformed blockers non-authoritative
- retain bounded JSON and Markdown blocker identity/reason output

## Validation

- `447 passed` for `tests/control_plane`
- `68 passed` for focused frontier, blocker, CLI projection, and todo consistency tests
- Ruff passed on all changed Python files
- `loopx check` passed with 0 errors and 0 warnings
- exact-head `loopx canary premerge --from-git-diff --tier standard` passed: 17/17 selected checks, no manual holds, self-merge allowed
- hot-path interface budget passed (`quota_should_run_json` remains within the existing bound)

The first premerge pass caught zero-blocker payload overhead; the projection now omits blocker fields when no authoritative blocker exists, and the final exact-head gate passed.

## Boundary

- no private state, credentials, local paths, raw logs, or generated evidence are included
- no production, permission, scheduler-host, or benchmark behavior changes

PR #2531, opened 2026-07-24
fix(quota): honor explicit blocker precedence

huangruiteng:
## Summary
- let an explicit, reasoned blocker claimed by the current agent own the vision-wait read model ahead of an unrelated deferred successor
- treat the authoritative current-agent blocker as a no-replan frontier rule after a blocker delta ACK clears an existing obligation
- keep the exact deferred successor visible in the agent-scope diagnostic projection and add precedence regressions

## Why
PR #2530 made blocker identity and reason visible, but the live quality-qualification lane still selected an older exact deferred successor first and immediately re-derived a stalled-monitor replan. That contradicted the documented blocker transition: the repair existed in state but could not own the frontier.

## Validation
- `449 passed` in `tests/control_plane`
- focused blocker/vision/rule tests: `45 passed`
- `loopx check`: 0 errors, 0 warnings, public boundary clean
- exact committed diff `loopx canary premerge --from-git-diff --git-diff-base origin/main --tier standard`: 17/17 checks passed, no manual holds, self-merge allowed
- live registry shadow with this head no longer emits `autonomous_replan_required`; the repair todo becomes the selected normal-run advancement while it is open

## Boundaries
No private state, raw logs, credentials, local paths, benchmark semantics, permissions, or production actions are included.

### DIFF
--- a/loopx/control_plane/goals/goal_frontier.py
+++ b/loopx/control_plane/goals/goal_frontier.py
@@ -1436,6 +1436,9 @@ def derive_goal_frontier_replan_obligation_from_summaries(
                     delta_kind="watch_lane_continuation",
                 )
             ),
+            current_agent_blocker_count=safe_non_negative_int(
+                (agent_todo_summary or {}).get("current_agent_blocker_count")
+            ),
             monitor_no_change_streak_triggered=(
                 monitor_no_change_trigger is not None
             ),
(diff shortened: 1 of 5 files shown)


huangruiteng:
Maintainer self-review:

- **Changed surfaces:** goal vision-wait precedence and goal-frontier replan rule selection, plus focused control-plane regressions.
- **Behavioral judgment:** an explicit typed blocker with a non-empty reason and an exact current-agent claim is a stronger statement of the lane's present frontier than an unrelated deferred successor. It now owns the wait projection. The deferred successor remains visible in `agent_scope_frontier.blocked_successor_wait_candidates`.
- **Safety judgment:** missing-reason and other-agent blockers remain non-authoritative. Strict vision gaps, required checkpoints, permissions, and production behavior are unchanged.
- **Validation:** 449 control-plane tests passed; focused blocker/vision/rule tests passed 45/45; Ruff, Python compile, and `git diff --check` passed; `loopx check` reported 0 errors and 0 warnings with a clean public boundary; exact committed diff premerge passed all 17 selected checks.
- **Failures/skips/manual holds:** none.
- **Coverage rationale:** the rule-table test proves blocker precedence over a stalled monitor, the integration test proves blocker precedence over an exact deferred successor while retaining successor diagnostics, and the complete control-plane suite covers surrounding quota/status behavior.

The diff is narrow, public-safe, and the risk gate reports `self_merge_allowed=true`. I consider it ready for self-merge after the required GitHub check passes.

PR #2550, opened 2026-07-25
feat(decision-context): add default-off profile entrypoint

huangruiteng:
## Summary

- add a default-off, goal-scoped Decision Context profile and public-safe activation status
- add a replaceable source-provider registry with one explicitly configured, read-only `local-file` adapter
- expose `inspect-profile` and `source-manifest` CLI entry points without provider access or Core mutation
- document the remaining P1 boundary: source scan orchestration, evidence/proposal composition, and private cursor checkpoint apply

Decision Context already had packet, source, and evidence-assembly contracts, but no safe activation path. This change gives operators a concrete way to prove that the capability is disabled by default, scoped to one goal and agent, and backed only by registered source adapters before any authority read occurs.

The implementation stays inside the existing `decision-context` capability. It does not change todo, gate, quota, authority, or action permissions. Private profile data, source locators, cursor values, exact-read content, and credentials remain transient and are excluded from public status and manifest packets.

This is intentionally a partial P1 slice. It makes activation and source-provider boundaries executable; it does not claim that end-to-end decision orchestration or private dogfood is complete.

## Validation

- `ruff check` on changed Python and test files
- strict `mypy` on the new profile, provider, and CLI modules
- focused Decision Context and capability catalog tests: 55 passed
- full test suite: 1273 passed, 2 skipped
- `loopx canary premerge --from-git-diff --format json`: passed
  - 4 direct checks
  - 13 selected canary/boundary checks
  - 0 failures, warnings, or manual holds
  - public/private boundary scan clean across all 8 changed files

## Merge boundary

Self-merge is requested under maintainer authorization for this capability slice. No benchmark semantics, permission model, Core lifecycle semantics, external write path, or public first-screen surface changes.


huangruiteng:
## Maintainer self-review

### Findings

No blocking findings after diff review.

One privacy issue was found and fixed before merge: the unavailable-provider status initially echoed an unregistered adapter identifier from the private profile. The final diff now projects only `unregistered` plus a count, and the negative test asserts the private adapter token is absent.

### Assumptions

- Provider and source ids in an enabled source manifest are caller-selected public-safe aliases.
- The built-in local-file adapter is an explicit, bounded, read-only authority source; source-scan orchestration and private cursor checkpoint apply remain follow-up work.

This is the correct capability boundary: activation, provider resolution, and public-safe projection live under `decision-context`, while Core todo/gate/quota/authority semantics remain unchanged. The slice solves the immediate missing entrypoint without claiming end-to-end decision automation. The main risk is private-profile leakage; strict profile parsing, transient private fields, redacted unavailable-provider status, and changed-file boundary scans cover that risk.

### Validation

- changed-file Ruff and strict Mypy: passed
- focused Decision Context and catalog tests: 55 passed
- full local suite before the final privacy-only hardening: 1273 passed, 2 skipped
- focused tests after the privacy hardening and profile/activation naming correction: 55 passed
- latest premerge gate: 4 direct checks and 13 selected checks passed
- public/private boundary scan: 8 changed files, 0 hits
- required GitHub pytest check on the final head: passed
- failures, warnings, skips requiring action, and manual holds: none

### Merge decision

Approved for maintainer self-merge once the required GitHub check is green. Owner authorization covers this Decision Context capability slice; no benchmark semantics, permissions, Core lifecycle semantics, external writes, or public first-screen surfaces change.

PR #2565, opened 2026-07-26
feat(materials): add lossless ranked entry rebuild

huangruiteng:
## Summary
- add provider-neutral ranked-entry rebuild plan and apply receipt contracts
- require oversized entries to split into independently ranked children instead of hiding overflow
- enforce exact coverage, unique membership, source lineage, contiguous ranks, protected anchors, and rollback metadata
- register and document the default-off Material Lifecycle capability surface

## Validation
- 42 focused Material Lifecycle tests passed
- Ruff passed
- git diff --check passed
- private dogfood rebuilt 30 source bundles into 40 entries with 89/89 exact coverage; canonical records and raw sources remained unchanged (private evidence only)

### DIFF
--- a/docs/reference/protocols/material-lifecycle-architecture-v0.md
+++ b/docs/reference/protocols/material-lifecycle-architecture-v0.md
@@ -72,6 +72,27 @@ copying raw content into another queue.
 requires an owner-gate reference, validation reference, before/after revisions,
 and rollback reference for applied changes.
 
+`material_ranked_entry_rebuild_plan_v0` handles structural queue repair that a
+bounded rank move cannot express. A ranked entry that exceeds the configured
+member budget must become two or more independently sortable entries; overflow
+cannot be hidden in a supporting-only index. The caller supplies the semantic
+grouping after exact reads. The provider-neutral builder then enforces:
+
+- at most `max_materials_per_entry` references per rebuilt entry;
+- every source material reference appears exactly once;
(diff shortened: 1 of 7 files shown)


huangruiteng:
No blocking findings.

- Motivation: a display-only cap can make oversized material bundles look smaller while leaving overflow outside the active ranking.
- Solved: the new default-off Material Lifecycle contract requires oversized entries to become independently sortable children and verifies exact one-time coverage, source lineage, contiguous ranks, protected anchors, and rollback metadata.
- User/operator impact: a bounded Top window can stay readable without making lower-priority materials disappear from the ranked system.
- Main risk: the generic contract must not learn private catalog semantics. The implementation only accepts opaque material references and delegates exact-read grouping and store writes to the caller.
- Design judgment: this belongs to the existing Material Lifecycle capability rather than Core or a project-specific adapter. Private dogfood exercised the contract, while no private labels, paths, content, or evidence entered this PR.

## Validation

- 42 focused Material Lifecycle tests passed.
- Full GitHub Python Tests check passed.
- `loopx canary premerge --from-git-diff` passed with 10 selected checks, zero failures, zero warnings, and no manual holds.
- `git diff --check origin/main...HEAD` passed.

## Merge decision

Self-merge is appropriate: the change is single-purpose, default-off, provider-neutral, outside runtime and permission hot paths, fully validated, and owner-authorized.

PR #2583, opened 2026-07-27
docs(course): explain ordered frontier replan rules

huangruiteng:
## Summary

- clarify that Turn uses predefined, ordered, resumable transaction phases rather than a fixed business workflow
- align the lecture pseudocode rule names with the real `GoalFrontierReplanRule` identities
- explain all 15 first-match Frontier Replan rules, including trigger meaning, short-circuit purpose, and whether each derives a new obligation
- group the ordering into authority/frontier preservation, continuity/direction repair, and monitor-tail audit

## Validation

- exact-scope change-quality receipt `cqr_8d498cbf3ceaa3172676`: valid, 0 findings
- `uv run --with pytest pytest -q tests/control_plane/test_goal_frontier_replan_rules.py`: 19 passed
- standard premerge: 13/13 selected checks passed, 0 failures, 0 warnings, 0 manual holds
- `loopx check --scan-path docs/development/control-plane-course/topic-long-horizon-convergence.md`: clean
- `git diff --check` and public/private boundary scan: passed

## Risk

Docs-only. The explanations are grounded in the existing rule implementation and focused rule tests; runtime behavior is unchanged.

### DIFF
--- a/docs/development/control-plane-course/topic-long-horizon-convergence.md
+++ b/docs/development/control-plane-course/topic-long-horizon-convergence.md
@@ -350,7 +350,7 @@ def advance_one_turn(status, goal_id, agent_id, host):
     return execution
 ```
 
-真实 Turn transaction 使用固定 phase：
+真实 Turn transaction 使用预定义且顺序固定的可恢复事务阶段：
 
 ```text
 host_execute
@@ -362,7 +362,7 @@ host_execute
   -> scheduler_ack
 ```
 
-固定 phase 有两个作用：
+这组事务阶段有两个作用：
 
 1. **恢复**：journal 已经记录 `durable_writeback` 时，重启后不能再次执行外部 effect；
 2. **归因**：只有 validation 通过并写回成功，才允许 spend；scheduler apply 与 ACK 也有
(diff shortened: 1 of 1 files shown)


huangruiteng:
No blocking findings.

- Motivation: the lecture showed the ordered Frontier Replan code, but the three summary bullets did not explain why several matching rules deliberately return `derives_obligation=false`.
- Solved: the teaching aliases now match the real rule identities, and all 15 first-match rules state their trigger meaning, short-circuit purpose, and obligation outcome. The three-layer grouping makes the precedence contract explicit without introducing another abstraction.
- User impact: a developer can now explain both “why this turn replans” and “why it intentionally does not replan,” reducing the risk of treating gates, successors, blockers, or monitors as interchangeable empty-frontier signals.
- Main risk: documentation can drift when the rule order changes. The table is grounded in the current enum order, implementation reasons, and parameterized decision-table tests.
- Design judgment: keeping one ordered table beside the pseudocode is the right-sized form. Duplicating the decision logic in a second executable helper would create another authority surface.

**Validation**

- Exact-scope quality receipt `cqr_8d498cbf3ceaa3172676`: valid; 0 blockers, 0 warnings, 0 advisories; no safe-fix pass.
- Focused rule tests: 19 passed.
- Standard premerge selection: 13/13 passed; 0 failures, 0 warnings, 0 manual holds.
- LoopX document contract check: 0 errors, 0 warnings.
- Diff hygiene and public/private boundary scan: passed.
- GitHub reported no hosted checks for this docs-only branch. Direct `pytest` was not on the shell PATH, and the first `uv run pytest` lacked the optional test dependency; the final explicit `uv run --with pytest` invocation passed all focused tests.

Coverage is sufficient because the PR changes one public lecture file, exercises the exact rule decision table it explains, and changes no runtime, permission, schema, or evidence behavior.

**Merge decision:** self-merge approved under the owner’s explicit authorization.

PR #2587, opened 2026-07-27
refactor(todos): centralize metadata field schema

huangruiteng:
## Summary

- make one canonical field schema own todo metadata reads, writes, validation, ordering, and encoding
- remove the seven singular metadata aliases and hyphenated-key normalization after the migration window; reads and writes now use canonical underscore/plural field names
- preserve the public formatter signature, canonical output order, and fail-closed removed-continuation repair marker
- add round-trip and negative semantic coverage, then retire the two parse/format maintainability exceptions

## Validation

- `pytest -q tests/control_plane/test_todo_metadata_schema.py tests/control_plane/test_contract_quality.py`: 15 passed
- active-state structured projection, quota action-scope, todo claim-visibility, and Explore result-node linkage smokes passed
- Ruff check passed on the three changed files
- diff hygiene and public/private boundary scan passed
- full suite intentionally not run: the delta is confined to todo metadata key recognition and the direct consumer smokes cover the affected read paths
- current-head GitHub `pytest` passed in 3m52s

## Review focus

Please confirm the strict metadata boundary: singular and hyphenated field names are ignored, canonical underscore/plural names remain the only read/write contract, and removed continuation-policy values remain an explicit fail-closed repair marker rather than an executable legacy alias.

### DIFF
--- a/examples/control_plane/active-state-structured-projection-smoke.py
+++ b/examples/control_plane/active-state-structured-projection-smoke.py
@@ -32,7 +32,7 @@ def active_state_text() -> str:
         "  <!-- loopx:todo todo_id=todo_user_gate status=open task_class=user_gate blocks_agent=codex-product-capability -->\n\n"
         "## Agent Todo\n\n"
         "- [ ] [P0] Add active-state structured projection parity smoke.\n"
-        "  <!-- loopx:todo todo_id=todo_agent_projection status=open task_class=advancement_task claimed_by=codex-product-capability target_capability=active_state_schema_migration -->\n"
+        "  <!-- loopx:todo todo_id=todo_agent_projection status=open task_class=advancement_task claimed_by=codex-product-capability target_capabilities=active_state_schema_migration -->\n"
         "- [ ] [P1] Watch external launch evidence.\n"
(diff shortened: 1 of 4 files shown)


huangruiteng:
Owner direction applied in `bdd8189e`.

Changed surfaces:
- removed the schema-level singular alias table for seven todo metadata fields;
- stopped normalizing hyphenated metadata keys to underscores and added a token-boundary guard so suffixes such as `legacy-status` cannot be misread as canonical `status`;
- migrated the remaining public projection smoke fixture to `target_capabilities`;
- retained removed continuation-policy handling only as a fail-closed repair marker.

Focused validation:
- 15 todo metadata schema/contract tests passed;
- active-state structured projection, quota action-scope, todo claim-visibility, and Explore result-node linkage smokes passed;
- Ruff check and `git diff --check` passed;
- public/private boundary scan found no new private material.

The full suite was intentionally skipped because the behavior change is limited to metadata-key recognition; direct parser tests plus four consumer smokes cover the affected boundary. Current-head CI remains the independent repository check. PR stays open for owner review because this is control-plane behavior.

PR #2640, opened 2026-07-29
fix(host): keep capability reentry visible outside Goal prompt

huangruiteng:
## Summary

- keep the Ark Managed Agent one-shot Goal prompt stable
- reuse the existing `runtime_capability_reentry_v0` packet instead of adding a second envelope or event protocol
- project that packet near the beginning of `quota should-run` JSON so bounded host tool-result capture cannot hide it behind large diagnostics
- keep capability claims real-callsite-gated, session-scoped, and non-durable

## Simplification

The earlier version introduced host continuation/event builders and a second capability envelope without a production caller. Those speculative surfaces and their duplicate tests have been removed. The effective PR is now 7 files and +139/-2; most additions are focused regression tests and contract documentation.

## Validation

- full repository suite: 1589 passed, 2 skipped
- focused capability/host tests: 10 passed
- Ruff: passed
- post-commit premerge canary: 16/16 selected checks passed
- public-boundary scan: passed
- real CLI projection: reentry packet starts at byte 204 in a 59 KB quota result


huangruiteng:
Updated after the host-boundary audit:

- rebased the branch state by merging current `main`;
- strengthened the regression contract so the Managed Agent Goal `task_body` rejects both typed packet names and generic capability-probing guidance (`Before the first quota guard`, `--available-capability <name>`);
- capability discovery/re-entry remains exclusively in the typed `quota_tool_result` / `host_continuation_input` contract.

Validation:
- focused pytest: 14 passed;
- current-source prompt readback: 2293 chars, all four capability prompt markers absent;
- `loopx canary premerge --from-git-diff`: passed, 10/10 selected checks, public boundary clean, no manual holds.

A whole-repository local pytest invocation was not claimed as passing: an unrelated process-isolation test terminated its parent shell around 4%, so CI remains the authoritative full-suite rerun for this updated head.

huangruiteng:
Live host follow-up (head `12c72960`): an authenticated Ark Managed Agent canary exposed a transport-budget failure in the earlier implementation.

The complete `quota should-run` JSON was 27,775 bytes, while the host captured about 16,443 bytes from the bash tool result. The nested `runtime_capability_reentry_v0` packet first appeared at byte 22,213, so the model never received it and improvised capability discovery instead of honoring the typed host boundary. No private session data or credentials are included here.

This update keeps the nested interaction-contract packet canonical and adds an identical top-level CLI hot-path projection near the start of JSON output. On the representative fixture the packet now begins at byte 153 (schema at byte 210), while the Goal task body still contains no capability guidance. A 20 KB diagnostic regression proves the packet remains within the first 1 KB and that normal decisions do not gain a false projection.

Validation:
- focused capability/Managed-Agent/activation tests: 28 passed
- Ruff check and `git diff --check`: passed
- standard premerge: 16/16 selected checks passed, public boundary clean, no manual holds

The PR remains ready for owner review; it is not self-merged.

huangruiteng:
Host/controller follow-up (head `d0676dda`):

The previous live canary proved that the initial re-entry packet must be visible before bounded tool-result truncation. It also exposed a second gap after verification: a later quota decision could contain inherited capability flags only inside the large nested interaction contract, so a host or model-side projection could drop them before refresh/spend.

This update adds the verified half of the same protocol:

- `runtime_capability_envelope_v0` carries session-scoped verified capabilities, exact repeatable CLI argv, and current next actions;
- the JSON CLI promotes the envelope near the beginning of quota output, while keeping the nested interaction-contract field canonical;
- the Managed Agent host exposes a typed builder that converts either a verification-required re-entry packet or a verified envelope into ordinary `host_continuation_input` before the next model turn;
- the Goal task body is not regenerated or mutated, and no durable capability grant is written.

Representative output was 26,758 bytes: the verified envelope began at byte 155 (schema at byte 192), while the canonical interaction contract began at byte 20,502. A same-invocation baseline comparison produced identical Goal task-body SHA-256 values and zero capability markers.

Validation:
- focused host/capability/Issue-Fix tests: 32 passed;
- full pytest: 1,594 passed, 2 skipped;
- Ruff and diff checks: passed;
- standard premerge: 18/18 selected checks passed, public boundary clean, no manual holds.

This proves the public controller seam and local transport contract. A cloud transport still needs a real between-turn controller/interposition hook to qualify deferred delivery end to end; this PR does not simulate that boundary by changing a Goal or system prompt.

PR #2648, opened 2026-07-29
fix(benchmark): close SkillsBench source and termination gaps

huangruiteng:
## Summary

- recognize BenchFlow's keyword environment argument so the host-local LoopX source mount is actually injected
- preserve the last public supervisor checkpoint when termination occurs
- emit a public-safe terminal closeout and clean local and remote processes on runtime termination

## Validation

- `uv run --python 3.12 --with pytest pytest -q tests/test_skillsbench_runtime_tools.py tests/test_skillsbench_reverse_tunnel_supervisor.py` (17 passed)
- `uv run --python 3.12 python examples/skillsbench-reverse-tunnel-supervisor-smoke.py`
- `uv run --python 3.12 python examples/skillsbench-benchmark-run-smoke.py`
- `loopx canary premerge --from-git-diff` (changed-Python compile and 4 catalog canaries passed; benchmark-sensitive manual hold remains)
- `git diff --check` and diff-only public/private boundary scan

## Scope

This does not change benchmark scoring, task semantics, leaderboard or submission behavior, permissions, or launch any benchmark job.

### DIFF
--- a/scripts/skillsbench_automation_loop.py
+++ b/scripts/skillsbench_automation_loop.py
@@ -369,6 +369,15 @@ def _benchflow_rollout_planes_class(module: Any) -> type[Any] | None:
     return klass if isinstance(klass, type) else None
 
 
+def _benchflow_environment_name(
+    call_args: tuple[Any, ...],
+    call_kwargs: dict[str, Any],
+) -> str:
+    if call_args:
+        return str(call_args[0])
+    return str(call_kwargs.get("environment") or "")
+
+
 DOCKER_APP_SKILLS_MOUNT_BEGIN = "# BEGIN LOOPX_SKILLSBENCH_APP_SKILLS_MOUNT"
 DOCKER_APP_SKILLS_MOUNT_END = "# END LOOPX_SKILLSBENCH_APP_SKILLS_MOUNT"
 DOCKER_APP_SKILLS_MOUNT_KEEP_FILE = "loopx_app_skills_keep"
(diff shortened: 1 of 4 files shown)


huangruiteng:
## Self-review

**Assumptions:** The terminal closeout intentionally defers to an existing benchmark compact when one is present; without one it emits a typed, non-countable exclusion and permits cohort rotation.

**Product and architecture judgment:** The motivation is to prevent a BenchFlow keyword-call mismatch from silently disabling the source mount and to ensure a bounded supervisor termination always produces a public-safe terminal decision. The change solves both observed gaps at their existing ownership boundaries: environment creation and supervisor lifecycle. Operators get deterministic cleanup and closeout instead of an indefinitely live or evidence-less case. The main risk is benchmark-runtime compatibility, contained by preserving the prior public checkpoint and by avoiding scoring, task, permission, submission, and launch behavior. This is a focused extension of the existing runner contracts rather than a new abstraction.

**Validation:**

- focused runtime and supervisor tests: 17 passed
- reverse-tunnel supervisor smoke: passed
- SkillsBench benchmark-run smoke: passed
- premerge canary: changed-Python compile and all 4 selected catalog canaries passed
- diff hygiene and diff-only public/private boundary scan: clean
- no live benchmark was launched

**Manual hold:** `loopx canary premerge --from-git-diff` retained the expected `benchmark_sensitive` manual-review hold. Owner standing authorization explicitly permits `benchmark_code_self_merge` and `benchmark_pr_self_merge`. The focused tests directly cover keyword environment mount recognition, runtime SIGTERM cleanup, active-phase terminalization, and compact closeout emission; this is sufficient for the narrow helper/runtime scope.

**Merge decision:** Approved for authorized self-merge.

PR #2650, opened 2026-07-29
fix(start-goal): keep the requested worktree route

huangruiteng:
## Motivation

Real Ark Managed Agent host parity runs started a fresh task from a linked worktree, but `start-goal --project .` silently reused an older goal registered by another worktree sharing the same Git common directory. The agent noticed the path mismatch, but the guided packet itself was wrong.

## Change

- keep the exact requested project route for `start-goal`
- retain canonical linked-worktree aliasing for the lower-level `bootstrap-command-pack` diagnostic path
- update the LoopX project skill and protocol/course documentation to state the boundary
- add a real linked-worktree regression that proves an old global goal is not selected

## Validation

- `python -m pytest -q tests/control_plane/test_start_goal_compact_projection.py tests/control_plane/test_actual_default_model_behavior_portfolio.py` — 20 passed
- `python examples/bootstrap-command-pack-smoke.py` — passed
- Ruff on changed Python files — passed
- `loopx canary premerge --from-git-diff` — 18 selected checks passed, 0 failures, 0 manual holds
- public/private boundary scan — passed

## Risk

The behavior change is limited to guided goal start. Existing diagnostic bootstrap callers keep their canonical alias behavior, and the existing linked-worktree smoke continues to cover that contract.


huangruiteng:
Rebased onto current `main` and updated the single commit at `c59c9450`.

Additional review/validation after the real linked-worktree incident:
- preserved exact `start-goal` routing for the fresh worktree while lower-level diagnostics retain canonical aliasing
- added the public-safe `linked_worktree_goal_route_reuse` self-repair pattern
- real route probe returned the requested worktree and its default goal id with no old-goal selection gate
- `20 passed` in the focused projection/default-model portfolio
- bootstrap command-pack smoke passed
- premerge canary passed `18/18`, with `0` failures and `0` manual holds

One unrelated latest-main baseline lint remains outside this PR scope: Ruff PIE810 at `loopx/bootstrap_command_pack.py:84`; the premerge Python compile and all selected canaries pass.

huangruiteng:
## Findings

No blocking findings on head `b21ba57d`.

## Open questions

None. The route split is explicit: guided `start-goal --project` preserves the requested worktree, while the lower-level diagnostic command pack keeps canonical linked-worktree aliasing.

This fixes a real identity error rather than a display-only mismatch. Previously, a fresh linked worktree could share a Git common directory with an older registered worktree and inherit that older project/Goal during guided startup. The new flag is applied only at the two guided-start call sites; the existing diagnostic default stays canonical.

The volume is justified. Production behavior is `+31/-7` in one module; most of the remaining diff is the real Git-worktree regression (`+94/-1`) and protocol/course/skill alignment. I found no safe simplification that would preserve both route contracts and the regression evidence. A second wrapper or route enum would add more structure without removing current complexity.

## Validation

- focused projection/default-model tests: `20 passed`
- bootstrap command-pack smoke: passed
- risk-based premerge canary on latest main: `18/18` selected checks passed
- GitHub pytest: passed

## Merge decision

Self-merge approved: single-purpose runtime fix, no permission or destructive behavior change, no manual holds, and both exact-route and canonical-route behavior are covered.

PR #2656, opened 2026-07-30
fix(skillsbench): gate scoring on lifecycle parity

huangruiteng:
## Summary
- add a task-free, scored-equivalent ACP lifecycle canary that reuses the scored install and case-state seed path
- require `agent_active` plus one case-local LoopX state read/write under a strict 300-second maximum budget
- gate every live scored launch on a cleanup-complete, revision-pinned public receipt

## Validation
- `python -m mypy`
- 125 focused SkillsBench tests
- `python examples/skillsbench-benchmark-run-smoke.py`
- six risk-selected premerge catalog canaries
- changed-file Ruff checks and public/private boundary scan

No benchmark task or verifier was launched.

### DIFF
--- a/loopx/benchmark_adapters/skillsbench_setup_preflight.py
+++ b/loopx/benchmark_adapters/skillsbench_setup_preflight.py
@@ -10,7 +10,11 @@ from .skillsbench_failure_signals import (
 )
 
 
-SCHEMA_VERSION = "skillsbench_setup_only_public_preflight_v0"
+SCHEMA_VERSION = "skillsbench_setup_only_public_preflight_v1"
+SCORED_LIFECYCLE_READINESS_GATE_SCHEMA_VERSION = (
+    "skillsbench_scored_lifecycle_readiness_gate_v0"
+)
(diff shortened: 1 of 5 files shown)


huangruiteng:
Self-review complete.

Changed surfaces: public SkillsBench setup preflight contract, task-free ACP lifecycle probe, live launcher receipt gate, and focused regression coverage.

Validation: strict mypy passed; 125 focused tests passed; the public SkillsBench runner smoke passed; all six risk-selected premerge catalog canaries passed; changed static Python files passed Ruff; shell/Python syntax and `git diff --check` passed; public/private boundary scan was clean. The large dynamic runner retains pre-existing Ruff findings, so no broad mechanical rewrite was applied.

The premerge command reports its expected benchmark-sensitive manual hold, with zero validation failures and a valid exact-scope change-quality receipt (`cqr_2a3f235a5918c8d281ca`). Standing LoopX decisions approve both `benchmark_code_self_merge` and `benchmark_pr_self_merge` for this agent. Scope does not alter scoring or task semantics and launches no benchmark task, so the focused coverage is sufficient for the authorized helper/runtime self-merge path.

PR #2645, opened 2026-07-29
feat(host): add Ark Managed Agent LoopX entry skill

huangruiteng:
## Summary

- install a generated `$loopx` entry skill through the fixed installer, with optional exact `ark-managed-agent` host binding
- let `$loopx <task>` create business todos before one-shot Goal submission while preserving host, task text, and capabilities across bootstrap replay
- make intermediate todo writeback a generic transition invariant so downstream scope, acceptance, and non-goals survive compaction
- document and validate the host/install lifecycle without exposing implementation-specific source paths

## Validation

- 66 focused pytest cases
- `examples/install-local-smoke.py`
- `examples/control_plane/agent-onboard-host-loop-activation-smoke.py`
- `examples/release/local-install-promotion-boundary-smoke.py`
- Ruff and `git diff --check`
- `loopx canary premerge --from-git-diff`: 18/18 selected checks passed, no warnings, holds, or public-boundary findings

## Review note

This is ready for maintainer review and intentionally not self-merged.

PR #2657, opened 2026-07-30
fix(install): verify fixed-installer skill readback

huangruiteng:
## What changed

- Verify the fixed installer manifest, required skill presence, source revision, and per-skill content digests.
- Make Ark Managed Agent onboarding and doctor project the filesystem readback without becoming a second installer.
- Include the generated loopx entry plus the four packaged workflow skills in the required set.
- Preserve the exact host surface in the no-clone install command and document the ownership boundary.

## Why

The already-reviewed #2625 stack was merged into its parent branch rather than main. After #2624 landed, the readback changes remained absent from main, so host checks could describe expected skills without proving the exact installed revision and contents. Rebasing the commits also exposed a newer compatibility requirement: #2645 generates a host-bound loopx entry skill, so readback and install guidance must cover that fifth skill.

The installer remains the sole filesystem writer. Onboarding and doctor are read-only verifiers, and filesystem materialization remains distinct from runtime loaded-skill readback.

## Validation

- Focused pytest: 29 passed.
- Agent onboarding host activation smoke: passed.
- Standard premerge canary: 13/13 selected checks passed; public boundary passed; no manual holds.
- Full pytest: 1622 passed, 2 skipped, 3 environment failures also reproduced on clean main. Two subprocess-import cases pass with an absolute PYTHONPATH; the remaining case requires unavailable system python3.11-venv/ensurepip.

### DIFF
--- a/docs/reference/protocols/host-integration-surface-v0.md
+++ b/docs/reference/protocols/host-integration-surface-v0.md
@@ -95,10 +95,12 @@ target, a canary-only install leaves the existing default skill root unchanged.
 
 The installer is the sole owner of filesystem mutation. The manifest records
 the materialized skill ids, source revision, and per-skill content digests so
-read-only host checks can verify delivery without becoming a second installer.
-Filesystem materialization is distinct from the host's runtime loaded-skill
-readback; the latter is still required before claiming that the skills were
-injected into an active agent context.
+onboarding and doctor can verify delivery read-only without becoming a second
+installer. Run both with the same `LOOPX_SKILLS_DIR`; they report the
+filesystem readback status and source revision. Filesystem materialization is
+distinct from the host's runtime loaded-skill readback; the latter is still
+required before claiming that the skills were injected into an active agent
+context.
 
 Local-development and cloud transports must send the exact same `task_body` as
 their goal prompt. They may differ in endpoint, authentication, session id, or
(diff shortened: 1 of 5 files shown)


huangruiteng:
## Findings

The original head had one false-ready path: it required `source.revision` to be non-empty but did not compare it with the active LoopX CLI revision. That allowed an intact old skill install, or an arbitrary non-empty manifest revision, to report `ready_for_host_load`.

Resolved on head `cd328004`:

- archive installs now persist `LOOPX_RESOLVED_SOURCE_GIT_COMMIT` into the skill readback manifest;
- onboarding and doctor compare the installed revision with the current package/release source revision;
- the verifier rejects source drift as `source_revision_mismatch`;
- the aggregate manifest digest is checked in addition to every required skill tree digest.

No remaining blocking findings.

## Open questions

None. Runtime loaded-skill readback remains deliberately separate; this PR proves filesystem materialization and revision/content integrity only.

The ownership boundary is sound: the fixed installer remains the only filesystem writer, while onboarding and doctor are read-only projections of one shared verifier. The change does not introduce a second installer or silently repair files.

The full diff has about 310 production additions, concentrated in the readback module plus two thin callers; tests contribute about 150 additions. That volume is necessary for the manifest parser, negative states, two host projections, and regression coverage. The highest-value simplification is already present: all manifest/revision/digest rules live in `skill_install_readback.py`, and callers only supply the source root.

## Validation

- focused Ark host tests: `12 passed`
- Ark onboarding/activation smoke: passed
- Ruff and changed-file compile: passed
- risk-based premerge canary on latest main: `13/13` selected checks passed, including local install, packaged install, and update smokes
- GitHub pytest: passed

## Merge decision

Self-merge approved after refinement: the filesystem readback now forms a fail-closed revision/content verification loop, with no manual holds and no change to installer authority.

PR #2668, opened 2026-07-30
test(quota): isolate fixture git config

huangruiteng:
## Summary

- isolate the quota workspace smoke's temporary repositories from user-global Git integrations
- prevent asynchronous Trace2 consumers from recreating `.git/ai` during `TemporaryDirectory` cleanup
- keep strict cleanup behavior instead of ignoring deletion failures

## Validation

- direct quota workspace smoke: 20/20 passed
- parent peer-agent runtime suite: 5/5 passed
- Ruff, mypy, py_compile, `git diff --check`: passed
- public boundary: 1 path, 0 errors, 0 warnings
- risk-based premerge: 5/5 selected checks passed, 0 holds
- Change Quality: `cqr_23f9abcfc6589a076af4`, exact diff verified

## Risk

Test-only environment isolation. No runtime, quota policy, persisted-state, benchmark, or release behavior changes.

### DIFF
--- a/examples/control_plane/quota-spend-workspace-causality-smoke.py
+++ b/examples/control_plane/quota-spend-workspace-causality-smoke.py
@@ -27,6 +27,10 @@ GOAL_ID = "quota-spend-workspace-causality"
 AGENT_ID = "codex-peer"
 DELIVERY_REPOSITORY = "git:example.invalid/loopx/delivery"
 
+# Keep fixture repositories isolated from user-global Git integrations that
+# may write into .git asynchronously during temporary-directory cleanup.
+os.environ["GIT_CONFIG_GLOBAL"] = os.devnull
+
 
 def run_git(cwd: Path, *args: str) -> None:
     subprocess.run(


huangruiteng:
Self-review / premerge evidence:

- Changed surface: one public control-plane smoke; no product runtime or protocol changes.
- Root cause: user-global Git Trace2 integrations can asynchronously create `.git/ai` in temporary fixture repositories while Python removes them.
- Simplification: one process-scoped `GIT_CONFIG_GLOBAL=os.devnull` override removes the external writer at the Git configuration boundary; no cleanup retry or ignored error path was added.
- Focused checks: direct smoke 20/20; parent peer-agent runtime suite 5/5; Ruff, mypy, py_compile, and diff check passed.
- Public/private boundary: 1 changed path scanned, 0 errors, 0 warnings.
- Risk-based premerge: 5 selected / 5 executed / 0 failures / 0 warnings / 0 manual holds.
- Change Quality: exact receipt `cqr_23f9abcfc6589a076af4` verified, no risks.
- Failures/skips: none after the fix. The pre-fix cleanup race was reproduced and attributed to external Trace2 writes.
- Merge decision: self-merge is sufficient because this is a 4-line, test-only environment isolation with no runtime, quota policy, benchmark, permission, or release behavior change.

PR #2669, opened 2026-07-30
fix(issue-fix): qualify negative candidate evidence

huangruiteng:
## Motivation

Issue-fix candidate screening previously treated the presence of `numeric_pr_evidence` and `semantic_pr_evidence` as sufficient. Naked empty arrays or a capped aggregate PR index could therefore be mistaken for proof that no prior implementation exists.

A live ark-managed-agent Goal then exposed a discoverability gap in the first revision: the fail-closed validation worked, but the packet did not tell the caller that each field expects one receipt object or which exact query scope to use.

## Approach

- Keep provider queries outside the LoopX core.
- Require each evidence input to be one issue-specific, complete, non-truncated receipt with repository, issue, exact query scope, and rows.
- Project the complete receipt schema and canonical scopes into the compact start-goal packet.
- Return an actionable error for list-shaped receipts.
- Allow an empty result to reach `proceed` only when both receipts satisfy that contract.
- Keep capability details in issue-fix state; generic Todo fields are unchanged.

## Validation

- Live callsite: rejected list-shaped evidence before any proceed decision; the follow-up now exposes cardinality and canonical scopes in the same packet.
- `python -m pytest -q` — 1639 passed, 2 skipped.
- `python -m loopx.cli canary premerge --from-git-diff` — 18/18 passed.
- Focused Ruff checks and `git diff --check` — passed.

## Risk

This intentionally tightens the unmerged v0 candidate-preflight input shape. Callers that supply naked evidence arrays must wrap them in one complete query receipt. Unconfigured non-issue-fix workflows are unchanged.


huangruiteng:
Superseded by #2673, which is now merged as `207258be4d0b95262c8c0bac60387a46adda75dd`.

#2673 retains this PR's strict negative-evidence requirement and completes the surrounding contract:

- all closing-PR, cross-reference, and maintainer-comment receipts are required;
- malformed, incomplete, truncated, wrong-scope, or revision-less rows fail closed;
- evidence collection, verification, admission, and terminal disposition are separate states;
- PR and maintainer-comment resolutions bind to current source revisions;
- pending cases project concrete recovery successors instead of a final-looking `comment_only`.

Closing this narrower PR without merging it separately.

PR #2699, opened 2026-07-31
fix(skill): continue active goal before bootstrap

huangruiteng:
## Motivation

The generated LoopX entry skill treated empty arguments as a bootstrap request even when the host task had already activated a Goal. A later turn could therefore re-enter start/bootstrap instead of continuing the existing Goal contract.

## Change

- prefer the active host task exact interaction contract or quota command for empty-argument continuation;
- reserve bootstrap inspection for sessions without an active Goal contract;
- cover generic, fixed-host, and Codex-installed skill materialization;
- extend the existing skill/CLI contract drift repair pattern.

## Validation

- 30 focused pytest cases passed;
- LoopX premerge canary passed: 7/7 selected catalog checks;
- git diff --check passed.

The PR changes only generated skill guidance and its focused tests; it does not add a new Goal lifecycle or host-specific runtime state.


huangruiteng:
Independent qualification complete at exact head `e65d09ad5c21b99977bdd4798804b79dd05dd73e`: no blocking finding.

Implementation check:
- The change updates the existing `_command_prompt_specs` source of truth rather than adding a second host-specific state machine. Empty-argument `/loopx` now distinguishes an already activated Goal from a first-session bootstrap: active tasks follow the projected CLI interaction/quota contract first, while sessions without an active Goal retain the existing bootstrap/status/slash-command discovery path.
- The same generated instruction is exercised by generic materialization, fixed managed-agent materialization, and Codex managed-skill upgrade tests. The adjacent self-repair pattern is updated to classify host-lifecycle drift alongside CLI syntax drift.
- No parser, quota, Goal lifecycle, permission, scheduler, or runtime-state behavior changes. Latest `origin/main` merges cleanly with the exact PR head, and the synthetic merge result preserves the focused behavior.

Validation:
- focused exact-head pytest: 30/30 passed (`test_slash_command_install.py` + `test_ark_managed_agent_host.py`);
- latest-main synthetic merge-result pytest: the same 30/30 passed;
- GitHub `pytest`: passed;
- Ruff check: passed; mypy: 15 source files passed; changed-file `py_compile`: passed;
- public/private boundary scan: 3 changed files, 0 errors, 0 warnings; `git diff --check`: passed;
- authoritative `loopx canary premerge --from-git-diff --goal-id loopx-meta`: passed, 4 direct checks plus 7/7 selected catalog canaries, 0 failures, 0 warnings, 0 advisory failures, 0 manual holds;
- exact-scope change-quality receipt `cqr_fe43e523d3582084056a` is valid for fingerprint `fe43e523d3582084056a487b7ece55d16e992dfe86eca605ae6f0169b9753a5a`; no safe-fix pass was needed.

Coverage is sufficient because this PR is a three-file generated-guidance correction: the tests verify every affected materialization surface, the synthetic merge test proves compatibility with current main, and premerge covers packaged install/update plus replan/writeback regressions. I found no reason to hold merge.

PR #2711, opened 2026-07-31
fix(control-plane): project delivery workspace causality

huangruiteng:
## What changed

- project a compact `delivery_workspace_causality_v0` contract for spend-capable selected Todos that declare `task_repository`;
- preserve that causality in the TurnEnvelope action signature;
- bind `turn run-once` refresh and spend to the host project only when the envelope carries that causal contract;
- let quota spend validate an explicit workspace without changing process-global cwd;
- preserve the existing fail-closed workspace guard and keep the default quota hot path unchanged;
- document the contract and add the corresponding self-repair pattern.

## Root cause

Quota accounting already records the workspace that produced a delivery and correctly rejects spend attempts from another checkout. The interaction contract exposed the causal rule, but the TurnEnvelope dropped it before the host could consume it. The active `turn run-once` writeback path therefore still refreshed and spent relative to controller context.

## Validation

- rebased onto the latest `main` without conflicts;
- 66 focused interaction, envelope, and turn-driver tests passed;
- full suite: 1783 passed, 2 skipped;
- quota workspace causality and integrated control-plane smokes passed;
- hot-path interface budget passed with the default quota payload unchanged at 12,547/12,550 characters;
- diff-driven standard premerge canary passed: 18/18 checks, no failures, warnings, or manual holds;
- public/private boundary scan passed for all 10 changed files.


huangruiteng:
## Self-review: no remaining blockers

### Findings

No actionable correctness, compatibility, or public-boundary findings remain on head `ce91ca7f`.

The refinement closes the missing active-path link: `delivery_workspace_causality` now survives the TurnEnvelope and its action signature, and `turn run-once` passes the host project to refresh/spend only when that causal contract is present. Default turns retain the prior cwd-based behavior. Quota spend gains one optional workspace input and still delegates validation to the existing fail-closed guard.

- **Motivation:** make workspace accountability machine-consumable so a host does not discover the recorded delivery workspace through failed spend attempts.
- **Solved:** the contract now reaches the actual writeback consumer and is covered by a causal Git checkout fixture.
- **User/operator impact:** long-running hosts can refresh and spend against the workspace that produced the delivery while controller cwd remains independent.
- **Main risk:** changing default quota payload size or silently blessing an unrelated checkout. The field is conditional on a selected Todo with `task_repository`, and the existing guard remains authoritative.
- **Design:** the change reuses the existing interaction, envelope, refresh, and quota guard contracts; it does not add a parallel state machine.

### Validation

- Rebased onto latest `main` without conflicts.
- Focused tests: 66 passed.
- Full suite in an isolated test environment: 1783 passed, 2 skipped.
- GitHub Actions `pytest`: passed on the final head.
- Diff-driven standard premerge gate: 18/18 checks passed, including Python compile, 9 catalog canaries, 8 risk-profile smokes, diff hygiene, and public/private boundary scan.
- Dedicated quota workspace causality smoke, integrated control-plane canary, and hot-path interface budget passed.
- Failures: none on the final environment. The first local full-suite attempt stopped during collection because the ambient interpreter lacked the declared `jsonschema` test dependency; rerunning with the repository-declared test dependencies completed successfully.
- Manual holds: none. Skips: 2 from the existing full suite.

Coverage is sufficient because it checks projection/signature integrity, the built-in turn host's real refresh/spend callbacks, explicit-path guard behavior with an independent controller cwd, default-path compatibility, integrated control-plane behavior, and the full regression suite.

### Merge decision

Self-merge is authorized by the owner and allowed by the premerge gate. Proceeding with squash merge.

PR #2578, opened 2026-07-27
feat: add planner-worker mode

maxliux5:
# Summary

- Add a provider-neutral planner-worker contract with strict JSON validation, readiness and dependency selection, executor routing, scoped worker prompts, validation, and execution receipts.
- Add one executable Planner -> Worker -> validation runtime slice that rejects Planner writes, unapproved validation commands, out-of-contract file changes, symlink escapes, and unsupported workspace entry types.
- Add an experimental TraeX provider with a read-only Planner, workspace-write Worker, bounded validation subprocesses, minimal Worker context, and focused contract/runtime/provider smokes.
- Record workspace changes as structured `{size, file_type}` entries, reject non-regular paths even when extra regular files are allowed, and audit the final workspace again after validation.

## Issue Or Task

- Closes #issue-a7rp3qjzux6hy
- Contributor task ID: issue-a7rp3qjzux6hy

## Validation

- [x] `python -m pytest -q tests/test_planner_worker.py tests/test_planner_worker_runtime.py` (`41 passed`)
- [x] `python examples/planner-worker-contract-smoke.py`
- [x] `python examples/planner-worker-runtime-smoke.py`
- [x] `python examples/traex-planner-worker-probe-smoke.py`
- [x] `ruff check loopx/planner_worker_runtime.py loopx/extensions/traex_planner_worker.py tests/test_planner_worker_runtime.py`
- [x] `mypy loopx/planner_worker_runtime.py loopx/extensions/traex_planner_worker.py`
- [x] `python -m py_compile loopx/planner_worker_runtime.py loopx/extensions/traex_planner_worker.py`
- [x] `loopx canary premerge --from-git-diff` (`standard`, 10 selected checks passed, 0 failures, 0 manual holds)

## Boundary Checklist

- [x] I did not commit `.loopx/`, `.codex/goals/`, live `ACTIVE_GOAL_STATE.md`, credentials, private benchmark traces, verifier output, raw agent sessions, internal document links, or local machine paths.
- [x] I did not duplicate maintainer-owned benchmark work unless a maintainer split out a public issue for it.
- [x] I kept the change scoped to the linked issue/task.

PR #2718, opened 2026-08-01
refactor(experiments): isolate planner-worker prototype

huangruiteng:
## Summary

- add a documented `loopx/experiments/<id>` placement and promotion contract
- move planner-worker contract, runtime, TraeX adapter, tests, examples, and probe into matching experiment paths
- enforce the one-way dependency rule: experiments may use core, but core cannot import experiments
- keep historical change-quality fixtures unchanged because they describe PR #2578 at its original paths

No compatibility wrappers are retained: planner-worker is explicitly experimental, has no registered CLI/capability/default runtime caller, and the old modules were not a stable product contract.

## Validation

- `68 passed` focused planner-worker, architecture, and historical shadow tests on latest `origin/main`
- `1784 passed, 2 skipped` full pytest before a disjoint main rebase; focused tests and premerge gate rerun after rebase
- planner-worker contract, runtime, and TraeX probe smokes passed
- Ruff and strict Mypy passed for moved surfaces
- `loopx canary premerge --from-git-diff --tier standard`: passed 18 selected checks, no warnings or manual holds

## Risk

This is a behavior-preserving package move plus an architecture guard. It does not register a CLI command, capability, scheduler route, or default provider.


huangruiteng:
No blocking findings.

**Motivation:** planner-worker is an opt-in prototype, but its contract and runtime were placed beside stable core modules while its TraeX adapter occupied the registered extension namespace. That made an experimental surface look more mature and more coupled than it is.

**Solved:** this PR gives executable experiments one clear home and moves the whole planner-worker unit there: contract, runtime, provider adapter, tests, examples, and manual probe. The architecture test enforces the important dependency direction: experiments may depend on stable LoopX code, while stable LoopX code cannot import experiments.

**Operator impact:** there is no new command or default behavior. Developers can still run the same bounded smokes and explicit TraeX probe from discoverable experiment paths. Promotion and removal criteria are now documented, so future prototypes do not need to invent placement rules.

**Main risk:** import compatibility for anyone who adopted the old Python modules directly. I consider that acceptable here because planner-worker was just introduced as experimental, has no registered CLI/capability/default runtime caller, and promised no stable import contract. Keeping wrappers would create exactly the second long-lived route this cleanup is intended to avoid.

## Changed surfaces

- `loopx/experiments/planner_worker/`: moved contract, runtime, and TraeX adapter
- matching `tests/experiments/`, `examples/experiments/`, and `scripts/experiments/` paths
- experiment placement, promotion, and removal policy
- package architecture guard against core-to-experiment imports
- contributor task smoke paths

## Validation

- focused tests on the exact rebased head: 68 passed
- full suite before a disjoint main rebase: 1784 passed, 2 skipped
- planner-worker contract, runtime, and TraeX probe smokes: passed
- Ruff and strict Mypy on moved surfaces: passed
- `loopx canary premerge --from-git-diff --tier standard`: 18 selected checks passed, 0 warnings

## Failures, skips, and holds

- An initial plain-Python full run stopped at collection because the active environment lacked declared test dependencies. The first isolated `uv --no-project` run then exposed two subprocess import-path failures caused by that temporary environment. Re-running with the declared dependencies and repository `PYTHONPATH` produced the full passing result above.
- Two repository tests were skipped by their existing conditions.
- Manual holds: none. The premerge gate reports `self_merge_allowed=true`.

## Merge decision

Self-merge after the remote pytest check succeeds and the PR head still equals the reviewed head.

PR #2729, opened 2026-08-01
fix(quota): reject watch-only safe-bypass spend

huangruiteng:
## Problem

An unchanged watch/readback turn could rewrite a user gate, monitor target, or durable Next Action, label that replan as outcome progress, and then consume quota through a safe-bypass path. The state acknowledgement was real, but it was not an accountable delivery.

## Change

- distinguish replan acknowledgement from quota-accountable frontier progress
- normalize watch-only replan outcomes to `outcome_gap`
- require the latest unspent accountable delivery before every safe-bypass spend
- bind the spend-event builder to the validated preview
- cover operator gate, recovery delivery, and outcome-floor recovery with one exactly-once matrix
- keep the heartbeat smoke at zero spend after watch-only refreshes

The rule remains in the existing quota ledger owner; no new state, receipt source, or abstraction is introduced.

## Validation

- focused pytest: 59 passed
- Ruff: passed on all changed Python surfaces
- mypy: passed on 15 repository-declared strict targets
- quota-plan smoke: passed
- heartbeat quota flow smoke: passed
- premerge canary: 18/18 passed, 0 manual holds
- exact change-quality receipt: `cqr_abfc6a0cc87d29ffacce`, valid

## Merge Decision

Owner-authorized self-merge after exact-head validation. The final change is narrow, public-safe, and fully covered across all supported safe-bypass modes.


huangruiteng:
No remaining blockers after refinement.

The operator-only check is now a single causal rule for every safe-bypass mode: a spend must bind the latest unconsumed accountable delivery. One parameterized matrix covers operator gate, recovery delivery, and outcome-floor recovery for missing delivery, first consumption, and duplicate rejection. The live heartbeat smoke separately proves that watch-only refreshes stay at `outcome_gap` and leave spend count at zero.

Validation on exact head `ed9f1782`: focused pytest 59 passed; Ruff passed; mypy passed; quota-plan and heartbeat quota-flow smokes passed; premerge canary passed 18/18 with no manual holds; public/private scan passed. Exact receipt `cqr_abfc6a0cc87d29ffacce` verifies as valid.

Architecture judgment: this keeps lifecycle ownership in the existing quota ledger and replan-delta contracts. It introduces no parallel receipt source or new state machine, and replaces three possible exception paths with one exactly-once rule.

Decision: approved for the owner-authorized self-merge path.

PR #2738, opened 2026-08-02
fix(quota): retain selected todo continuation context

huangruiteng:
## Problem

A compact quota turn can select the correct runnable Todo while dropping that Todo's latest continuation note. When an older durable `Next Action` is still projected beside the selected Todo, a resumed agent can follow the obsolete prose even after the current Todo and `recommended_action` are correct.

## Change

- derive one bounded, secret-screened `continuation_hint` for the selected open advancement Todo
- preserve the hint through agent-lane and selected-Todo projections without copying raw evidence
- when the selected Todo supersedes durable action prose, keep it as the only executable hot-path action and move the shadowed fields behind `--include-detail agent-todos`
- document the self-repair pattern and cover positive propagation, secret rejection, output budgets, compact CLI routing, and diagnostic readback

## Validation

- initial focused suite: 27 pytest cases passed
- follow-up shadowing suite: 13 pytest cases passed
- Ruff passed on changed Python files
- real quota readback selected the expected Todo, retained its continuation, and omitted shadowed action prose from the default packet while preserving it on the detail path
- `cli-output-budget-regression-smoke.py` passed after removing the redundant hot-path diagnostic object
- `loopx canary premerge --from-git-diff`: 17/17 selected checks passed, no warnings, no manual holds
- `git diff --check` passed


huangruiteng:
Self-review/refine complete on `966252ad`.

## Findings addressed

- **[P1] Explicit `continuation_hint` could bypass the bounded text sanitizer.** The raw field was copied into compact Todo/selected-Todo projections before the derived sanitizer ran, so a rejected value could survive. The refine removes that duplicate path: only sanitized hints from open advancement Todos are projected, and the selected-Todo boundary screens the value again.
- Expanded inline credential screening to cover API keys, authorization/bearer forms, access-key ids, secrets, tokens, and passwords. Negative tests cover both note-derived and explicit hints, plus non-advancement Todos.
- Reconciled the branch with current `main` and kept selected Todo identity as the condition for suppressing stale durable action prose. Detailed/cold-path output still preserves those diagnostics.
- Extended the existing actual-default model portfolio rather than adding a parallel harness. Its selected-Todo invariance case now carries a continuation hint while stale durable actions are omitted from the compact packet.

## Validation

- Focused pytest: **51 passed**
- Ruff on changed Python surfaces: passed
- Repository mypy config: **15 source files passed**
- `cli-output-budget-regression-smoke`: passed
- `monitor-poll-writeback-smoke`: passed
- Live Doubao actual-default portfolio: **15/15 scenarios**, two calls each, and **4/4 contrasts** passed
- `loopx check` public/private scan: **0 errors, 0 warnings**
- Exact-scope change-quality receipt: `cqr_f3b71911240e785b1228`, valid
- Risk-based premerge canary: **17/17 selected checks passed**, no failures, skips, or manual holds

No remaining review blockers. Coverage exercises the model-facing continuation/stale-action distinction as well as deterministic projection and security boundaries, so this head is ready to merge.

PR #2744, opened 2026-08-02
test(replan): keep bounded monitor fixture unexpired

huangruiteng:
## Problem

The autonomous replan smoke used a bounded monitor expiry fixed at 2026-08-02T13:00:00Z. Once wall clock time passed that instant, the positive fixture became an expired monitor and the main-branch premerge gate failed deterministically.

## Change

Move only the positive fixture expiry to a stable far-future instant. The negative fixture still omits expiry and continues to prove fail-closed behavior.

## Validation

- autonomous-replan-obligation smoke passed
- premerge canary passed all 5 selected checks and allowed self-merge
- git diff check passed

The file has pre-existing repository Ruff findings around executable mode and import layout; this one-line fixture correction does not touch them.

### DIFF
--- a/examples/autonomous-replan-obligation-smoke.py
+++ b/examples/autonomous-replan-obligation-smoke.py
@@ -651,7 +651,7 @@ def assert_watch_replan_ack_requires_bounded_state_evidence() -> None:
         state_path.write_text(
             state_path.read_text(encoding="utf-8").replace(
                 "next_due_at=2026-08-01T13:00:00Z",
-                "next_due_at=2026-08-01T13:00:00Z expires_at=2026-08-02T13:00:00Z",
+                "next_due_at=2026-08-01T13:00:00Z expires_at=2099-08-02T13:00:00Z",
             ),
             encoding="utf-8",
         )


huangruiteng:
No blocking findings.

- **Motivation:** the positive autonomous-replan smoke encoded a bounded monitor whose fixed expiry has now elapsed, so the fixture flipped from valid continuation evidence to an expired monitor and began failing against unchanged product behavior.
- **Solved:** yes. The one-line change restores the intended positive state while the adjacent negative case still proves that a monitor without an expiry is rejected.
- **Design judgment:** using the repository's established far-future fixture convention is appropriately narrow here. Introducing a clock abstraction into this end-to-end smoke would be more machinery than the regression warrants.
- **Risk:** limited to a public validation fixture; no runtime, protocol, permission, or production-state behavior changes.

**Validation on exact head `244dcd89`**

- `uv run python examples/autonomous-replan-obligation-smoke.py`: passed
- `git diff --check origin/main...HEAD`: passed
- `uv run loopx canary premerge --from-git-diff`: passed, 5/5 selected checks, no manual holds, self-merge allowed
- public/private boundary scan selected by premerge: passed

Decision: self-merge is appropriate for this single-purpose fixture repair.

PR #2745, opened 2026-08-02
fix(integration): observe published integration head

huangruiteng:
## Problem

Remote refresh observed configured base and source refs, but not the integration branch upstream itself. Another agent could advance the published integration branch while status still reported in_sync; the conflict surfaced only as a non-fast-forward push. Tree-equivalent candidates could also omit the published ancestry.

## Change

- narrow-fetch the configured integration upstream together with base and source refs
- expose remote-ahead and diverged ancestry as integration_remote_head_moved drift
- reconcile that exact remote head into automatic candidates
- require supplied candidates to contain the observed remote head
- retain no-prune, no-FETCH_HEAD-write, no-push, and no-force-update boundaries
- register the reusable self-repair pattern and document the contract

## Validation

- integration branch focused tests: 14 passed, including remote-ahead and same-tree diverged cases
- Ruff on changed Python modules and tests passed
- premerge canary passed all 17 selected checks with no manual holds
- public boundary scan passed
- git diff check passed

This PR is stacked on #2744 only because that one-line main-branch smoke expiry fix is required for the premerge gate. After #2744 lands, this PR can target main without code changes.


huangruiteng:
## Independent refinement review

No blocking findings on final head `7a0e007191208f3c3fe46ef938a0ed719d53cf6a`.

### Motivation and contract

The change closes the real stale-observation gap where another agent can advance the published integration branch while the local branch remains unchanged and status still reports `in_sync`. It now:

- discovers the integration branch's exact configured upstream and adds only that ref to the existing narrow fetch set;
- distinguishes remote-ahead and diverged ancestry as `integration_remote_head_moved`;
- requires both generated and supplied reconciliation candidates to contain the observed remote head;
- preserves the existing read-only status/fetch boundary and the existing build/push path.

### Scope judgment

The code is cohesive and right-sized for the shipped behavior. The two small helpers centralize upstream discovery and reconciliation inputs; collapsing them would duplicate Git configuration/ancestry rules or couple status projection to candidate construction. I therefore made no speculative refactor or behavior edit. I only merged current `main` into the branch so this exact diff is qualified on the latest base.

### Validation

- focused integration-branch suite: `14 passed`
- Ruff on changed Python: passed
- Mypy kernel contracts: passed
- `git diff --check origin/main...HEAD`: passed
- public/private boundary scan over all five changed files: passed, no warnings
- exact change-quality receipt: `cqr_658afaa436a8a1946187`, valid for the final base/head fingerprint
- risk-based premerge gate: `17/17` selected checks passed, `0` manual holds, self-merge allowed
- GitHub `pytest`: passed in 4m38s

This coverage exercises remote-ahead and same-tree divergence, stale supplied-candidate rejection, final ancestry, output projection, and the repository boundary around narrow read-only fetches. It is sufficient for this deterministic Git/control-plane change; model-behavior qualification is not relevant to the changed contract.

Owner authorization covers review/refine/self-merge for this integration-discovered issue-fix flow, so I am proceeding with the validated self-merge.

PR #2747, opened 2026-08-02
fix(goal): bound issue-fix capability intent

huangruiteng:
## Problem

Goal capability selection combined independent keywords across an entire objective. A broad engineering Goal that mentioned generic fixes in one requirement and a PR lifecycle in another was incorrectly routed to issue-fix. The host correctly rejected that route because no durable issue admission binding existed, so an otherwise valid long-running Goal could not start.

## Root cause

The matcher treated action-present and target-present as two goal-wide booleans. It did not require the action and the issue or PR target to describe the same intent. Bootstrap command-pack construction also normalized multiline Goal text before capability selection, so a matcher-level newline boundary alone would not protect the real host path.

## Change

- Preserve explicit `issue-fix` selection.
- Require action and issue or PR target terms to share a bounded intent phrase.
- Treat sentence punctuation, newlines, and common coordinators as intent boundaries.
- Preserve concrete fix plus GitHub issue or PR URL routing, including `Fix: <url>`.
- Select the capability route from raw Goal text while retaining normalized text for display and command construction.
- Document the reusable false-positive repair pattern.

The bound uses phrase boundaries plus independent word and character limits: the word limit covers spaced English text, while the character limit also bounds CJK text without spaces.

## Validation

- Focused route and host matrix: 21 passed.
- Ruff: passed.
- Mypy: passed.
- Public/private scan: 0 errors, 0 warnings.
- Git diff check: passed.
- GitHub Actions pytest: passed.
- Change-quality receipt: `cqr_27925682e2515b9fad0d`, exact final scope valid.
- LoopX premerge: 9/9 catalog canaries passed, 0 failures, 0 skips, 0 manual holds.

## Scope

This is a standalone generic LoopX routing fix. It contains no host-adapter implementation and no private runtime evidence.


huangruiteng:
Self-review/refine complete for final head `73665967`.

### Findings resolved

- **P1: real multiline boundaries were lost before matching.** `build_loopx_bootstrap_command_pack` normalized Goal text before route selection, so a newline-aware matcher still misrouted the actual host path. Capability selection now consumes the raw Goal while the public display/command text remains normalized.
- **P1: the original clause rule was too narrow.** It did not separate comma- or coordinator-delimited intents such as `fix generic tests and follow the PR lifecycle`. The matcher now treats punctuation, newlines, and common English/CJK coordinators as phrase boundaries, while preserving `Fix: <issue URL>`.

### Changed surfaces

- issue-fix intent matcher
- bootstrap/start-goal capability-route projection
- focused positive/negative route tests
- reusable self-repair pattern documentation

### Validation

- focused route/host tests: **21 passed**
- Ruff and Mypy: **passed**
- public/private scan: **0 errors, 0 warnings**
- GitHub Actions pytest: **passed**
- exact change-quality receipt: **`cqr_27925682e2515b9fad0d` valid**
- `loopx canary premerge --from-git-diff --goal-id loopx-meta --timeout-seconds 600`: **9/9 passed**, 0 failures, 0 skips, 0 manual holds

The wider timeout only accommodates two known slow shared-runtime projection smokes; both completed successfully. The final diff is four files, single-purpose, contains no private evidence, and the authoritative merge gate reports `self_merge_allowed=true`. Approving for self-merge.

PR #2751, opened 2026-08-03
perf(cli): fast-path version output

Ray0907:
## Summary

- add a lightweight console entrypoint for the exact `loopx --version` path
- defer importing `loopx.cli` until another command needs the full command graph
- point the packaged `loopx` console script at the lightweight entrypoint

## Why

The existing console script imported `loopx.cli` before argparse could handle
`--version`. That eagerly loaded every command and capability module for a static
version string. Local installed-package measurements improved from roughly
657–2117 ms to 18.8–19.5 ms across five runs.

All non-fast-path arguments still delegate to the existing `loopx.cli:main`
implementation, so parser and command behavior remain centralized.

## Validation

- `pytest -q`: 1919 passed, 2 skipped
- `pytest -q tests/test_cli_entrypoint.py`: 3 passed
- `ruff check loopx/entrypoint.py tests/test_cli_entrypoint.py`: passed
- `python examples/cli-version-command-modularization-smoke.py`: passed
- `loopx canary premerge --from-git-diff --tier standard`: passed, no manual holds
- clean package install verified both `loopx --version` and JSON `loopx version`

### DIFF
--- a/loopx/entrypoint.py
+++ b/loopx/entrypoint.py
@@ -1,0 +1,16 @@
+from __future__ import annotations
+
+import sys
+
+from . import __version__
+
+
+def main(argv: list[str] | None = None) -> int:
+	raw_argv = sys.argv[1:] if argv is None else list(argv)
+	if raw_argv == ["--version"]:
+		print(f"loopx {__version__}")
+		return 0
+
+	from .cli import main as cli_main
+
+	return cli_main(raw_argv)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ test = [
 ]
 
 [project.scripts]
-loopx = "loopx.cli:main"
+loopx = "loopx.entrypoint:main"
 loopx-lark-provider = "loopx.extensions.lark.provider:main"
 loopx-openviking-semantic-preference = "loopx.extensions.openviking_semantic_preference.provider:main"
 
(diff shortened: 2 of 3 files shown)


huangruiteng (approved):
## Review result

No blocking findings.

## Open assumptions

- The fast path intentionally applies only to the exact `loopx --version` argv. All other argument shapes continue through `loopx.cli:main`, preserving parser and command behavior.
- The package console-script mapping remains the single public entrypoint contract; the structured `loopx --format json version` path must stay on the full CLI.

The change removes full command-graph import cost from a static version query without duplicating parser behavior. A thin dispatcher is the right boundary here: it improves installed CLI responsiveness while keeping every non-fast-path command centralized. The main compatibility risk is argv/output drift, and the focused tests cover both the skipped import and full-CLI delegation.

## Validation

- Focused entrypoint tests: 3 passed.
- Real console commands: `loopx --version` and structured `loopx --format json version` passed.
- CLI version modularization smoke passed.
- Premerge gate passed: changed-file compile and diff checks plus the selected maintainability canary; no failures, warnings, or manual holds.
- GitHub `pytest` check is green.

## Merge decision

Owner-authorized self-merge. The PR is narrow, validated, and has no permission, state-migration, private-evidence, or benchmark impact.

PR #2447, opened 2026-07-22
fix: make paused quota authoritative

cocolord:
## Summary
- allow `configure-goal --quota-compute 0` as the supported pause configuration
- make paused quota override workspace repair, capability bridge repair, autonomous replan, monitor, inbox, and fallback execution paths
- project a quiet no-spend interaction and scheduler contract while paused

## Validation
- `uv run --extra test pytest -q tests/control_plane/test_quota_paused_precedence.py` (4 passed)
- adjacent control-plane tests (29 passed)
- `uv run loopx canary premerge --from-git-diff` (passed, 17 checks; one unrelated inherited maintainability advisory)
- live readback against three paused goals and four registered agents: all returned `state=paused`, `decision=skip`, `should_run=false`, `effective_action=quota_skip`

## Hold
Draft for maintainer review. No goals are resumed or merged by this PR.

PR #2755, opened 2026-08-03
fix(quota): make a paused Goal an authoritative terminal decision (supersedes #2447)

cocolord:
## Why this exists (supersedes #2447)

`quota.compute=0` is a Goal-level hard pause, but the previous implementation was not authoritative. A paused Goal could still be promoted into capability or workspace repair, and `configure-goal --quota-compute 0` was rejected.

This PR makes the pause one terminal control-plane contract: no delivery, repair, monitor, inbox, or recurring host heartbeat remains runnable until quota is explicitly resumed.

## What changed

1. `--quota-compute 0` is accepted as a hard pause; negative values remain invalid.
2. The selector short-circuits paused Goals before capability, workspace, replan, monitor, user-gate, or inbox lanes are constructed.
3. Paused automation liveness now returns `keep_active=false` and permits the recurring automation to stop.
4. A typed `quota_paused` scheduler hint asks Codex App to pause or delete the current heartbeat, recommends no replacement RRULE, requires no scheduler ACK, and resumes only after `quota.compute > 0`.
5. Execution remains paused when status health is unhealthy, but the returned `ok` and `status_health_ok` fields preserve the real false health fact.
6. The quota allocation reference distinguishes a Goal-level hard pause from one agent's `monitor_only` lane.

## Review repairs

This head addresses both findings from the prior `REQUEST_CHANGES` review:

- recurring Codex App wakeups no longer survive the hard pause;
- the paused terminal projection no longer rewrites unhealthy status as healthy.

The scheduler repair lives in the existing scheduler bounded context and reuses `apply_scheduler_execution_context`; the selector-front terminal design remains unchanged.

## Verification

- exact head: `ccc4282dc1f53aa4330e805f5cbf591e62255e4f`
- paused precedence suite: 11 passed, including Codex App host-action/RRULE behavior and paused-plus-unhealthy status
- complete quota/scheduler/heartbeat test cluster: passed
- Ruff on all changed Python surfaces: passed
- `git diff --check`: passed
- `loopx canary premerge --from-git-diff`: exit 0; direct checks, compile, 9 catalog canaries, risk-profile smokes, and public-boundary checks passed
- rebased onto current `origin/main` before the maintainer repair was pushed

No production action or external scheduler mutation was performed by these tests.

### DIFF
--- a/docs/quota-allocation.md
+++ b/docs/quota-allocation.md
@@ -31,8 +31,19 @@ Examples:
   minute-slots.
 - `0.3`: 30% duty cycle, roughly 7.2 hours per day or 30% of scheduling
   minute-slots.
-- `0`: compute-paused. The goal remains visible but should not receive
-  automatic Codex turns.
+- `0`: compute-paused. This is a **Goal-level hard pause**: the goal stays
+  visible but receives no automatic Codex turns, and `quota should-run` returns
+  a single authoritative paused contract. Because the pause is a typed terminal
+  decision evaluated before any selector lane is built, no capability-bridge,
+  workspace, replan, monitor, or inbox lane can leave a contradicting execution
+  signal underneath it (`should_run=false`, all automatic permissions false,
+  `DONT_NOTIFY`, scheduler never `run_now`, no quota spend). Set it with
+  `loopx configure-goal --quota-compute 0`; negative values are rejected.
+
+  This is different from a single agent's `monitor_only` work mode.
+  `quota.compute=0` pauses the **whole Goal** for every agent and every lane;
+  `monitor_only` is a per-agent lane state where one agent still runs bounded
+  read-only monitor polls while the Goal itself remains active for other work.
 
 The number can be interpreted as either a duty cycle or a relative weight,
 depending on the executor:
--- a/loopx/cli_commands/registry_admin.py
+++ b/loopx/cli_commands/registry_admin.py
@@ -248,7 +248,7 @@ def register_registry_admin_commands(subparsers: argparse._SubParsersAction) -> 
         ),
     )
     configure_goal_parser.add_argument("--goal-id", required=True, help="Goal id to configure.")
-    configure_goal_parser.add_argument("--quota-compute", type=float, help="Per-goal quota compute multiplier.")
+    configure_goal_parser.add_argument("--quota-compute", type=float, help="Per-goal quota compute multiplier. 0 hard-pauses the whole Goal (all automatic agent turns stop); negative values are rejected.")
     configure_goal_parser.add_argument("--quota-window-hours", type=float, help="Quota rolling window in hours.")
     configure_goal_parser.add_argument(
         "--self-repair-enabled",
--- a/loopx/configure_goal.py
+++ b/loopx/configure_goal.py
@@ -154,6 +154,14 @@ def _positive_number(value: float | None, *, field: str) -> float | None:
     return float(value)
 
 
+def _non_negative_number(value: float | None, *, field: str) -> float | None:
+    if value is None:
+        return None
+    if value < 0:
+        raise ValueError(f"{field} must be greater than or equal to 0")
+    return float(value)
+
+
 def _non_negative_int(value: int | None, *, field: str) -> int | None:
     if value is None:
         return None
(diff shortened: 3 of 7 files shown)


huangruiteng (changes requested):
## Findings

**[P1] `quota.compute=0` still leaves Codex App heartbeat turns permanently active.**

The paused builder correctly disables delivery/repair authority, but it then calls the generic automation-liveness and scheduler projections. `build_automation_liveness()` falls through to `keep_active=true` / `pause_allowed=false`, while the quiet-wait scheduler path asks Codex App to keep polling at a 30/60-minute cadence. On Codex App, each scheduled heartbeat is itself an automatic model turn, so the documented promise that a hard-paused Goal "receives no automatic Codex turns" is not met; the Goal continues waking indefinitely even though no quota slot is recorded.

Please add a typed `quota_paused` liveness/scheduler branch that suspends the recurring host automation until an explicit quota resume, or narrow the public contract to say that delivery is paused while low-frequency control-plane polling remains active. Add a Codex App regression asserting the intended `automation_liveness`, host action, RRULE/apply behavior, and resume trigger. Relevant paths: `loopx/quota.py` around the paused finalization, `loopx/control_plane/scheduler/automation_liveness.py`, and the quiet-wait branch in `loopx/control_plane/scheduler/scheduler_hint.py`.

**[P2] The paused terminal path overwrites real health failures with `status_health_ok=true`.**

`_build_quota_plan_for_goal()` still computes scoped Goal health, but `_build_quota_paused_should_run_payload()` does not receive it and unconditionally emits both `ok=true` and `status_health_ok=true`. With a paused Goal plus a failed global registry/contract health projection, the computed health is false while the returned decision reports true. Pausing execution may preempt repair work, but it must not rewrite the status fact.

Please pass the computed `goal_health_ok` into the paused builder, preserve the existing `ok`/`status_health_ok` semantics, and add a paused-plus-unhealthy regression.

The selector-front terminal route is the right design and fixes the original capability/workspace authority contradiction. The lane matrix and recursive execution-signal assertions are valuable. The remaining issue is that the terminal execution decision is not yet terminal across the host-liveness and health-reporting contracts.

The `+423/-4` scope is mostly justified by the 280-line semantic test matrix. The roughly 120-line paused payload in the already large `loopx/quota.py` is partly avoidable: a typed terminal decision/finalizer under `loopx/control_plane/quota/` would reduce duplicated top-level protocol fields and make health/liveness consistency harder to miss.

## Validation

- exact reviewed head: `a5ee23f725835ab1f56b5fb64274c9d232f76034`
- new paused-precedence suite: 9 passed
- Python compile, Ruff, and `git diff --check`: passed
- merge-tree with current `origin/main`: clean
- remote `pytest` check: passed

## Decision

Request changes. The core precedence fix should remain, but the hard-pause liveness contract and paused health projection need the focused repairs above before merge.

huangruiteng (approved):
## Findings

No remaining blockers on exact head `ccc4282dc1f53aa4330e805f5cbf591e62255e4f`.

The two prior contract findings are resolved:

- **P1 resolved:** `quota_paused` now projects `automation_liveness.keep_active=false` and a typed scheduler stop. On Codex App it requests one pause/delete host action, emits no replacement RRULE, requires no ACK or quota spend, and resumes only after `quota.compute > 0`.
- **P2 resolved:** the paused builder receives `goal_health_ok`, so execution remains terminally paused while `ok` and `status_health_ok` preserve an unhealthy false status.

The motivation is sound: `compute=0` must be one Goal-level authority decision rather than a weak lane hint. The selector-front short circuit is the correct boundary because it prevents capability, workspace, monitor, and fallback lanes from recreating execution authority.

The repair stays cohesive. Host-liveness behavior is owned by the existing scheduler modules, while the quota builder only carries the computed health fact. The total `+532/-4` PR is test-heavy: 331 lines are the semantic matrix, and the new scheduler production delta is 56 lines across the existing liveness/hint owners. No additional abstraction is justified for this one terminal mode.

## Validation

- paused precedence suite: 11 passed
- complete quota/scheduler/heartbeat test cluster: passed
- Ruff on all changed Python surfaces: passed
- GitHub `pytest`: passed on the exact reviewed head
- `loopx canary premerge --from-git-diff`: exit 0; all selected checks and canaries passed
- public/private boundary scan and `git diff --check`: passed
- merge-tree against current `origin/main`: clean

Failures: none. Skips: none. Manual holds: none.

## Merge decision

Approved. The hard-pause contract is now consistent across selector authority, automation liveness, scheduler host behavior, health reporting, and regression coverage. Owner authorization permits self-merge after this exact-head review.

PR #2765, opened 2026-08-04
fix(todos): keep todo list JSON identities unique

### LINKED ISSUE #2762: todo list --format json emits duplicate entries for the same todo_id
## Summary

`loopx todo list --format json` returns repeated entries for the same `todo_id` — consumers must dedupe by `todo_id` to get an accurate list.

## Context

Observed on v0.4.1 (release snapshot from 3afc4cfc) during a live pilot: a goal with 4 todos, driven headless via direct CLI calls with an explicit absolute `--registry` on every call (project-local registry, one registered agent, goal synced to the global registry). The JSON output repeated entries; deduplicating on `todo_id` restored the correct set of 4.

## Caveat

We have not traced the emitting code path, so this report is observation-only — it may be intentional (e.g. one entry per lane/view/route projection) rather than a defect. If it is intentional, a `distinct`-style flag or a documented uniqueness key for JSON consumers would resolve the ambiguity; if not, deduping on `todo_id` at emission would.

Happy to provide the exact registry/goal shape that reproduces it.

superche:
## Summary

- merge Markdown active-state and event-projection todos by goal-wide `todo_id` before splitting them into user and agent lanes
- let the event projection replace stale Markdown role data while preserving existing overlay diagnostics and ordering
- add a regression test that proves top-level JSON todo identities remain unique across a cross-role projection overlay

## Root cause

`_merge_todo_projection_fields` deduplicated items independently inside each role. When an event-projected todo had the same `todo_id` as a Markdown todo but a different role, both role buckets retained a copy and the flattened `todos` array emitted the identity twice.

## User impact

`loopx todo list --format json` now emits one top-level todo per `todo_id` for projection-overlay reads. Consumers no longer need to deduplicate cross-role copies, and the event-projected role is authoritative.

## Validation

- `uv run --with 'pytest>=8,<9' python -m pytest tests/control_plane -q` — 708 passed
- `uv run --with 'pytest>=8,<9' python -m pytest tests/control_plane/test_todo_list_agent_lane_projection.py -q` — 3 passed
- `python3 examples/control_plane/todo-list-event-projection-smoke.py` — passed
- `uv run --with 'ruff>=0.12,<0.16' ruff check loopx/todos.py tests/control_plane/test_todo_list_agent_lane_projection.py` — passed
- `uv run loopx canary premerge --from-git-diff` — 17 selected checks passed, 0 failures, 0 warnings, 0 manual holds

No validation failures or skips. The system Python did not include pytest, so pytest was run in an isolated uv environment. The candidate paths passed the public/private boundary scan; generated `uv.lock` and local environment state were excluded.

Fixes #2762

### DIFF
--- a/examples/cli-control-plane-command-modularization-smoke.py
+++ b/examples/cli-control-plane-command-modularization-smoke.py
@@ -45,7 +45,11 @@ def assert_source_shape() -> None:
     assert_contains(init_text, "handle_quota_command", "__init__")
     assert "todo_parser = sub.add_parser" not in cli_text
     assert "quota_parser = sub.add_parser" not in cli_text
-    assert_contains(cli_text, "register_todo_command(sub)", "cli.py")
+    assert_contains(
+        cli_text,
+        "register_todo_command(sub, add_subcommand_format)",
+        "cli.py",
+    )
     assert_contains(cli_text, "register_quota_command(sub)", "cli.py")
     assert_contains(cli_text, "handle_todo_command(", "cli.py")
     assert_contains(cli_text, "handle_quota_command(", "cli.py")
--- a/loopx/cli.py
+++ b/loopx/cli.py
@@ -276,7 +276,7 @@ def build_parser() -> LoopXArgumentParser:
     register_dreaming_commands(sub, add_subcommand_format)
     register_evidence_log_command(sub, add_subcommand_format)
     register_explore_commands(sub, add_subcommand_format)
-    register_todo_command(sub)
+    register_todo_command(sub, add_subcommand_format)
     register_task_lease_command(sub, add_subcommand_format)
     register_quota_command(sub)
 
@@ -735,6 +735,7 @@ def main(argv: list[str] | None = None) -> int:
             args,
             registry_path=registry_path,
             runtime_root_arg=args.runtime_root,
+            format_name=output_format(args),
             print_payload=print_payload,
             append_cli_rollout_event=append_cli_rollout_event,
         )
--- a/loopx/cli_commands/todo.py
+++ b/loopx/cli_commands/todo.py
@@ -46,7 +46,10 @@ PrintPayload = Callable[
 ]
 
 
-def register_todo_command(subparsers: argparse._SubParsersAction) -> None:
+def register_todo_command(
+    subparsers: argparse._SubParsersAction,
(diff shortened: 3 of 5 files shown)


superche:
## Differential end-to-end validation

Ran the same generated registry, Markdown active state, and event log through the real JSON CLI on base `eadaccae` and head `bcebe5cc`:

```text
python -m loopx.cli --format json --registry <fixture-registry> todo list --goal-id <fixture-goal>
```

| Assertion | Base `eadaccae` | Head `bcebe5cc` |
| --- | ---: | ---: |
| `todo_count` | 280 | 279 |
| returned `todos` | 280 | 279 |
| unique `todo_id` values | 279 | 279 |
| duplicate IDs | `todo_agent_unclaimed` | none |
| copies of target ID | 2 (`user`, `agent`) | 1 (`user`) |
| overlay classification | both `markdown_only` and `event_only` | `overlaid` only |

The differential E2E assertions passed. This confirms the base behavior reproduces issue #2762 and the fix removes the duplicate through the public CLI path while keeping the event-projected role authoritative.

superche:
Follow-up validation after checking the exact command reported by the user:

`loopx todo list --format json`

The first revision only exercised the canonical global-option form. The exact command above was not accepted on `main`, so this PR now also adds todo-local output-format parsing and a real subprocess CLI regression.

End-to-end fixture: four Markdown todos, with one matching event-projection row whose authoritative role is `user`.

- `main`, exact command: exits 2 (`unrecognized arguments: --format json`)
- `main`, canonical global-format command: 5 returned rows / 4 unique IDs; `todo_first` appears in both `user` and `agent`
- PR head, exact command: 4 returned rows / 4 unique IDs; no duplicate IDs; `todo_first` appears once with role `user`

Validation on the final diff:

- exact four-todo CLI differential E2E: passed
- focused CLI/todo tests: 86 passed
- full control-plane tests: 708 passed
- Ruff on changed Python files: passed
- event-projection and CLI modularization smokes: passed
- `loopx canary premerge --from-git-diff`: 18 selected checks passed, 0 failures, 0 warnings, 0 manual holds; public-boundary scan passed

The first canary attempt was blocked by sandbox permissions for Git worktree creation and the uv cache. It was rerun in the authorized environment and passed; these were environment failures, not skipped product checks.

PR #2800, opened 2026-08-05
fix(scheduler): honor monitor heartbeat deadlines

huangruiteng:
## Summary

- lower Codex App heartbeat intervals when a runnable continuous monitor is due sooner than the default floor
- reuse the canonical Todo cadence parser, including minute words and second-based forms
- preserve the earliest monitor deadline after quota hot-payload compaction
- ignore expired monitors when deriving the compacted frontier deadline

This is a maintainer successor to the retracted #2795. Thanks @0xrep for identifying the monitor cadence direction and contributing the initial contract shape.

## Why

The scheduler previously treated the 15-minute monitor wait floor as authoritative even when a monitor declared a tighter cadence or had a nearer `next_due_at`. It also read only the retained monitor rows, so quota compaction could hide the actual earliest deadline.

The change keeps one source of truth: canonical Todo cadence parsing plus the existing `frontier_deadline` projection. It does not widen the quota payload or add a new scheduler protocol.

## Validation

- focused scheduler pytest: 174 passed
- Ruff: passed
- `examples/control_plane/monitor-scheduler-contract-smoke.py`: passed
- `git diff --check`: passed
- original review head standard premerge canary: 18/18
- latest-main premerge canary: all executed checks passed except the existing `hot-path-interface-budget-smoke.py` baseline (`heartbeat_prompt_json` has 31 top-level keys vs budget 30), reproduced unchanged on clean `origin/main`

No credentials, private paths, raw logs, or local state are included.


huangruiteng:
Closing this maintainer successor. Per maintainer direction, the fix will continue in the original contributor PR #2795.

PR #2802, opened 2026-08-05
fix(heartbeat): keep standard prompt within interface budget

huangruiteng:
## Summary

- omit the inactive `visible_goal_host` marker from ordinary heartbeat JSON
- preserve the explicit `traex-cli` host projection
- add a focused regression assertion for the standard heartbeat contract

## Motivation and design

The standard heartbeat payload reached 31 top-level keys against its durable 30-key hot-path budget after TraeX support added a host marker that was `null` for every ordinary heartbeat. Conditional projection removes information-free hot-path state while keeping the explicit visible-goal contract unchanged. It follows the adjacent optional projection pattern and does not add a compatibility wrapper or raise the budget.

## Validation

- `hot-path-interface-budget-smoke.py`: passed; heartbeat JSON is 30/30 top-level keys
- focused host-loop and CLI budget tests: 86 passed
- heartbeat prompt, quota flow, status/quota/review parity, active-state budget, and performance smokes: passed
- peer-agent runtime and onboarding host-loop smokes: passed
- repository-configured strict mypy: passed for 15 source files
- public/private boundary scan: 9 checks, 0 warnings
- diff-based premerge canary: 15/15 selected checks passed; no failures, skips, or manual holds
- exact-scope change-quality receipt: valid

The first premerge attempt used a 180-second per-check limit and timed out the composite peer-agent runtime smoke. That smoke passed independently with the supported Python runtime, and the authoritative 300-second rerun passed all 15 checks.

## Risk

Low and localized to heartbeat JSON shape. Generic consumers already use optional lookups, while TraeX tests continue to require `visible_goal_host=traex-cli` on the explicit host path.

### DIFF
--- a/loopx/heartbeat_prompt.py
+++ b/loopx/heartbeat_prompt.py
@@ -685,7 +685,11 @@ def build_heartbeat_prompt(
         "registered_agents": normalized_registered_agents,
         "runtime_profile": runtime_profile,
         "scheduler_execution_context": scheduler_execution_context,
-        "visible_goal_host": visible_goal_host,
+        **(
+            {"visible_goal_host": visible_goal_host}
+            if visible_goal_host
+            else {}
+        ),
         **(
             {
                 "initial_runtime_capability_projection": (
--- a/tests/test_host_loop_activation.py
+++ b/tests/test_host_loop_activation.py
@@ -463,6 +463,12 @@ def test_opencode_activation_uses_bridge_tool_and_generic_cli_quota() -> None:
     assert "--runtime-profile generic_cli" in packet["commands"]["heartbeat_prompt"]
 
 
+def test_standard_heartbeat_omits_inactive_visible_goal_host() -> None:
+    payload = build_heartbeat_prompt(goal_id="standard-heartbeat-fixture", thin=True)
+
+    assert "visible_goal_host" not in payload
+
+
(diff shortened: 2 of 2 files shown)


huangruiteng:
No blocking findings.

**Product and architecture judgment**

The regression came from projecting a host-specific field as `null` on the standard heartbeat hot path. Omitting inactive optional state solves the operator-facing budget failure without weakening the 30-key contract or changing TraeX activation semantics. The implementation reuses the payload's existing conditional projection style and is preferable to raising the budget.

**Validation**

- changed surfaces: `loopx/heartbeat_prompt.py` and one focused host-loop test
- hot-path budget: passed at 30/30 top-level keys
- focused pytest: 86 passed
- related heartbeat/quota/status and peer-runtime smokes: passed
- mypy and public/private boundary scan: passed
- diff-based premerge canary: 15/15 passed with zero failures, skips, warnings, or manual holds
- exact-scope change-quality receipt: valid

The initial 180-second canary attempt timed out the composite peer-runtime smoke; the exact smoke passed independently, then the authoritative canary rerun with a 300-second per-check limit passed in full.

**Merge decision**

Self-merge is appropriate: the change is single-purpose, validated against the exact final diff, carries no private artifacts, and the authoritative premerge gate reports `self_merge_allowed=true`.

PR #2857, opened 2026-08-07
fix(quota): preserve typed monitor successor routing

huangruiteng:
## Summary

- add typed successor routing options to `quota monitor-poll` and persist action, repository, capabilities, continuation, and target identity
- require explicit action semantics; require an explicit successor repository only when the source monitor is repository-bound
- default continuation to `independent_handoff` and derive a stable transition target when omitted
- expose successor IDs/receipts in artifacts, run indexes, replay payloads, and rollout events
- fail before monitor state mutation when successor routing is incomplete or invalid

## Validation

- focused monitor followthrough and CLI projection pytest: passed
- monitor-poll writeback smoke: passed
- mypy, ruff, CLI output budget, maintainability ratchet, public-boundary scan: passed
- `loopx canary premerge --from-git-diff --goal-id loopx-meta`: 18/18 passed, no manual holds
- change-quality receipt: `cqr_406e0dc1a6850de23e71` (valid)

This is a runtime/control-plane change, so it is opened for normal review rather than self-merged.


huangruiteng:
Qualification note

Changed surfaces: `quota monitor-poll` CLI/API routing, monitor Todo writeback, run-index/artifact/rollout receipt projection, replay readback, focused contract tests, and the self-repair pattern catalog.

Checks: focused pytest passed; monitor-poll writeback smoke passed; mypy and ruff passed; CLI output-budget regression passed; maintainability ratchet passed; public-boundary scan passed; exact committed-diff premerge passed 18/18 with receipt `cqr_406e0dc1a6850de23e71`.

Failure history: the first premerge run exposed an oversized decision-function ratchet; receipt parsing was extracted and the final committed diff was requalified. Skips: none. Manual holds: none.

Coverage is sufficient because it exercises both negative routes (missing action and repository-bound successor without an explicit repository), verifies failure before monitor mutation, validates safe defaults for non-repository successors, and reads the same successor ID back from Todo state, artifacts, indexes, replay, and rollout events.

PR #2880, opened 2026-08-08
fix(quota): honor allowed_slots=0 as a real cap instead of unlimited

### LINKED ISSUE #2876: fix(quota): allowed_slots=0 is treated as unlimited, so the strictest cap never throttles
## Problem

`quota_status` treats `allowed_slots == 0` as *no cap* rather than *a cap of
zero*, so the strictest possible quota setting produces an unbounded loop.

`loopx/quota.py:511`:

```python
elif waiting_on == "codex":
    if allowed_slots > 0 and spent_slots >= allowed_slots:
        state = "throttled"
    else:
        state = "eligible"
```

The `allowed_slots > 0` guard is the only place in the module that special-cases
zero. With it, `spent_slots=99, allowed_slots=0` short-circuits to `eligible`.

This contradicts the documented contract in `docs/quota-allocation.md:1115`:

> if `after.spent_slots >= after.allowed_slots`, `after.state` should be
> `throttled` and `after.should_run` should be `false`

and inverts the documented purpose of the override at
`docs/quota-allocation.md:98`, which lists **"a deliberately stricter
experimental cap"** as a reason an operator would set `allowed_slots` explicitly.

`compute: 0` is a separate control and is handled earlier (`quota.py:486`,
`state = "paused"`), so it does not mask this: a goal can hold `compute: 1.0`
and `allowed_slots: 0` at the same time and stay `eligible` forever.

## Reproduction

```python
from loopx.quota import quota_status

def goal(**quota):
    return {"id": "demo", "quota": {"compute": 1.0, "window_hours": 24,
                                    "slot_minutes": 1, **quota}}

for allowed, spent in [(1, 0), (1, 1), (0, 0), (0, 1), (0, 99)]:
    r = quota_status(goal(allowed_slots=allowed, spent_slots=spent), waiting_on="codex")
    print(f"allowed={allowed} spent={spent:3d} -> {r['state']}")
```

Observed at `82d006ff`:

```
allowed=1 spent=  0 -> eligible
allowed=1 spent=  1 -> throttled     <- correct
allowed=0 spent=  0 -> eligible
allowed=0 spent=  1 -> eligible      <- over the cap, still eligible
allowed=0 spent= 99 -> eligible      <- 99 slots over a 0-slot cap
```

`allowed_slots=1` throttles correctly at `spent=1`, so the boundary logic itself
is fine — zero is uniquely mishandled.

## Impact

`state` is the input to the downstream gates: `normal_delivery_allowed` is
`goal_health_ok and state == "eligible"` (`loopx/quota.py:1551`), and
`eligible_spend` in `control_plane/quota/slot_accounting.py` keys off the same
value. So the setting an operator would reach for to mean *"spend nothing"*
leaves every gate green and `spent_slots` growing without bound. This is a
fail-open on the safety mechanism, which is the direction that matters most for
a component whose whole job is to stop a scheduler from spending.

## Secondary note (not a separate bug, but relevant to the fix)

`allowed_slots` can also reach 0 without an explicit override, because the
default is rounded (`quota.py:398`):

```python
default_allowed_slots = round((window_hours * 60 / slot_minutes) * compute)
```

With the shipped defaults (`window_hours=24`, `slot_minutes=1`) this is
unreachable — `compute=0.01` still yields 14, and `compute<=0` is caught by the
`paused` branch first. It requires the documented "non-minute scheduler" case,
e.g. `window_hours=1, slot_minutes=60, compute=0.4` → `round(0.4)` → `0`.

Worth stating plainly: fixing the guard changes those configurations from
*silently unlimited* to *throttled*. I think throttled is the correct and safe
reading of "zero slots", and it is at least visible rather than silent, but if
you would rather the window-derived default floor at 1 when `compute > 0`, that
is a product call and I am happy to follow it — I kept it out of the proposed
change deliberately.

## Proposed scope

In scope:

- drop the `allowed_slots > 0` guard so the cap is honoured at zero;
- regression coverage for the zero cap, the non-zero boundary, and `paused`
  precedence.

Out of scope:

- any change to how the default `allowed_slots` is derived;
- the `compute` pause path, which already behaves correctly.

## Validation plan

```bash
python -m pytest -q tests/control_plane/test_quota_zero_allowed_slots.py
python -m pytest -q tests/control_plane/test_quota_slot_accounting.py \
  tests/control_plane/test_quota_plan_policy.py \
  tests/control_plane/test_quota_run_decision.py \
  tests/control_plane/test_quota_paused_precedence.py \
  tests/presentation/test_quota_markdown_renderers.py
python -m pytest -q
python -m ruff check tests loopx/canary loopx/control_plane loopx/domain_packs loopx/presentation
python -m mypy
```

I have the fix and regression tests ready and will open a PR against this issue.

shani-singh1:
## Summary

- Drop the `allowed_slots > 0` guard in `quota_status`, so a zero cap throttles
  instead of short-circuiting to `eligible`.
- Add focused regression coverage for the zero cap, the non-zero boundary, and
  `paused` precedence.

Production diff is one line. The rest is tests.

## Issue Or Task

- Closes #2876
- Contributor task ID: n/a (fail-open found by review)

## What Changed

`loopx/quota.py:511`:

```diff
-        if allowed_slots > 0 and spent_slots >= allowed_slots:
+        if spent_slots >= allowed_slots:
```

`allowed_slots > 0` was the only place in the module that special-cased zero, so
`spent_slots=99, allowed_slots=0` projected `eligible`. Because `state` feeds
`normal_delivery_allowed` (`quota.py:1551`) and `eligible_spend`
(`control_plane/quota/slot_accounting.py:443`), the setting that most obviously
means "spend nothing" left every downstream gate green while `spent_slots` grew
without bound.

`docs/quota-allocation.md:1115` already states the intended contract — 
`spent_slots >= allowed_slots` must project `throttled` — and
`docs/quota-allocation.md:98` lists "a deliberately stricter experimental cap" as
a reason to set `allowed_slots` explicitly. So this restores documented behavior
rather than introducing new behavior; no doc change is needed.

`compute <= 0` is handled earlier (`quota.py:486`) and is untouched, so `paused`
keeps precedence and a goal can no longer sit at `compute: 1.0` with
`allowed_slots: 0` and stay eligible forever.

## Behavior Change Worth Your Call

`allowed_slots` can also reach 0 without an explicit override, since the default
is rounded (`quota.py:398`). With the shipped defaults (`window_hours=24`,
`slot_minutes=1`) that is unreachable — `compute=0.01` still yields 14, and
`compute<=0` is caught by `paused` first — so it requires the documented
"non-minute scheduler" case, e.g. `window_hours=1, slot_minutes=60, compute=0.4`
→ `round(0.4)` → `0`.

Those configurations move from *silently unlimited* to *throttled*. I believe
throttled is the correct reading of "zero slots" and it is the fail-closed
direction for a spend limiter, but if you would rather the window-derived default
floor at 1 when `compute > 0`, that is a product decision — say the word and I
will follow up. I deliberately left the derivation untouched here to keep this
diff to the fail-open itself.

## Validation

Run on Linux / CPython 3.11 (CI is ubuntu-only), container run as a non-root user.

- [x] `python -m ruff check loopx/quota.py tests/control_plane/test_quota_zero_allowed_slots.py` — clean
- [x] `python -m mypy` — `Success: no issues found in 15 source files`
- [x] `python -m pytest -q tests/control_plane/test_quota_zero_allowed_slots.py` — 10 passed
- [x] quota-adjacent suites — 50 passed (`test_quota_slot_accounting`, `test_quota_plan_policy`, `test_quota_run_decision`, `test_quota_paused_precedence`, `test_quota_markdown_renderers`)
- [x] `git diff --check` — clean

The production change is a single line inside one `elif` branch of `quota_status`;
the tests above exercise that branch and its neighbours (non-zero boundary,
`paused` precedence, the window-derived default). I scoped validation to that
blast radius rather than re-running the whole suite for a one-line guard removal.

Behavior before/after, same fixture:

```
                     before        after
allowed=1 spent=  1  throttled     throttled     <- unchanged
allowed=0 spent=  0  eligible      throttled
allowed=0 spent=  1  eligible      throttled
allowed=0 spent= 99  eligible      throttled
```

## Tests

`tests/control_plane/test_quota_zero_allowed_slots.py`, 10 cases. The 3 zero-cap
cases fail on `main` and pass here; the other 7 pass on **both**, which is the
point — they pin that the non-zero boundary, `paused` precedence, and the
window-derived default are all unaffected by the change.

- `test_zero_allowed_slots_throttles_instead_of_running_unbounded[0|1|99]` — the fail-open itself.
- `test_nonzero_allowed_slots_boundary_is_unchanged` — `(1,0)`, `(1,1)`, `(10,9)`, `(10,10)`, `(10,11)`; guards against turning this into an off-by-one.
- `test_zero_compute_still_reports_paused_not_throttled` — `compute: 0` keeps precedence over the cap.
- `test_window_derived_default_stays_eligible_for_normal_config` — a normal `compute=0.01` goal still derives 14 slots and stays `eligible`.

## Boundary Checklist

- [x] I did not commit `.loopx/`, `.codex/goals/`, live `ACTIVE_GOAL_STATE.md`, credentials, private benchmark traces, verifier output, raw agent sessions, internal document links, or local machine paths.
- [x] I did not duplicate maintainer-owned benchmark work unless a maintainer split out a public issue for it.
- [x] I kept the change scoped to the linked issue/task.

### DIFF
--- a/loopx/quota.py
+++ b/loopx/quota.py
@@ -508,7 +508,7 @@ def quota_status(
         payload["blocked_action_scope"] = "delivery_focus"
         payload["focus_wait"] = True
     elif waiting_on == "codex":
-        if allowed_slots > 0 and spent_slots >= allowed_slots:
+        if spent_slots >= allowed_slots:
             state = "throttled"
             reason = f"{compute:g} compute quota spent {spent_slots}/{allowed_slots} slots in this window"
         else:
--- a/tests/control_plane/test_quota_zero_allowed_slots.py
+++ b/tests/control_plane/test_quota_zero_allowed_slots.py
@@ -1,0 +1,84 @@
+from __future__ import annotations
+
+import pytest
+
+from loopx.quota import goal_quota_config, quota_status
+
+
+GOAL_ID = "zero-allowed-slots-fixture"
+
+
+def _goal(**quota: object) -> dict[str, object]:
+    return {
+        "id": GOAL_ID,
+        "quota": {
+            "compute": 1.0,
+            "window_hours": 24,
+            "slot_minutes": 1,
+            **quota,
+        },
+    }
+
+
+@pytest.mark.parametrize("spent_slots", [0, 1, 99])
+def test_zero_allowed_slots_throttles_instead_of_running_unbounded(
+    spent_slots: int,
+) -> None:
+    """`allowed_slots: 0` is the strictest cap, not an absence of one.
+
+    `docs/quota-allocation.md` documents `allowed_slots` as an explicit override
+    for "a deliberately stricter experimental cap", and states that
+    `spent_slots >= allowed_slots` must project `throttled`. Treating 0 as
+    unlimited inverts the operator's intent on the safety mechanism itself.
+    """
+
+    payload = quota_status(
+        _goal(allowed_slots=0, spent_slots=spent_slots), waiting_on="codex"
+    )
+
+    assert payload["allowed_slots"] == 0, payload
+    assert payload["state"] == "throttled", payload
+    assert "0" in payload["reason"], payload
+
+
+@pytest.mark.parametrize(
+    ("allowed_slots", "spent_slots", "expected"),
+    [
+        (1, 0, "eligible"),
+        (1, 1, "throttled"),
+        (10, 9, "eligible"),
+        (10, 10, "throttled"),
+        (10, 11, "throttled"),
+    ],
+)
+def test_nonzero_allowed_slots_boundary_is_unchanged(
+    allowed_slots: int, spent_slots: int, expected: str
+) -> None:
+    payload = quota_status(
+        _goal(allowed_slots=allowed_slots, spent_slots=spent_slots),
+        waiting_on="codex",
+    )
+
+    assert payload["state"] == expected, payload
+
+
+def test_zero_compute_still_reports_paused_not_throttled() -> None:
+    """`compute: 0` is the pause control and keeps precedence over the cap."""
+
+    payload = quota_status(
+        _goal(compute=0, allowed_slots=0, spent_slots=5), waiting_on="codex"
+    )
+
+    assert payload["state"] == "paused", payload
+
+
+def test_window_derived_default_stays_eligible_for_normal_config() -> None:
+    """The documented default must not collapse to a throttling cap."""
+
+    config = goal_quota_config(_goal(compute=0.01))
+
+    assert config["allowed_slots"] == 14, config
+    assert (
+        quota_status(_goal(compute=0.01, spent_slots=0), waiting_on="codex")["state"]
+        == "eligible"
+    )


huangruiteng (commented):
### Review: no blockers

head `9def43a92e71262cf1f0092dfd676684b1ef259c`。生产改动只有一行，修复 `allowed_slots=0` 被当作无限配额而始终 eligible 的 fail-open，符合文档契约；测试覆盖充分，未发现阻断项。

## 动机

`quota_status` 之前的判断是 `allowed_slots > 0 and spent_slots >= allowed_slots`，因此 operator 设置最严格的 `allowed_slots: 0` 时，`spent_slots` 无论多大都会投影为 eligible，`normal_delivery_allowed` 与 `eligible_spend` 持续绿灯，配额变成无界。文档 `docs/quota-allocation.md` 已写明 `spent_slots >= allowed_slots` 应投影 throttled，并允许用显式 `allowed_slots` 做更严格的实验性 cap，所以这是修复实现与文档的偏差，而不是新行为。

## 改动思路

删除 `allowed_slots > 0` 这个零值特判，让比较本身决定状态：`spent_slots >= allowed_slots` 即 throttled。`compute <= 0` 的 paused 分支在更早位置，保持不变，所以 pause 优先级不受影响；窗口推导出的默认值（如 `window_hours=1, slot_minutes=60, compute=0.4` 会 round 到 0）从“静默无限”变成“throttled”，这是限流器 fail-closed 的正确方向。

## 具体改动

2 文件、+85/-1。`loopx/quota.py:511` 删除一行 guard；新增 `tests/control_plane/test_quota_zero_allowed_slots.py` 覆盖零 cap（0/1/99 spent）、非零边界（1/10 的临界点）、`compute=0` 的 paused 优先级、以及正常窗口默认值仍 eligible。

### 关键代码讲解

`quota_status` 的 throttled 分支：
```python
elif waiting_on == "codex":
    if spent_slots >= allowed_slots:
        state = "throttled"
```
这是 quota 状态机里唯一决定“spend cap 已耗尽”的守卫。输入是 goal quota 的 `allowed_slots` 与窗口内 `spent_slots`；输出 state 会继续流向 `normal_delivery_allowed` 与 `eligible_spend`。删除零值特判后，0 成为“一个配额都不允许”，而非“没有配额上限”，语义与文档一致。

## 对主干的风险

低。唯一行为面变化是窗口默认值可能 round 到 0 的非分钟调度配置从 eligible 变为 throttled；这是 fail-closed 且与“0 slots”字面含义一致。测试覆盖了该风险附近的三类边界（零 cap、非零边界、paused precedence），7 个 CI check 全绿。

## 我的整体评价

改动小而准确：一行生产修复 + 84 行针对性测试，PR 描述把 before/after 行为和边界讲得很清楚。无阻断项，merge 按 loopx-pr-merge 政策单独路由。

huangruiteng (approved):
### Approval

head `9def43a92e71262cf1f0092dfd676684b1ef259c`. No blockers found; detailed review already published at https://github.com/huangruiteng/loopx/pull/2880#pullrequestreview-4888668200. Merge follows loopx-pr-merge policy.

PR #2884, opened 2026-08-08
feat(goal-channel): reuse shared provider targets

huangruiteng:
## Summary
- add named, local-private Lark provider targets that multiple Goal Channel bindings can reference
- add `goal-channel setup --target` and bounded `goal-channel attach` while keeping control messages, Kanban state, receipts, and cooldowns Goal-local
- resolve target configuration only at setup/runtime entry points; the gate decision and notification send/readback contract remain unchanged

## Validation
- `pytest`: 2171 passed, 2 skipped
- focused Goal Channel tests: 105 passed
- Ruff on changed Python, mypy, and `git diff --check`: passed
- standard LoopX premerge: 13/13 passed, 0 failures, 0 warnings, no manual holds
- change-quality: `cqr_25435f10532e2421dc07` for fingerprint `25435f10532e2421dc070398db8378a542f1bbc40119be4c29a000b1fa348568`

## Safety
- no chat ids, bot profiles, credentials, local paths, or private runtime state are committed
- the same target alias can be configured independently on each machine; target values are not copied into Goal state
- inbound reply routing remains Goal-explicit

### DIFF
--- a/docs/architecture/rfcs/goal-channel-collaboration-v0.md
+++ b/docs/architecture/rfcs/goal-channel-collaboration-v0.md
@@ -82,6 +82,9 @@ Suggested command surface:
 
 ```bash
 loopx goal-channel setup --provider lark --goal-id <goal-id>
+loopx goal-channel target add --name <target> --provider lark ...
+loopx goal-channel setup --goal-id <goal-id> --target <target>
+loopx goal-channel attach --target <target> --goal-id <goal-a> --goal-id <goal-b>
 loopx goal-channel doctor --goal-id <goal-id>
 loopx goal-channel sync --goal-id <goal-id>
 loopx goal-channel notify-gate --goal-id <goal-id>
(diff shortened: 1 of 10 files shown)


huangruiteng:
Premerge review packet:

- **changed_surfaces:** Goal Channel CLI/config resolution, Lark extension runtime entry points, provider-target private store, bilingual RFC, focused regression tests
- **direct_checks:** committed/staged/unstaged diff checks and changed-Python compile passed
- **catalog_canaries:** 4/4 passed
- **risk_profile_smokes:** 8/8 passed
- **public_private_boundary:** clean across all 10 changed files
- **failures_or_skips:** no premerge failures or skips; full pytest had 2171 passed and 2 existing skips
- **manual_holds:** none
- **merge_decision:** ready for independent review; core gate selection and notification delivery contract are unchanged
- **change_quality_receipt:** `cqr_25435f10532e2421dc07`, exact-scope fingerprint `25435f10532e2421dc070398db8378a542f1bbc40119be4c29a000b1fa348568`

huangruiteng:
### Post-merge review (exact head `25503bbdf7071c5960bca13bb50a6890f350c9dc`)

Direction is right: reusable local-private provider targets remove repeated per-goal credential/config copying while keeping control messages, Kanban, receipts, and cooldowns Goal-local. The target store is private, atomic, and 0600; setup forbids mixing `--target` with direct channel/identity flags; attach is bounded to 8 goals and fails fast; and the tests cover shared target isolation, per-goal Kanban separation, and public-packet redaction. Overall I would have approved this shape.

Two follow-ups to consider on main:

**P2: target-linked bindings can hit an internal assertion when no provider target is passed.** In `notify_lark_goal_channel_gate`, `raw_binding` is resolved without a target, then `binding_for_goal(..., provider_target=provider_target)` is called and `assert binding is not None` (goal_channel_runtime.py:321-333). A binding with `target_ref` but no `provider_target` is returned as-is and then fails the assert, producing an internal error instead of the same clean `provider_target_missing` packet the CLI already emits. Direct library/extension callers that omit the target (for example a future automation path that calls `notify_lark_goal_channel_gate` without re-reading the target store) have no safe error route. Minimal fix: when the raw binding has a `target_ref` and `provider_target is None`, return a `provider_target_missing` operation packet; add one negative test for that path.

**P2: migrating a shared target to another group leaves the old pinned message id in the raw binding until reattach succeeds.** The raw binding stores only `pinned_message_id` plus `target_ref`; `binding_for_goal` merges in the target channel (goal_channel_contracts.py:97-123). After a target moves to a new group, re-running `attach` sends and pins a new control message and then replaces `pinned_message_id`, so the durable end state is correct. In the window before reattach, however, a resolved binding points at the new chat with the old group's pinned id; the RFC already says to rerun attach after moving a target, so this is consistent with documented behavior, but a small invariant test asserting `pinned_message_id` is reset/omitted when the target's chat changes would make the migration contract machine-visible.

PR #2970, opened 2026-08-08
fix: prevent stale claims from starving priority

iscarson:
## Summary
- order explicit todo priority before claim ownership in capability and agent-lane selection
- reject overlapping same-agent open advancement claims for one target key
- document the focus-saturation repair pattern and update routing smokes

## Validation
- 48 focused pytest cases passed
- ruff passed on changed Python surfaces
- capability ordering and work-lane smokes passed
- standard premerge canary passed: 18/18 selected checks, 0 failures

Claims remain ownership signals and only break ties within an explicit priority bucket. This prevents a mature claimed lane from indefinitely starving newer P0/P1 product gaps.


iscarson:
Closing per maintainer preference; this fix will remain in the contributor fork and will not be proposed upstream.

PR #3007, opened 2026-08-09
fix(install): isolate skill promotion from caller cwd

huangruiteng:
## Summary

- isolate installer-owned Python imports from the caller's current working directory with `PYTHONSAFEPATH=1`
- preflight workflow-skill imports and exact-host entry materialization before switching the default `loopx` executable
- keep the previous default executable active and remove the candidate release when skill preflight fails
- add durable regressions for an incomplete shadow `loopx` package and a promotion-blocking user-owned exact-host skill

The installer previously allowed caller CWD state to select an unrelated Python package and performed default promotion before workflow-skill validation. That could report a failed install while leaving a newly activated CLI and partially updated skills. The change keeps source ownership in the existing installer and entry-skill contracts: a read-only preflight closes the known failure boundary, while the real install still uses the established lock and materialization path. No new installer framework or public protocol is introduced.

## Validation

- focused Ruff checks passed
- `python examples/install-local-smoke.py` passed, including the shadow-package CWD fixture
- `python examples/release/local-install-promotion-boundary-smoke.py` passed, including fail-closed default preservation
- `pytest tests/test_slash_command_install.py -q`: 37 passed
- `loopx canary premerge --from-git-diff --goal-id loopx-meta`: 18 selected checks passed; 0 failures, warnings, or manual holds
- strict change-quality receipt `cqr_30041cb378b680be97a4` is valid for the exact 3-file diff

## Boundary

No credentials, private runtime state, generated logs, local paths, benchmark evidence, or unrelated artifacts are included. The generated local dependency lock was excluded.

### DIFF
--- a/examples/install-local-smoke.py
+++ b/examples/install-local-smoke.py
@@ -63,10 +63,15 @@ def source_git_commit(root: Path = REPO_ROOT) -> str:
     return commit
 
 
-def run_install(env: dict[str, str], release_id: str) -> subprocess.CompletedProcess[str]:
+def run_install(
+    env: dict[str, str],
+    release_id: str,
+    *,
+    cwd: Path = REPO_ROOT,
+) -> subprocess.CompletedProcess[str]:
     return subprocess.run(
         [str(INSTALL_SCRIPT)],
-        cwd=REPO_ROOT,
+        cwd=cwd,
         env={**env, "LOOPX_RELEASE_ID": release_id},
         check=True,
         capture_output=True,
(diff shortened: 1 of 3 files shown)


huangruiteng:
## Findings

No blocking findings at head `0065661ffc4611f24623f8b201ec318a331b84b4`.

- This PR closes the observed caller-CWD import shadowing and the known exact-host skill collision before default promotion. It does not claim a cross-filesystem transaction for every possible disk or permission failure after preflight.
- Python 3.11+ is already an installer prerequisite, so `PYTHONSAFEPATH=1` is within the supported runtime contract.

- Motivation: a failed update could still activate the candidate CLI because workflow-skill validation happened after promotion, and stdin Python imports could resolve an unrelated package from the caller CWD.
- Solved: the shadow-package fixture proves source imports are CWD-independent; the negative promotion fixture proves a known skill preflight failure preserves the old executable and removes the candidate release.
- Operator impact: `loopx update` no longer depends on which checkout or directory launched it for these imports, and a blocked exact-host skill cannot leave a misleading promoted default.
- Main risk: release/install compatibility. The diff keeps the existing lock, materializer, candidate doctor, and actual install path; no permission, benchmark, scoring, or public protocol surface changes.
- Design: the read-only preflight is the smallest cohesive boundary fix. A larger generalized installer transaction would be speculative for this incident and should require additional real failure cases.

## Validation

- `bash -n scripts/install-local.sh`: passed
- focused Ruff: passed
- `python examples/install-local-smoke.py`: passed
- `python examples/release/local-install-promotion-boundary-smoke.py`: passed
- `pytest tests/test_slash_command_install.py -q`: 37 passed
- `loopx canary premerge --from-git-diff --goal-id loopx-meta`: 18/18 passed, 0 failures, 0 warnings, 0 manual holds
- strict receipt `cqr_30041cb378b680be97a4`: valid for exact fingerprint `30041cb378b680be97a4969981868afd4676f6fab6fdc178e2b42fb36098cc51`

## Merge decision

Approved for the owner-authorized self-merge path after required GitHub checks pass at this exact head.

PR #3009, opened 2026-08-09
fix(control-plane): require causal outcome-continuity waits

huangruiteng:
## Summary

- require an open vision acceptance gap to identify the active Todo lineage before blocked/deferred work can suppress replanning
- traverse only explicit blocker target/successor edges, excluding unrelated deferred Todo items
- require an outcome checkpoint after five scoped advancement completions even when another runnable Todo exists
- keep the hot quota payload cap at three while evaluating the rule against the full completion history

## Validation

- 101 focused control-plane tests passed after rebasing onto latest `origin/main`
- Ruff passed on all changed Python files
- repository-configured strict mypy passed on 15 source files
- compile and git diff checks passed
- strict change-quality receipt: `cqr_2991b158ca95f06c3921`
- premerge selected 17 canaries: 16 passed; the sole `quota_should_run_json` budget failure reproduces exactly on detached `origin/main` and this branch (12,591 vs 12,550), so this diff adds zero bytes to that payload

## Scope

This PR changes outcome-continuity selection and its focused regression coverage only. It does not alter installer behavior, the recent-completion payload display cap, or begin the M7.1 typed Effect Program runtime work.

### DIFF
--- a/loopx/control_plane/goals/goal_frontier/__init__.py
+++ b/loopx/control_plane/goals/goal_frontier/__init__.py
@@ -69,6 +69,13 @@ TODO_TASK_CLASS_ADVANCEMENT = "advancement_task"
 TODO_TASK_CLASS_MONITOR = "continuous_monitor"
 LONG_TODO_CHAIN_ADVANCEMENT_THRESHOLD = 15
 LONG_TODO_CHAIN_OPEN_THRESHOLD = 20
+VISION_FRONTIER_TODO_DELTA_ACTIONS = {
+    "activate",
+    "create",
+    "reopen",
+    "resume",
+    "retain",
+}
 
 
 def safe_non_negative_int(value: Any) -> int:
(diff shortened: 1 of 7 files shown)


huangruiteng:
Self-review and premerge qualification

Changed surfaces:
- `goal_vision_wait`: restrict wait suppression to active vision/current causal Todo lineage and explicit target/successor edges.
- `goal_frontier/outcome_continuity`: force a checkpoint after five scoped advancement completions, including when runnable work remains.
- `quota_summary`: retain the three-item hot display cap; the rule consumes the full internal summary.
- focused tests: unrelated deferred waits, exact lineage, missing lineage, 4/5 completion boundary, checkpoint reset, runnable-work precedence, and full-vs-compact history.

Checks:
- 101 focused tests passed.
- Ruff passed for every changed Python file.
- repository-configured strict mypy passed (15 source files).
- compile and diff checks passed.
- strict receipt `cqr_2991b158ca95f06c3921` is valid for base `66b067976a858754ec6b87397fc3ed7829368a9a` and head `89898251f53f0a8efcafe959bfe1510e6f82698a`.

Failure/skip accounting:
- Premerge executed all 17 selected canaries: 16 passed and 1 failed. The failure is the pre-existing `quota_should_run_json` size ratchet (12,591 actual / 12,550 budget). I replayed the same smoke in a detached latest `origin/main` worktree and got the identical 12,591 result, so this PR contributes zero payload growth. No selected canary was skipped.
- No manual product hold remains. Remote required checks must still pass before merge.

Coverage is sufficient because the changed selection rules are exercised at their positive, negative, threshold, precedence, agent-scope, and payload-compaction boundaries; the sole unrelated baseline failure is independently reproduced on the exact base.

PR #3022, opened 2026-08-09
fix(quota): preserve completed-chain evidence window

huangruiteng:
## Summary

- expose five recent completed advancement Todos in the quota hot-path payload, matching the five-completion outcome checkpoint cadence
- single-source the completed-chain threshold in goal-vision policy so decision logic and agent-facing evidence cannot drift
- retain the existing larger raw summary/cold-path bound and avoid widening unrelated diagnostic lanes

## Why five

Three hides two of the causal records required to explain a five-completion replan. Five is the minimum complete evidence window. Values above five add recurring payload cost without changing the checkpoint decision, so this PR deliberately does not choose seven, eight, or the full raw lane.

## Validation

- 237 focused quota/goal-vision/projection/scheduler tests passed
- Ruff 0.15.7 passed on all four changed Python paths
- repository strict mypy passed for all 16 configured files
- absolute CLI output budget and 97-row base/head differential passed with zero review signals
- premerge: 4 direct checks plus 17 risk-selected checks passed; no failures, warnings, or manual holds
- exact change-quality receipt: `cqr_36b9eee9c9af70f5fb84`

### DIFF
--- a/loopx/control_plane/goals/goal_frontier/outcome_continuity.py
+++ b/loopx/control_plane/goals/goal_frontier/outcome_continuity.py
@@ -5,6 +5,7 @@ from typing import Any
 
 from ...agents.agent_scope import agent_scope_item_claimed_by
 from ...work_items.repair_delta import repair_delta_kinds_have_frontier_delta
+from ..goal_vision_policy import COMPLETED_TODO_CHAIN_REPLAN_THRESHOLD
 from ..goal_vision_state import goal_vision_state_is_closed
 
 VISION_OUTCOME_CHECKPOINT_REQUIRED_TRIGGER = "vision_outcome_checkpoint_required"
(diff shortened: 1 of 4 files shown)


huangruiteng:
Self-review complete on exact head `9912324b6`.

Changed surfaces: goal-vision policy constant, outcome-continuity consumer, quota payload compaction, and one causal-window regression test. The raw summary still retains its existing larger history; only the hot-path projection moves from three to five items. No unrelated diagnostic lane changes.

The value is intentionally five: it is the smallest window that can display every completion used by the five-completion replan checkpoint. Three is causally incomplete; values above five increase recurring packet size without adding decision evidence. The implementation single-sources this value, so future threshold changes update both decision and projection.

Validation: 237 focused tests; Ruff 0.15.7; strict mypy 16 configured files; absolute CLI budget; 97-row differential with no review signals; premerge 4 direct + 17 selected checks. Failures, warnings, skips, manual holds: none. Exact receipt: `cqr_36b9eee9c9af70f5fb84` for scope `36b9eee9c9af70f5fb8417a7b41c239b96f5b3ed292aa00a1927de469f8a68a3`.

PR #3233, opened 2026-08-15
fix(codex-app): isolate fallback receipt identity

### LINKED ISSUE #3231: Codex App heartbeat backoff fallback fails with empty error and repeats NOTIFY
## Symptom and impact

On LoopX 0.4.6 (installed release `20260813T192037Z`) with the Codex App heartbeat, an active-work scheduler hint asked for a 3-minute cadence. When the decision moved to monitor wait and `stateful_backoff.apply_needed=true` requested a 30-minute RRULE, the `loopx-apply-rrule` fallback exited with code 1 and printed only:

```text
loopx should-run failed (1):
```

No stderr detail was emitted. The host automation therefore stayed at `FREQ=MINUTELY;INTERVAL=3`, and subsequent `quota should-run` decisions alternated between `user_todo_blocker_push` and `user_gate_cooldown_wait`. Even when `heartbeat_recommendation.notify=DONT_NOTIFY`, `interaction_contract.user_channel.notify=NOTIFY` with `action_required=true` kept firing, so the user received repeated notifications instead of quiet backoff.

Impact: notification spam, wasted heartbeat turns, and scheduler backoff not being applied on the host.

## Expected behavior

- The fallback bridge should either use a fresh idempotent turn id for its internal `quota should-run` call, or surface the full `quota should-run` failure payload instead of an empty error.
- When `heartbeat_recommendation.notify=DONT_NOTIFY`, the user notification channel should stay quiet during the cooldown window even if `user_channel.notify` still carries the pending items.

## Actual behavior

1. `quota should-run --codex-app --turn-instance-id <T>` settles a heartbeat receipt.
2. `loopx-apply-rrule --turn-instance-id <T>` internally runs `quota should-run` with the same `<T>`.
3. The internal call exits 1 with empty stderr; `codex_app_apply_rrule.py` prints `loopx should-run failed (1): ` and returns failure.
4. Host RRULE remains at 3 minutes; `quota scheduler-fail-current` records the failed target/observed pair.
5. Later heartbeat decisions alternate between blocker push and cooldown wait, and repeated NOTIFY messages are delivered despite `heartbeat_recommendation.notify=DONT_NOTIFY`.

## Minimal public-safe reproduction

Use a public/synthetic goal with an open user gate and a Codex App scheduler hint:

1. Run `loopx quota should-run --goal-id <fixture> --agent-id <agent> --codex-app --turn-instance-id T`.
2. In the same turn, run `loopx-apply-rrule --goal-id <fixture> --agent-id <agent> --turn-instance-id T`.
3. Observe exit code 1 and empty stderr on the second command when the should-run receipt for `T` is already settled.
4. Repeat `quota should-run` and observe the notification mode flip between `user_todo_blocker_push` and `user_gate_cooldown_wait`.

No task text, credentials, private paths, internal URLs, or project identifiers are included.

## Environment

- LoopX: 0.4.6, installed release `20260813T192037Z`
- Host: Codex App heartbeat on macOS
- Fallback bridge: `scripts/codex_app_apply_rrule.py`

## Suspected surface

- `scheduler_hint.codex_app.stateful_backoff` / `loopx-apply-rrule` fallback lifecycle
- `user_gate_notification_cooldown` projection when `heartbeat_recommendation.notify=DONT_NOTIFY`
- Failure diagnostics from `codex_app_apply_rrule._load_scheduler_hint`

## Validation already run

- Repeated independent heartbeat turns reproduced the 3-minute host RRULE persistence and the repeated NOTIFY pattern.
- `quota scheduler-fail-current` recorded the failed target/observed pair without quota spend.
- The exact failure occurred with a settled turn-instance id; fresh turn ids applied RRULE successfully.

huangruiteng:
Fixes #3231.

## What changed

- Treat the fallback command's `--turn-instance-id` as the parent host turn and derive a stable, public-safe child receipt identity for its internal `quota should-run` query.
- Preserve bounded stdout and stderr diagnostics from failed LoopX subprocesses, so JSON quota errors are no longer rendered as an empty message.
- Ship `loopx-apply-rrule` as a release-owned executable and have `install-local.sh` atomically install/upgrade it with the main CLI.
- Document the parent/child receipt contract and add focused runtime and installer regressions.

## Root cause

The heartbeat guard had already settled the outer turn against its selected Todo. After delivery changed the selected Todo, the fallback bridge queried `quota should-run` with that same turn id, which correctly failed the receipt identity check. The quota CLI emitted the structured failure on stdout, while the bridge only displayed stderr. Because the bridge stopped before applying and ACKing the requested RRULE, the host remained on the tight cadence and repeatedly resurfaced the downstream gate.

The existing domain-local scheduler failure/cooldown reducer remains authoritative; this fix repairs the host bridge before that layer rather than introducing an Effect Program abstraction.

## Validation

Passed:

- `28 passed` — focused fallback, scheduler hint, and interaction arbitration pytest modules
- `examples/install-local-smoke.py`
- `examples/install-local-overwrite-smoke.py`
- `examples/codex-cli-packaged-install-smoke.py`
- `examples/release-wrapper-import-isolation-smoke.py`
- `examples/release-artifacts-smoke.py`
- `examples/release/local-install-promotion-boundary-smoke.py`
- Ruff, shell syntax, Python compile, and `git diff --check`
- Public/private boundary scan

Risk-based `loopx canary premerge --from-git-diff` executed 16 checks: 14 passed. Its two failures are bounded and independently classified:

- `install-local-smoke.py` hit the canary's 120-second per-check limit; the same full smoke passed standalone.
- `monitor-scheduler-contract-smoke.py` fails with the identical assertion on a detached clean `origin/main@221929602`, so it is an existing baseline drift outside this PR.

No private state, credentials, raw logs, local paths, or generated lockfiles are included.

### DIFF
--- a/docs/heartbeat-automation-prompt.md
+++ b/docs/heartbeat-automation-prompt.md
@@ -546,7 +546,10 @@ If `automation_update` is unavailable in the session and
 `fallback_hint.cli_args` (`loopx-apply-rrule`) once instead. It backs up
 `codex-dev.db`, syncs the automation TOML and SQLite row, and runs the bound
 ACK; direct SQLite edits bypass the app API, so this is a bounded fallback and
-never the routine path. When the automation id could not be resolved,
+never the routine path. The bridge treats the provided turn id as its parent
+and derives a stable child receipt id for the internal quota query, so it does
+not reuse the already-settled heartbeat receipt. When the automation id could
+not be resolved,
 `fallback_hint.available=false` and the pasteable heartbeat gate is the correct
 stop - never guess an automation id.
 
--- a/examples/install-local-smoke.py
+++ b/examples/install-local-smoke.py
@@ -179,6 +179,9 @@ def main() -> int:
         stale_loopx = bin_dir / "loopx"
         stale_loopx.write_text("#!/usr/bin/env bash\nexit 99\n", encoding="utf-8")
         stale_loopx.chmod(0o755)
+        stale_fallback = bin_dir / "loopx-apply-rrule"
+        stale_fallback.write_text("#!/usr/bin/env bash\nexit 98\n", encoding="utf-8")
+        stale_fallback.chmod(0o755)
         stale_canary_target = root / "stale-canary-target"
         stale_canary_target.mkdir()
         (bin_dir / "loopx-canary").symlink_to(stale_canary_target)
@@ -226,6 +229,10 @@ def main() -> int:
         assert "non-blocking" in install.stderr, install.stderr
         assert "examples/canary/canary-promotion-readiness-smoke.py" in install.stderr, install.stderr
         assert f"- executable: {bin_dir / 'loopx'}" in install.stdout, install.stdout
+        assert (
+            f"- Codex App fallback executable: {bin_dir / 'loopx-apply-rrule'}"
+            in install.stdout
+        ), install.stdout
         assert "- release: " in install.stdout, install.stdout
         assert f"- canary executable: {bin_dir / 'loopx-canary'}" in install.stdout, install.stdout
         assert "- executable compatibility: none" in install.stdout, install.stdout
@@ -255,11 +262,25 @@ def main() -> int:
 
         wrapper = bin_dir / "loopx"
         assert wrapper.is_symlink(), wrapper
+        fallback_wrapper = bin_dir / "loopx-apply-rrule"
+        assert fallback_wrapper.is_symlink(), fallback_wrapper
         assert not (bin_dir / "goal-harness").exists()
         assert (bin_dir / "goal-harness.legacy-disabled").is_symlink()
         assert wrapper.resolve() != REPO_ROOT / "scripts" / "loopx", wrapper.resolve()
         assert wrapper.resolve().name == "loopx", wrapper.resolve()
         release_root = wrapper.resolve().parents[1]
+        assert fallback_wrapper.resolve() == (
+            release_root / "scripts" / "loopx-apply-rrule"
+        ), fallback_wrapper.resolve()
+        fallback_help = subprocess.run(
+            [str(fallback_wrapper), "--help"],
+            cwd=root,
+            env=env,
+            check=True,
+            capture_output=True,
+            text=True,
+        )
+        assert "outer host turn identity" in fallback_help.stdout, fallback_help.stdout
         release_python = release_root / ".loopx-python"
         assert release_python.read_text(encoding="utf-8").strip() == sys.executable
         assert (release_root / "loopx" / "cli.py").is_file(), release_root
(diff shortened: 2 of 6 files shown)


huangruiteng:
## Self-review

### Findings

No blocking finding in the PR diff.

### Assumptions

- A fallback invocation is a child effect of one host turn, so its internal quota receipt must be distinct from the already-settled delivery receipt but stable across retries of that same fallback.
- Receipt identity remains scoped by goal and agent; the fixed-size hash only namespaces the child turn id and does not grant or infer authority.
- `loopx-apply-rrule` is a reserved LoopX command and should be upgraded atomically with the release instead of remaining a user-maintained wrapper pinned to an older snapshot.

- **Motivation:** #3231 strands Codex App on a tight heartbeat cadence because the fallback bridge reuses a settled receipt identity, then hides the structured conflict from the operator.
- **Solved:** the bridge now derives a replay-safe child identity, preserves bounded stdout/stderr diagnostics, and is installed from the same release as `loopx`; focused tests prove both identity and diagnostic contracts, and install smokes prove replacement of a stale regular-file wrapper.
- **User impact:** fallback RRULE apply/ACK can complete after the outer delivery turn settles, and a future failure remains actionable instead of rendering an empty message.
- **Main risk:** installed-user compatibility and scheduler hot-path behavior. The public CLI argument remains accepted, the internal id stays within the existing 128-character public-safe contract, and the installer/readback smokes cover release promotion and wrapper isolation.
- **Design:** this belongs at the host-adapter boundary. The existing domain-local scheduler failure/cooldown reducer remains the source of truth after a real host failure; moving this bridge-local receipt derivation into the generic Effect Program would broaden authority without a second reusable ACK scenario.

### Validation

- 28 focused pytest cases passed for fallback, scheduler hint, and interaction arbitration.
- Standalone install, overwrite, packaged-install, release-wrapper isolation, release-artifact, promotion-boundary, and quota scheduler ACK smokes passed.
- Ruff, shell syntax, Python compile, diff hygiene, and public/private boundary scans passed.
- Risk-based premerge executed 16 checks: 14 passed, with no manual holds. `install-local-smoke.py` exceeded the canary's 120-second wrapper timeout but passed standalone. `monitor-scheduler-contract-smoke.py` fails identically on clean `origin/main@221929602`, proving unrelated baseline drift.
- GitHub dependency review passed. Frontstage Pages fails on two existing strict-link warnings; the current `main` run and its seven predecessors fail the same workflow, and neither warned file is changed here.
- GitHub Python Tests passed in 10m2s, including lint, mypy, CLI qualification, and the full fast suite.

### Merge decision

Self-merged with owner authorization as `fb5bf9838`. Admin bypass applied only to the documented Frontstage baseline failure; no changed or directly affected surface had a failing check.

PR #3366, opened 2026-08-19
feat(desktop): add minimal Tauri control plane shell

steven-kid:
## Summary

- add a thin Tauri 2 shell around the existing LoopX personal workspace dashboard
- start or reuse fingerprinted loopback status and Chat services, then clean up only processes owned by the shell
- route packaged Chat requests to the dedicated loopback service with loopback-only CORS
- keep the existing Goal, Todo, Gate, and Chat authority model unchanged

## Maintainer Refinement

- parse the service probe body as JSON and require the exact top-level LoopX fingerprint; marker-like response headers and nested values no longer pass
- cap every probe response at 1 MiB so an unexpected local listener cannot cause unbounded reads
- document a narrow Dependency Review exception for `GHSA-wrw7-89jp-8q8g`: Tauri 2 still pins Linux GTK3 `glib 0.18.x`, while the affected `VariantStrIter` API is absent from the application and locked dependency call sites; remove the exception when Tauri's Linux runtime reaches `glib >= 0.20`

## Validation

- `cargo fmt --check`
- `cargo test` (4 passed)
- `cargo clippy --all-targets -- -D warnings`
- focused Python Chat CORS and status-server tests (4 passed)
- repository MyPy (16 source files) and Ruff on changed Python files
- dashboard `build:desktop`
- dashboard and desktop `npm audit --audit-level=moderate` (0 vulnerabilities)
- `tauri build --bundles deb`
- inspected the generated 5.9 MB `.deb`, desktop entry, binary, icons, and dynamic library resolution
- Xvfb/Openbox production-binary smoke: dashboard rendered, status/Chat probes passed, second instance reused the first, and ports 8766/8767 closed after window exit
- LoopX public/private boundary scan (0 errors, 0 warnings)
- exact-scope change-quality receipt: `cqr_6a75a505e6a482c2f8a4`

## Known Limits

- AppImage packaging was not completed locally because the Tauri `linuxdeploy` download from GitHub stalled on the current network. The `.deb` package completed successfully; CI or a faster network should exercise AppImage packaging.
- exact-base standard pre-merge canary passed 17 of 18 checks. `hot-path-interface-budget-smoke.py` reports `dashboard_status_json` at 18,305 characters against an 18,200-character budget; the exact same result reproduces on base commit `765220b8244700d16f6b3944a65b1b1d482591a8`, and this PR does not change that payload.

## Security And Authority

- all desktop HTTP services remain bound to `127.0.0.1`
- the WebView can navigate only within its own random loopback asset origin
- no arbitrary shell IPC is exposed to frontend code
- existing preview/apply authority remains in LoopX services


huangruiteng:
Maintainer review — APPROVED at `70c7fa4eaab9b1d5b62e64faace24952af260c18`

I reviewed the desktop shell as a thin, source-built Tauri control-plane client rather than a second authority surface. The review covered service ownership and child-process cleanup, loopback navigation/CSP/permissions, dashboard Chat origin injection, status/Chat CORS, the Rust lockfile, and dependency-review policy.

Two findings were refined on the PR before approval:

- Service reuse no longer accepts response-wide marker substrings. It requires HTTP 200 plus the exact top-level JSON fingerprint for each typed `ServiceKind`, rejects header/nested lookalikes, and caps probe reads at 1 MiB.
- The Tauri 2 Linux `glib` advisory is handled with an exact `GHSA-wrw7-89jp-8q8g` allowlist entry, not a global severity downgrade. The locked dependency path was confirmed, the affected `VariantStrIter` API has no application or locked-source call site, and the exception carries a removal condition for `glib >= 0.20`.

Capability-owned review lenses:

- **Typed state:** service identity is selected by `ServiceKind` and validated through exact JSON fields; marker-like prose cannot classify a service.
- **Domain neutrality:** the shell reuses LoopX Goal/Todo/Gate/Chat contracts without introducing benchmark- or project-specific control-plane obligations.
- **Behavior disclosure:** the desktop README and PR body disclose service reuse, loopback-only startup, experimental/source-built packaging, and the scoped dependency exception.
- **Guidance vs. obligation:** loopback binding, exact service identity, startup failure, and child cleanup are machine-enforced behavior, and are described as such.

Validation:

- `cargo fmt --check`
- `cargo test` (4 passed)
- `cargo clippy --all-targets -- -D warnings`
- Python compile, focused CORS/status tests (4 passed), mypy, and Ruff
- dashboard `npm ci && npm run build:desktop`
- desktop `npm ci && npm audit --audit-level=moderate` (0 vulnerabilities)
- public/private boundary and local-path scans
- risk-based pre-merge canary: 17/18 selected checks passed; the only failure is the unchanged `dashboard_status_json` interface-budget value (18,305 vs 18,200), reproduced at the exact base commit `765220b8244700d16f6b3944a65b1b1d482591a8`. This PR does not touch that payload.
- exact-head change-quality receipt: `cqr_6a75a505e6a482c2f8a4`
- required GitHub checks are green at merge time.

No private material, credentials, raw benchmark evidence, destructive Git behavior, new production authority, or benchmark/scoring semantics are introduced. The remaining `glib` item is the documented, API-unreachable Tauri 2 Linux transitive advisory above; it is not being silently suppressed.

With the owner-authorized self-merge and the bounded base-parity exception documented here, this head is approved for merge.

huangruiteng (approved):
Approved after exact-head maintainer review, scoped security refinement, required checks, and documented base-parity canary evidence.

PR #3374, opened 2026-08-19
fix(dashboard): ship loopback chat bundle

huangruiteng:
## Summary

- align the chat-route smoke with the ESM/Vite contract introduced by #3366;
- make the Vite origin read safe when the same module is executed by the Node smoke;
- regenerate the installed chat bundle so the wheel/PWA no longer carries the pre-#3366 runtime.

This is a release-blocking consistency fix, not a new capability. The source-owned `VITE_LOOPX_CHAT_ORIGIN` contract remains in the dashboard module, the existing smoke now executes that real module under ESM, and the installed artifact is generated by the repository's normal Vite build. The main risks are hidden configuration drift and stale generated assets; the ESM route smoke, complete build, installed PWA smoke, browser smoke, and exact bundle pointer cover those boundaries without adding a second routing implementation.

## Validation

- `npm run build`
- `npm run smoke:chat-route`
- `npm run smoke:pwa-bundle`
- `npm run smoke:personal-workspace`
- `npm audit --audit-level=high` — 0 vulnerabilities
- generated bundle `node --check`
- `python examples/control_plane/check-public-boundary-smoke.py`
- `git diff --check`
- exact change-quality receipt `cqr_f5eb01dc08cc68b1f0a1` — valid, 0 unresolved blockers
- risk-based premerge — 6/6 selected canaries and 3/3 direct checks passed

`npm run smoke:demo-readiness` also passed its launch-agent, promotion-gate, home-route, usage-progress, and home-browser stages, then stopped at the existing ops Decision Freshness selector. Running that exact sub-smoke on base commit `bc1dba7ce9012ba4b2cb89825d81d6672f89ee7e` fails at the same selector, so this is recorded as pre-existing baseline drift rather than attributed to this diff.

## Boundary

No credentials, local paths, private state, benchmark evidence, or new permission surface are included.


huangruiteng:
## Maintainer review

### Findings

No remaining blocker in the final four-file diff.

The release-prep build exposed two concrete issues in the #3366 follow-through: the installed chat bundle still represented the earlier source, and the route smoke's CommonJS compiler could not execute the new `import.meta.env` contract. This PR fixes both without changing the intended origin policy: Vite still supplies the desktop loopback origin, browser/PWA builds retain relative URLs, and the Node smoke observes the same source through ESM with an optional env read.

- **Motivation:** prevent source, smoke, and shipped web-runtime drift before v0.5.0.
- **Solved:** yes; the installed bundle is regenerated from current source and both the route smoke and browser workspace execute successfully.
- **Operator impact:** source-built desktop chat can use its dedicated loopback service while the installed browser/PWA path remains relative.
- **Main risk:** configuration-mode divergence and stale generated assets.
- **Design:** the implementation keeps one source-owned URL function and one generated artifact; it does not add a test-only routing path.

### Validation

- complete Dashboard build;
- chat-route, PWA-bundle, and personal-workspace browser smokes;
- high-threshold dependency audit: 0 vulnerabilities;
- generated JavaScript syntax and public/private boundary scan;
- exact receipt `cqr_f5eb01dc08cc68b1f0a1` valid with no blockers;
- premerge: 6/6 selected canaries and 3/3 direct checks.

The wider demo-readiness lane has a pre-existing ops Decision Freshness selector failure reproduced on the base commit; it is disclosed in the PR body and is not caused by this diff.

**Decision:** self-merge after required GitHub checks pass, under the owner's explicit release authorization.

PR #3420, opened 2026-08-21
fix(control-plane): rotate rearmed replan obligations

huangruiteng:
## Summary

- rotate a vision-gap replan obligation after its prior generation is acknowledged
- keep the new generation deterministic across repeated reads
- retain the gap timestamp while preventing an old ACK from satisfying the rearmed obligation
- verify quota/write-time settlement accepts only the fresh successor identity

## Root cause

Vision-gap obligations derived the same identity after a prior generation had been acknowledged. When the gap rearmed, the stale ACK could still match that reused identity and incorrectly close the new obligation.

## Validation

- `pytest tests/control_plane/test_goal_frontier_replan_rules.py tests/control_plane/test_refresh_state_replan_gate.py` — 62 passed
- exact changed paths: Ruff and `py_compile` passed
- repository strict `mypy` scope — 12 source files passed
- public/private boundary scan — clean
- `loopx canary premerge --from-git-diff --goal-id loopx-meta` — passed; 0 failures, 0 manual holds
- exact-scope quality receipt `cqr_b59ec68bb8d93bf3f8c2` — valid

No private goal state, raw run history, credentials, or generated logs are included.

### DIFF
--- a/loopx/control_plane/goals/goal_frontier/__init__.py
+++ b/loopx/control_plane/goals/goal_frontier/__init__.py
@@ -1333,15 +1333,7 @@ def derive_goal_frontier_replan_obligation_from_summaries(
                 "successor work, update the agent vision, record evidence gap, or "
                 "record no-follow-up"
             ),
-            extra_fields=(
-                {
-                    "rearmed_after_obligation_id": (
-                        rearmed_after_obligation_id
-                    )
-                }
-                if rearmed_after_obligation_id
-                else None
-            ),
+            rearmed_after_obligation_id=rearmed_after_obligation_id,
         )
     if replan_rule.rule is GoalFrontierReplanRule.LONG_TODO_CHAIN:
         assert long_chain_trigger is not None
--- a/loopx/control_plane/goals/goal_frontier/__init__.py
+++ b/loopx/control_plane/goals/goal_frontier/__init__.py
@@ -11,6 +11,7 @@ from ...agents.agent_scope import (
 from ...agents.profile import agent_profile_requires_vision
 from ...agents.runtime_model import peer_work_key, select_peer_for_work
 from ...runtime.time import parse_timestamp
+from ...todos.contract import normalize_todo_replan_obligation_id
 from ...todos.projection import todo_item_is_watch_only_monitor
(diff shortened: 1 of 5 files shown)


huangruiteng:
## Maintainer self-review and pre-merge evidence

No actionable blocker remains after reviewing the exact final diff.

- The fix closes the real lifecycle bug: once a vision-gap obligation generation is acknowledged, a still-open gap re-arms with a deterministic successor identity, so the prior ACK cannot satisfy new work.
- Stable reads remain stable because the successor identity is derived from typed lineage plus the existing trigger identity; the original gap timestamp is preserved.
- Related refinement applied: `rearmed_after_obligation_id` is now an explicit obligation-builder parameter instead of an arbitrary `extra_fields` entry. This keeps lineage ownership at the existing Python replan-builder boundary and makes the next change easier to localize and test.
- TypeScript effect authority was considered. It currently owns effect-program execution, not goal-frontier replan generation, so moving or duplicating this rule there would create split authority rather than improve the boundary.

### Validation

- Focused replan identity suite: **74 passed**.
- Exact changed paths: Ruff passed; `py_compile` passed.
- Repository strict mypy scope: **12 source files passed**.
- `loopx canary premerge --from-git-diff --goal-id loopx-meta`: **17/17 selected checks passed**, no warnings, no manual holds; merge gate passed.
- Exact-scope quality receipt: `cqr_5b1271b3d69a2ec3a038` (valid).
- GitHub Linux `pytest`: passed, including TypeScript control-plane typecheck/tests, Ruff, mypy, CLI output budget, and full fast tests.
- Sign-off, dependency review, and build: passed.
- Public/private boundary scan: clean; no credentials, local paths, private state, raw run history, or generated logs are included.

The Windows job reached **19 passed** and was externally interrupted with `KeyboardInterrupt` in `socket.py`. The owner confirmed that shared Windows-runner failure is being fixed in another PR and is not a blocker for this delivery.

### Merge decision

Self-merge is justified: the runtime behavior is single-purpose, negative paths cover stale-ACK rejection, the related refactor tightens the existing owner boundary, all applicable product gates pass, and the sole red check is an acknowledged independent runner issue.

PR #3435, opened 2026-08-21
fix(dashboard): fence reused services by release

huangruiteng:
## Summary

- expose a public-safe immutable runtime identity from dashboard status and chat capability endpoints
- require the desktop supervisor and CLI launcher to match that identity before reusing an existing local service
- reject stale-but-schema-compatible services with actionable restart guidance
- record the recurring stale-service repair pattern and cover the contract in Python and Rust tests

## Root cause

The existing service handshake checked only the API schema. After an upgrade, an older LoopX service with the same schema could remain bound to the local port and silently receive mutations from the newer frontend. This allowed fixed lifecycle behavior in the new release to be bypassed by the stale process.

## Validation

- `uv run --extra test pytest -q tests/test_dashboard_command.py tests/test_chat_server_cors.py tests/test_status_server_fast_path.py tests/control_plane/test_goal_activation.py` — 44 passed
- `uv run --extra test ruff check ...` — passed
- `cargo fmt --check` — passed
- `cargo test` — 6 passed
- `cargo clippy --all-targets -- -D warnings` — passed
- `loopx canary premerge --from-git-diff --tier standard` — passed; 9 catalog canaries, 8 risk-profile smokes, public-boundary scan, zero failures/warnings/holds
- live regression probe classified an older schema-compatible service as stale

## Safety boundary

LoopX does not kill an unrelated or stale process automatically. It refuses reuse and tells the owner to stop the old dashboard/app service and reopen it. No local registry data, process logs, crash reports, credentials, or private evidence are included.


huangruiteng:
## Self-review closeout

### Findings

- **Resolved before merge — PATH-selected runtimes could silently disable the desktop release fence.** The desktop launcher can fall back to a bare `loopx` command, which `Command` resolves through `PATH`, while the identity reader previously tried to canonicalize that bare name relative to the working directory. A successful service probe could therefore skip the runtime-identity comparison. Commit `6e261e940` resolves the same executable through `PATH` (including common Windows executable extensions) before reading its release manifest, with regression coverage for this exact path.

No unresolved findings remain.

The PR fixes a real installed-user hazard: an older schema-compatible local service could accept mutations from a newer frontend. The immutable release manifest remains the single identity authority; the status/chat endpoints expose only its public-safe projection, and launchers compare rather than recreate that authority. The bounded future-facing pass was applied to the adjacent desktop executable-resolution boundary so the new fence cannot fail open on a supported launch path.

### Validation

- `cargo test` — 7 passed, including the new PATH-resolution regression.
- `cargo fmt --check` — passed.
- `cargo clippy --all-targets -- -D warnings` — passed.
- `loopx canary premerge --from-git-diff --git-diff-base origin/main --tier standard` — passed: 18 executed checks, 0 failures, 0 warnings, 0 manual holds; public-boundary scan clean.
- GitHub checks on exact head `6e261e940` — DCO, dependency review, release build, pytest, Windows PowerShell, macOS desktop, and Windows desktop all passed. Release upload/publish jobs were skipped as expected for a pull request.

### Merge decision

Self-merge approved: the runtime boundary is now complete, validation is green, and no private state, permission boundary, scoring behavior, or destructive migration is involved.

Issue #3082, opened 2026-08-11
[Feature]: wire caller-approved validation into agent self-reported completion


NIU-123370:
### Preflight

- [x] I searched existing issues, discussions, and contributor tasks for this request.
- [x] This request and its attachments contain no credentials or private/internal material.

### Request origin

Observed need in a real workflow

### Problem or unmet outcome

LoopX already ships strong, independent validation primitives:

- `loopx/capabilities/issue_fix/acceptance_loop.py` runs a caller-declared
  `validation_command` via subprocess and requires a real `git diff` before
  `review_ready`.
- `loopx/capabilities/issue_fix/repository_commit_evidence.py` performs real
  git ancestry/remote checks.
- `loopx turn run-once` is specified as "validate independently, write back,
  spend once".

Gap: on paths where an agent self-reports completion with free-text evidence,
the claim is recorded without invoking any caller-approved validation. For
example `complete_task(todo_id, agent_id, evidence: str)` in
`loopx/claude_goal_mode/mcp/loopx_mcp.py` validates the writeback packet
structure but not the truth of the claim. `promotion_gate` is freshness-only and
non-blocking; `operator_gate` is a manual decision recorder.

Failure cost: an agent can record a "done" todo from a free-text claim with no
independent check, unless someone manually runs the issue-fix caller-repo-branch
command.

### Use case

A long-running engineering or issue-fix goal. An agent finishes a todo and
self-reports completion. If that work item has declared a caller-approved
`validation_command` (e.g. `pytest tests/test_x.py`, `ruff check path/`), LoopX
runs it before accepting the completion. On pass, the completion commits; on
fail, completion is blocked and the failure is surfaced. This turns "done" from
a self-report into a verified state, so a hallucinated or premature claim cannot
silently advance the goal.

### Proposed outcome

When an agent self-reports a todo completion and that work item has declared a
caller-approved validation command, LoopX independently runs that validation
before the completion is allowed to commit. On pass the completion proceeds; on
fail it is blocked and the failure reason is surfaced. The desired behavior is
that a "done" todo reflects a verified outcome, not just a claim — achieved by
reusing LoopX's existing validation machinery rather than adding a new validator.

### LoopX area

Control plane (goals, todos, quota, scheduler, registry, runtime)

### Alternatives considered

- Status quo: caller-declared validation only runs when someone explicitly
  invokes the issue-fix caller-repo-branch command; self-reported completions go
  unchecked.
[...]

PR #3436, opened 2026-08-22
fix(todos): gate event-projected completion validation

Duang777:
## Summary
- Event-only todos that declare `validation_command` or `validation_command_argv` were treated as undeclared on `complete_goal_todo`, so the fail-closed completion gate never ran.
- The gate now reads the raw event-sourced projection when the todo is not materialized in markdown. Markdown stays authoritative when `find_todo_block` finds the same todo.
- Public list/status projection still strips the command, argv, label, and timeout down to `completion_validation_required`. The gate does not reuse `event_projection_todo_context`, because that helper performs that strip.

This is the remaining hole after the everyday markdown `todo add` / `todo update --status done` path was already gated (#3082). Everyday `todo add` still writes markdown and was not the broken path.

## Test plan
- [x] `.venv/bin/python -m pytest -q tests/control_plane/test_todo_completion_validation.py tests/control_plane/test_todo_mutation_authority.py` (88 passed)
- [x] `ruff check` on the four changed files
- [ ] CI on this PR

## Notes
- Bounded future-facing pass: shared `_declaration_from_mapping` / `_copy_todo_added_validation_fields`, prefer a later declared snapshot over an earlier undeclared one, and align list projection so argv-only todos also set `completion_validation_required`.
- Out of scope: successor `TODO_ADDED` write-back, `note` projection, `complete_goal_todo` when the markdown state file is missing, and tightening the global public-safe path regex.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/loopx/control_plane/todos/completion_validation.py
+++ b/loopx/control_plane/todos/completion_validation.py
@@ -5,14 +5,21 @@ import subprocess
 from pathlib import Path
 from typing import Any
 
+from ...event_sourced_state import (
+    AppendOnlyStateEventStore,
+    StateEventError,
+    build_state_projection,
+)
 from ...history import load_registry
 from ...materials import find_registry_goal, goal_repo
+from ..goals.active_state_event_projection import state_event_log_candidates
+from ..goals.path_resolution import resolve_goal_local_path
 from ..runtime.validation_command import (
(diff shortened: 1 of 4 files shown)


Duang777:
Closing in favor of #3437 on `fix/event-projected-completion-validation` (same commit; branch name should not use the `codex/` prefix).

PR #3440, opened 2026-08-22
refactor(scheduler): move durable state into TS runtime

huangruiteng:
## Summary

Move scheduler durable-state ownership and the coarse backoff transition into the existing managed TypeScript Effect runtime.

- TypeScript now owns scheduler-state normalization, retention, paths, atomic persistence, replay detection, cadence, and host-transition decisions.
- Python remains a compatibility facade for existing callers and decodes typed runtime results; it no longer recreates scheduler state-machine semantics.
- The hot scheduler-hint path sends one coarse transition request instead of assembling a rule pipeline in Python.
- The implementation reuses the existing Effect runtime, handler registry, atomic JSON writer, and cross-process mutation lock. It does not add a scheduler daemon, cache, or second persistence authority.

## Why this boundary

This is a bounded Strangler Fig step toward the TypeScript control-plane runtime. The cut is organized around one cohesive authority boundary—scheduler state plus its transition—not around file-by-file translation. Existing Python APIs remain stable while their internal source of truth moves once.

The related future-facing refactor is included: the runtime exposes a discriminated scheduler transition result and a single coarse backoff operation. The generic mutation-lock diagnostic is intentionally split into #3441; this PR is temporarily based on that branch and reuses it without owning the wording change. Broader scheduler CLI or host-adapter migration remains out of scope.

## Behavior and compatibility

- Pinned characterization fixtures preserve nine legacy Python normalization/build cases exactly.
- Python ISO timestamps with `Z`, colon offsets, compact offsets, and hour offsets remain accepted.
- Malformed persisted state fails closed as missing state rather than becoming partial authority.
- Same-key writers are lock-serialized and write through atomic rename.
- Repeating the exact same state is an explicit replay and does not rewrite the file.
- A stale writer lock can be reclaimed; runtime crash/retry preserves effect identity and does not repeat a committed write.
- No model-visible packet or prompt contract changes. A model-behavior test would not add a meaningful oracle here; typed parity, runtime lifecycle, and full scheduler tests cover the changed authority directly.

## Validation

- `npm run typecheck:control-plane` — passed
- `npm run test:control-plane` — 57/57 passed
- full scheduler regression — 275 passed
- post-rebase focused scheduler state/transition/backoff/host/execution suite — 226 passed
- managed Effect runtime integration — 11 passed
- scheduler ACK/current-host binding regression — 3 passed
- Ruff and Python compile checks for all changed Python paths — passed
- `uv build` — sdist and wheel passed; wheel was built from the sdist
[...]

PR #3441, opened 2026-08-22
refactor(runtime): generalize mutation lock timeout

huangruiteng:
## Summary

Generalize the shared Effect runtime mutation-lock timeout diagnostic:

```text
Turn journal mutation lock timed out
```

becomes:

```text
Effect runtime mutation lock timed out
```

`withFileMutationLock` is now shared by runtime domains beyond the turn journal. Keeping the old label would misidentify the owner when scheduler state or later Effect handlers time out.

## Scope

- one diagnostic string at the existing shared lock boundary
- no lock, timing, recovery, exception type, or caller behavior change
- no new abstraction, branch, compatibility wrapper, or test-only product hook

This future-facing cleanup is separated from #3440 so the generic runtime contract can land independently and #3440 can focus on scheduler authority migration.

## Validation

- `npm run typecheck:control-plane` — passed
- `npm run test:control-plane` — 46/46 passed on current `main`
- exact-scope change-quality receipt `cqr_8d4db53c50cd88ca46aa` — verified valid
- premerge canary — passed with the project Python 3.13 environment
- diff, DCO, and public/private boundary hygiene — clean

The first premerge invocation selected system Python 3.9 for one nested canary and failed because that interpreter lacks `tomllib`. Running the same canary directly and rerunning the whole gate through the repository's Python 3.13 environment passed. No product code was changed to accommodate the wrong interpreter.

### DIFF
--- a/loopx/control_plane/effect_runtime_io.ts
+++ b/loopx/control_plane/effect_runtime_io.ts
@@ -88,7 +88,7 @@ export async function withFileMutationLock<T>(
       if ((error as NodeJS.ErrnoException).code !== "EEXIST") throw error;
       await reclaimStaleMutationLock(lockPath);
       if (Date.now() >= deadline) {
-        throw new Error("Turn journal mutation lock timed out");
+        throw new Error("Effect runtime mutation lock timed out");
       }
       await new Promise((resolve) => setTimeout(resolve, MUTATION_LOCK_POLL_MS));
     }


huangruiteng:
Exact-head qualification for `9aa0d2fe6`:

- changed surface: shared Effect runtime mutation-lock diagnostic only
- TS typecheck: passed
- TS control-plane tests: 46/46 passed
- exact-scope quality receipt: `cqr_8d4db53c50cd88ca46aa`, valid
- standard premerge canary through repository Python 3.13: passed
- failures/skips: the initial canary invocation exposed one environment-only failure because a nested check selected system Python 3.9 without `tomllib`; the same check passed directly and the complete gate passed through the repository interpreter. No product failure and no product-code workaround.
- manual holds: none
- boundary scan: clean; no private state, local paths, credentials, or generated evidence included

This is enough coverage because the PR changes no lock timing, recovery, exception type, persisted state, or public protocol—only the owner label on an existing error path. The existing complete typed-control-plane suite is the durable regression oracle; a new exact-string test would freeze incidental prose without adding semantic coverage.

Future-facing pass: applied by moving the label to the already-canonical Effect runtime vocabulary. No broader error hierarchy or compatibility seam was added.

PR #3442, opened 2026-08-22
fix(install): recover ownerless release locks

huangruiteng:
## Summary

- recover an interrupted local install when `.install-lock` exists without a usable owner PID
- give a newly-created lock one second to publish its PID before deciding it is ownerless
- preserve valid live-owner locks and keep stale-lock takeover atomic
- add durable coverage for both ownerless-lock recovery and live-owner preservation

## Root cause

An interrupted install can leave the release lock directory behind before its `pid` file is written. The existing recovery path only reclaimed a lock when a valid PID existed and was already dead, so every later installer waited for the full lock timeout on an ownerless directory.

## Validation

- changed Python `py_compile`
- focused ownerless recovery and live-owner preservation assertions
- `examples/release/release-promotion-concurrency-smoke.py` on the final rebased head
- standard premerge selection: direct checks passed; catalog canaries 9/9 passed
- the command harness terminated the monolithic standard run at its five-minute ceiling after catalog completion; the same selected risk-profile commands were then executed individually with the repository Python and passed 8/8
- changed-file public/private boundary scan: clean

No check was skipped. There are no manual holds. The boundary command reported only pre-existing global registry warnings unrelated to these two changed files.

## Safety and scope

- a valid live PID is never stolen
- stale takeover still uses an atomic rename before exact-directory cleanup
- no unrelated release directory, local state, credential, or generated evidence is included
- the bounded future-facing pass was applied at the existing installer lock-owner boundary; no broader installer framework was added

Self-merge is requested under the maintainer-authorized narrow installer/runtime-helper policy after exact-head checks complete.

### DIFF
--- a/examples/release/release-promotion-concurrency-smoke.py
+++ b/examples/release/release-promotion-concurrency-smoke.py
@@ -50,6 +50,25 @@ def release_path(stdout: str) -> Path:
     raise AssertionError(stdout)
 
 
+def write_lock_test_python(path: Path) -> None:
+    path.write_text(
+        "#!/bin/sh\n"
+        "if [ \"${1:-}\" = \"-\" ] && "
+        "{ [ -z \"${LOOPX_INSTALL_GUARD_FILE:-}\" ] || "
+        "[ \"${LOOPX_INSTALL_GUARD_HELD:-0}\" = \"1\" ]; }; then\n"
+        "  cat >/dev/null\n"
+        "  printf '%s\\n' \"$0\"\n"
+        "  exit 0\n"
+        "fi\n"
+        "case \" $* \" in\n"
+        "  *\" promotion-gate \"*) exit 0 ;;\n"
+        "esac\n"
+        "exec \"$LOOPX_TEST_REAL_PYTHON\" \"$@\"\n",
+        encoding="utf-8",
+    )
+    path.chmod(0o755)
+
+
 def assert_install_waits_for_promotion_guard(root: Path) -> None:
     env = {**install_env(root), "LOOPX_RELEASE_ID": "guarded"}
     releases_dir = Path(env["LOOPX_RELEASES_DIR"])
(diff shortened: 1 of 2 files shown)


huangruiteng:
## Maintainer self-review — exact head `b34299651c4e336e0d5329eeafa4b06700b57693`

No actionable findings remain.

### Review lenses

- **State/ownership rule:** the existing installer lock boundary remains authoritative. Missing/invalid/dead owner state is recovered after a bounded PID-publication grace period; a valid live owner remains untouched.
- **Domain neutrality:** behavior and errors remain generic local-install semantics; no DSH-, benchmark-, or project-specific contract leaked into core install code.
- **Behavior disclosure:** the PR explicitly documents the prior full-timeout behavior and the new ownerless-lock recovery behavior.
- **Guidance vs obligation:** lock acquisition and live-owner preservation are machine-enforced and tested, not described as optional guidance.
- **Bounded future-facing pass:** applied at the existing lock-owner boundary, with no duplicate TypeScript/Python authority and no speculative installer abstraction.

### Validation

- shell syntax, changed Python compile, and `git diff --check`: pass
- final-head `release-promotion-concurrency-smoke.py`: pass, including ownerless recovery, live-owner preservation, concurrent release IDs, and failed-candidate rollback
- standard premerge catalog selection: 9/9 pass
- selected install/public-entry risk profile: 8/8 pass when executed with the canary-normalized repository Python
- changed-file public/private boundary: clean
- DCO, dependency review, pytest, and Windows PowerShell: required before merge

The monolithic standard command was terminated by the local command harness at its five-minute ceiling after catalog completion, so the selected risk commands were completed individually. An initial ad hoc direct invocation picked macOS system Python 3.9 and was rejected by the expected Python 3.11+ guard; rerunning with the exact normalized interpreter passed. The first post-rebase full smoke also exposed a test-only cleanup race, which was fixed before this exact head and then passed end to end.

No validation is skipped and there are no manual holds. This is a narrow, public installer/runtime-helper repair with no permission, benchmark, scoring, task-semantic, or release-evidence-policy change; maintainer-authorized self-merge is appropriate once the required exact-head checks pass.

PR #3455, opened 2026-08-22
feat(capability): gate loopx-benchmark skill on benchmark-toolkit enablement

huangruiteng:
## Summary
- Add `skills/loopx-benchmark` playbook (read board → source fence → preregister running → terminal closeout → read comparisons) as the agent-facing benchmark-toolkit workflow.
- Add a capability-gated host-skill lifecycle: `loopx-benchmark` is **not** in the always-installed packaged set; it is installed only when `benchmark-toolkit` is enabled for a Goal and uninstalled when disabled.
- Add goal-scoped builtin capability `enable`/`disable` (`enabled_capabilities` on the Goal registry entry) wired to install/uninstall of the gated skill; `--execute` writes the registry and installs/uninstalls, without it previews.
- Uninstall preserves locally modified gated skills (hash-matched before removal).

## Validation
- `tests/test_workflow_skill_install.py` + `tests/test_packaged_skill_metadata.py` + `tests/extensions/test_external_capability_admission.py` pass (36 tests).
- Public-safe scan on all changed surfaces: clean.
- CLI end-to-end: `capability enable benchmark-toolkit --execute` installs the skill; `capability disable` removes it.

## Scope
Runtime/control-plane behavior + one provider-neutral skill. No benchmark jobs launched, no scoring/runner change for existing benchmarks, no credentials/private state/logs.

### DIFF
--- a/loopx/cli_commands/capability.py
+++ b/loopx/cli_commands/capability.py
@@ -15,6 +15,11 @@ from ..capabilities.catalog import (
 from ..extensions.capability_admission import (
     bind_external_capability_to_goal,
     invoke_external_capability,
+    set_goal_enabled_capability,
+)
+from ..workflow_skill_install import (
+    install_capability_gated_skills,
+    uninstall_capability_gated_skills,
 )
 from ..extensions.runtime import (
     MAX_EXTENSION_REQUEST_BYTES,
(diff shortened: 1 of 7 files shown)


huangruiteng:
Post-merge architecture review found that the useful agent workflow should stay, but the per-Goal built-in toggle should not.

The issue is authority and lifecycle scope: a host skill is installed globally, while `enabled_capabilities` belonged to one Goal. Disabling Goal A could therefore uninstall a skill still needed by Goal B, and the Goal registry write could not commit atomically with filesystem install/uninstall. It also mixed three separate contracts: catalog availability, task-level workflow activation, and runtime execution authority.

Follow-up #3470 keeps `benchmark-toolkit` built-in and catalog-ready, packages `loopx-benchmark` with the normal release-owned host skill set, and makes it task-triggered. External provider bindings still select exact external providers; todo required capabilities still enforce runtime prerequisites. Skill discovery grants no runner, network, credential, private-evidence, or Goal-mutation authority.

The follow-up also fixes the wheel delivery omission, narrows model triggers/lanes, adds doctor/catalog/install/readback coverage, and removes the global install/uninstall race rather than preserving a compatibility facade for a not-yet-released state field.

Validation and the explicit behavior boundary are recorded in #3470.

PR #3458, opened 2026-08-22
fix(update): avoid os.getuid on platforms where it is unavailable

huangruiteng:
## Summary

Follow-up to #3457: `restart_managed_loopx_services()` called `os.getuid()` directly, which does not exist on Windows and broke the added unit test in the `windows-powershell` CI job. Resolve the uid defensively (`hasattr`) and mock it in the test so the helper is platform-neutral.

## Validation

- `python3 -m py_compile loopx/self_update.py tests/test_self_update_runtime_activation.py`: clean.
- `git diff --check`: clean.

### DIFF
--- a/loopx/self_update.py
+++ b/loopx/self_update.py
@@ -563,9 +563,10 @@ def restart_managed_loopx_services() -> list[str]:
             continue
         labels.append(plist.stem)
     restarted: list[str] = []
+    uid = os.getuid() if hasattr(os, "getuid") else 0
     for label in labels:
         result = subprocess.run(
-            ["launchctl", "kickstart", "-k", f"gui/{os.getuid()}/{label}"],
+            ["launchctl", "kickstart", "-k", f"gui/{uid}/{label}"],
             capture_output=True,
             text=True,
             timeout=30,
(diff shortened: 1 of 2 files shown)


huangruiteng:
Post-merge acceptance: no blocking finding remains on current `main`.

Findings and resolution:

- The production change in #3458 is correct: `restart_managed_loopx_services()` is already macOS-gated, and the guarded uid lookup prevents import/runtime failure when `os.getuid` is unavailable.
- The original `windows-powershell` failure was confined to the test fixture: `monkeypatch.setattr(...os.getuid...)` used the default `raising=True` against an attribute that does not exist on Windows.
- #3459 removed that unnecessary monkeypatch. Its exact head passed native `windows-powershell`, Linux `pytest`, DCO, and dependency review.

- This helper is a post-install macOS host adapter for best-effort LaunchAgent restart. It does not own LoopX control-plane state or transitions, so the TypeScript Effect-runtime refactor does not justify moving this platform adapter into the typed state-machine boundary.
- The change is narrowly scoped to managed status/chat LaunchAgents and does not expand repository, production, or cross-agent authority.
- Future-facing pass: no adjacent refactor is needed; the current host/platform boundary is the smallest coherent owner.

Validation:

- Latest `main` at `8fa1ee6bcfef5e2d36190a57bccb19c9d17864cd`.
- `uv run --isolated --extra test python -m pytest -q tests/test_self_update_runtime_activation.py`: `7 passed, 1 skipped`.
- `python3 -m py_compile loopx/self_update.py tests/test_self_update_runtime_activation.py`: passed.
- `git diff --check` for both #3458 and #3459 merge commits: passed.
- Public/private boundary scan on the two touched paths: no non-public material introduced.

Acceptance decision: treat #3458 + #3459 as the delivered unit. #3458 alone had an incomplete Windows test fix; the merged follow-up closes that gap with exact native-Windows evidence.

PR #3464, opened 2026-08-22
refactor(turn): cut settlement over to a coarse TypeScript transaction

huangruiteng:
## Summary

- cut Turn settlement over from seven leaf effect-runtime calls to one coarse `turn.settlement.reduce` transaction contract
- use a fail-closed two-phase boundary while real providers remain in Python: TypeScript preflight authorizes the exact pending effects, Python executes/checkpoints only those effects, and TypeScript performs the final reduction
- join terminal closeout into the same settlement transaction, remove six obsolete leaf handlers plus their Python facades/tests, and update the bilingual migration RFC

## Product and architecture judgment

The previous leaf-RPC shape made Python reconstruct Effect Program control flow across runtime calls. This change moves identity validation, replay/prefix validation, provider ordering, failure classification, receipt composition, and canonical result projection into the typed TypeScript owner.

The first draft used only a final post-provider reduction. Self-review rejected that shape because malformed identity or journal state could be discovered after a real provider side effect. The final protocol therefore uses fail-closed preflight before any still-Python provider runs, then a final reduction over durable checkpoints. Fully replayed settlement needs one RPC; new provider work needs two. The facade can collapse to one request only after those provider effects move into the TypeScript runtime.

## Migration economics

- final scope: 13 files, `+843 / -860` overall
- product code: `+575 / -519` (`+56`); tests: `+225 / -296` (`-71`); RFCs: `+43 / -45` (`-2`)
- cross-runtime calls: new work `7 -> 2`; fully replayed work `7 -> 1`
- deleted authority: six leaf runtime handlers, Python leaf orchestration facades, and obsolete facade-level tests
- retained bridge: one thin Python transaction adapter while actual provider side effects remain Python-owned

Warm 100-iteration local measurement:

| path | p50 | p95 |
| --- | ---: | ---: |
| previous seven-RPC baseline | 18.255 ms | 53.059 ms |
| new provider-work path | 0.810 ms | 1.710 ms |
| new replay path | 0.448 ms | 0.683 ms |

## Validation

- `npm run typecheck:control-plane -- --pretty false`
- `npm run test:control-plane` — 81/81 passed
- focused Python cross-runtime suite — 254 passed
- changed-path Ruff — passed
- repository Mypy surface — 12 source files passed
- wheel + sdist build — passed; both include `turn_driver/settlement.ts`
- fresh supported-Python wheel install + `runtime.ping` — passed
[...]

PR #3465, opened 2026-08-22
fix(control-plane): make replan settlement binding explicit

huangruiteng:
## Summary

- project a TS-owned, domain-local `replan_settlement_contract_v0` that separates the one causal settlement binding from the semantic replan obligation
- keep Todo-bound replan writeback and spend on `--todo-id` only; keep Todo-less replan directly bound to `--replan-obligation-id`
- project the complete material refresh-then-spend chain only when a turn-scoped settlement plan makes it executable
- retain the executable contract in the hot-path Turn envelope and document the invariant in the protocol and self-repair catalog

## Diagnosis

The protocol was not truly deadlocked: the original quota receipt always had one valid settlement identity, and the Todo-only writeback/spend path succeeded. The failure mode was projection ambiguity. A selected Todo id and a replan obligation id appeared together without a typed relationship, so weaker agents treated both as settlement flags and entered trial-and-error.

The adjacent implementation also used `replan_settlement_bound` for two different questions: whether the semantic obligation itself owned the receipt, and whether an accountable settlement chain was ready. That caused Todo-bound autonomous replan actions to omit delivery fields and the projected spend even though `spend_after_validation=true` and a Todo-bound settlement plan existed.

## Architecture

This is intentionally a domain-local `work_items/replan_settlement.ts` contract exposed through the existing TS runtime. It reuses the generic settlement identity invariant but does not expand the generic Effect Program with replan-specific ACK semantics. Python adapts the typed result and does not create a second binding rule.

The branch is rebased over #3464. That coarse TS Turn transaction reducer owns transaction order, receipt reduction, and replay; this projector owns only the replan-domain choice between semantic target and causal settlement binding. Keeping those layers separate avoids putting replan semantics into the generic transaction algebra.

## Behavior change

For a turn-scoped `autonomous_replan` with a selected Todo, `next_cli_actions` now contains:

1. one material `refresh-state` bound only to the selected Todo;
2. one matching `quota spend-slot` bound to the same Todo and Turn.

Todo-less autonomous replan remains directly obligation-bound. Dual-binding settlement remains invalid. Unscoped quota, diagnose, and turn-plan reads retain compact non-executable guidance instead of advertising a settlement contract or spend without a Turn identity.

## Validation

- Ruff and diff checks
- TypeScript typecheck and all 84 native control-plane tests
- 53 focused Python tests covering the TS facade, settlement algebra, Todo-bound/Todo-less real CLI chains, unscoped compact projection, and in-flight continuation
- agent-facing CLI output base/head budget regression smoke
- `loopx canary premerge --from-git-diff`: passed 18/18 selected checks
  - changed surfaces: control plane, public protocol docs, self-repair catalog
  - failures/skips/manual holds: 0/0/0
  - hot-path interface budget and maintainability ratchet: passed
- GitHub DCO, dependency review, release/frontstage builds, Linux pytest, and Windows lifecycle checks: passed

The validation is sufficient because it covers the typed owner, Python runtime boundary, hot-path projection, real refresh/spend receipts for both binding modes, compact unscoped reads, the prior in-flight vision-continuity repair, and repository-wide risk-selected canaries.


huangruiteng:
CI follow-up: the first pytest run correctly caught an agent-facing CLI budget regression. The initial projection exposed the full replan settlement contract and refresh/spend chain even on unscoped diagnostic reads. The refined boundary now keeps TypeScript as the domain authority, but projects the executable contract and ordered chain only when a turn-scoped settlement plan exists. Todo-bound and obligation-bound turns both retain the full chain; unscoped reads keep compact non-executable guidance.

Validation on exact head 4f126c91a:
- 53 focused Python settlement/CLI tests passed
- CLI output budget regression smoke passed
- Ruff and diff checks passed
- risk-based premerge canary passed 18/18, including hot-path interface budget, maintainability ratchet, and public-boundary checks
- manual holds: 0

Future-facing pass: applied at the bounded domain edge by separating binding kind from settlement-chain readiness. The generic Effect Program remains unchanged.

PR #3362, opened 2026-08-19
fix(goal): project terminal Todo settlement actions

huangruiteng:
## What changed

- project exact completed Todo ids and completion turn keys for terminal succession gaps;
- prefer direct `todo complete --no-follow-up` settlement when no real work remains;
- keep typed successor creation as the path only when concrete implementation or validation work remains;
- return lifecycle-specific CLI actions before re-reading quota, instead of routing through generic semantic refresh and quota spending;
- add regression coverage for the obligation projection and interaction hot path.

## Why

A completed advancement Todo without a successor is a lifecycle decision, but the interaction projection previously exposed only generic replan guidance. Agents could spend unnecessary turns re-reading the control contract or create a lifecycle-only filler successor even when the correct outcome was a direct no-follow-up settlement.

This change makes the terminal path explicit and bounded while preserving genuine technical replanning.

Fixes #3358.

## Validation

- `167 passed` across focused goal, quota, Todo authority, settlement, and host-context tests;
- `ruff check --select F` on all changed Python files;
- `git diff --check`;
- `loopx canary premerge --from-git-diff`: standard tier, 17/17 selected canaries passed, no manual holds;
- public/private scan found no credentials, private paths, raw trajectories, or verifier output.

## Scope

This changes Goal control-plane projection and interaction guidance only. It does not change benchmark scoring, task semantics, permissions, or runner launch behavior.

PR #3484, opened 2026-08-22
fix(goal): project terminal Todo settlement on refresh

### LINKED ISSUE #3473: bug(goal): terminal Todo closeout can bypass direct settlement projection
## Summary

Follow-up to #3358 / #3362: current `main` correctly projects exact
`todo complete --no-follow-up` settlement commands when the agent reruns
`quota should-run`, but a normal terminal work turn can still bypass that
projection and loop on `refresh-state`.

The observed sequence on current `main` was:

1. create two task-specific advancement Todos;
2. implement, validate, and commit the change;
3. complete both Todos without a successor or `--no-follow-up`;
4. follow the usual closeout route directly into `refresh-state`;
5. receive the generic typed-semantic-delta rejection repeatedly.

Because no post-completion quota read occurred, the exact settlement actions
added by #3362 were never delivered. The Goal remained active after a clean
commit while the agent retried lifecycle commands instead of terminating.

## Why this is distinct from #3358

#3362 repaired the quota projection. It did not close the transition seam
between `todo complete` and `refresh-state`:

- `todo complete` can create a `completed_advancement_without_successor`
  obligation but does not return the resulting exact settlement action;
- `refresh-state` rejects the write but reports only generic semantic-delta
  guidance;
- the useful exact Todo ids and `--no-follow-up` commands are available only
  after a separate quota read that the terminal path does not require.

This is especially easy to trigger when guided planning creates sibling
implementation and validation Todos. Completing both produces two succession
gaps even though the task itself is terminal.

## Expected behavior

Keep the strict succession invariant, but make one real terminal path bounded
and machine-executable. Any one cohesive design is acceptable, for example:

- when `todo complete` creates a succession obligation, return an explicit
  next action requiring `quota should-run` before `refresh-state`; or
- project the exact direct-settlement commands in the `refresh-state`
  rejection packet; or
- make the host closeout driver automatically re-enter quota after Todo
  lifecycle transitions and before semantic refresh.

The agent should not need to infer a new technical replan, inspect CLI help, or
repeat generic refresh attempts when the only missing fact is terminal Todo
settlement.

## Synthetic reproduction

1. Start a Goal with one registered agent.
2. Add two advancement Todos representing implementation and independent
   validation, without successor links.
3. Complete both normally.
4. Run `refresh-state --classification validated_progress` without an
   intervening quota call.
5. Observe the generic replan rejection and absence of exact direct-settlement
   commands in that response.

## Acceptance criteria

- A terminal agent that completes the last Todo can reach the #3362 direct
  settlement route without guessing that an extra quota read is required.
- The machine packet names exact completed Todo ids and preserves completion
  turn keys where required.
- No technical replan, filler successor, quota spend, or repeated semantic
  refresh is required for lifecycle-only settlement.
- Genuine runnable-successor and semantic-replan paths remain strict.
- A focused regression covers the real
  `todo complete -> refresh-state` transition, not only a standalone quota
  projection.

huangruiteng:
## What changed

- keep `refresh-state` strict when terminal advancement Todos still need lifecycle settlement, but return a typed `replan_transition` with the exact completed Todo ids, completion turn keys, and direct `todo complete --no-follow-up` actions;
- preserve `completion_turn_key` through the real Todo summary path instead of losing it during compaction;
- move lifecycle-settlement reentry projection from Python string assembly into the TypeScript Effect runtime, with Python limited to typed adaptation and result-shape validation;
- append the exact transition actions to the human-readable error while preserving the structured JSON packet;
- cover the real `todo complete` twice -> `refresh-state` rejection path and the TS/Python runtime boundary.

## Why

The direct settlement semantics from #3362 were only visible after another quota read. A normal terminal turn could complete its last Todos and call `refresh-state`, which reduced the new lifecycle obligation to a generic semantic-delta error. That made the model guess an extra quota read or loop on refresh even though the control plane already knew the exact settlement actions.

This change closes that projection seam without weakening successor or semantic-replan enforcement. Genuine remaining work still requires a real runnable successor; terminal lifecycle settlement does not spend quota or require another semantic refresh.

## Validation

- TypeScript control-plane suite: 87 passed;
- focused and adjacent Python suites: 194 passed;
- focused post-refinement Python suites: 78 passed;
- TypeScript typecheck passed;
- Ruff undefined-name scan and `git diff --check` passed;
- `loopx canary premerge --from-git-diff`: standard tier, 17/17 selected checks passed, no failures, skips, or manual holds;
- public/private scan found no credentials, private paths, raw logs, or local state in the candidate paths.

## Scope

This changes Todo lifecycle/replan projection and its CLI rejection packet. It does not change quota accounting, Todo succession strictness, permissions, benchmark behavior, or runner launch behavior.

Fixes #3473.

### DIFF
--- a/loopx/cli_commands/project_lifecycle.py
+++ b/loopx/cli_commands/project_lifecycle.py
@@ -25,6 +25,10 @@ from ..control_plane.work_items.delivery_batch_scale import (
 )
 from ..control_plane.work_items.delivery_outcome import DELIVERY_OUTCOME_CHOICES
 from ..control_plane.work_items.progress_observation import ProgressResultClass
+from ..control_plane.work_items.semantic_replan_writeback import (
+    ReplanWritebackRejected,
+    project_replan_writeback_rejection,
+)
 from ..extensions.lark.goal_channel_lifecycle import (
     goal_channel_gate_sync_failure,
     sync_human_gate_after_refresh,
@@ -694,6 +698,16 @@ def handle_project_lifecycle_command(
                 "dry_run": bool(args.dry_run),
                 "error": str(exc),
             }
+            if isinstance(exc, ReplanWritebackRejected):
+                transition = project_replan_writeback_rejection(
+                    exc,
+                    goal_id=args.goal_id,
+                    agent_id=args.agent_id,
+                )
+                payload["replan_transition"] = transition
+                payload["error"] += " Required transition: " + "; ".join(
+                    transition["next_cli_actions"]
+                )
         projected_capabilities = runtime_capabilities_for_cli_projection(
             args.available_capabilities
         )
--- a/loopx/control_plane/effect_runtime_handlers.ts
+++ b/loopx/control_plane/effect_runtime_handlers.ts
@@ -63,7 +63,10 @@ import {
   evaluateDeliveryContinuity,
 } from "./turn_driver/delivery_continuity.ts";
 import { reduceTurnSettlementTransaction } from "./turn_driver/settlement.ts";
-import { projectReplanSettlementContract } from "./work_items/replan_settlement.ts";
+import {
+  projectReplanSettlementContract,
+  projectTodoLifecycleSettlementReentry,
+} from "./work_items/replan_settlement.ts";
 
 type EffectRuntimeHandler = (params: JsonObject) => unknown | Promise<unknown>;
 
@@ -376,6 +379,10 @@ export function createEffectRuntimeHandlers(
     ],
     ["turn.settlement.reduce", reduceTurnSettlementTransaction],
     ["work_item.replan_settlement.project", projectReplanSettlementContract],
+    [
+      "work_item.replan_settlement.reentry",
+      projectTodoLifecycleSettlementReentry,
+    ],
   ]);
 }
 
--- a/loopx/control_plane/todos/todo_summary.py
+++ b/loopx/control_plane/todos/todo_summary.py
@@ -501,6 +501,7 @@ def compact_todo_item(item: dict[str, Any]) -> dict[str, Any]:
         "evidence",
         "reason",
         "completed_at",
+        "completion_turn_key",
         "updated_at",
         "superseded_by",
         "completion_validation_required",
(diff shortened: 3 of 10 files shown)


huangruiteng:
Maintainer self-review for exact head `3845efa3c8b5c1d163743f1dfc4f65f6d02d236a`:

- **Changed surfaces:** terminal Todo lifecycle compaction, autonomous-replan writeback rejection, CLI projection, the effect runtime registry, and the new typed TypeScript settlement-reentry boundary.
- **Behavioral result:** when a terminal Todo creates an unresolved settlement obligation, a rejected `refresh-state` now returns the exact no-follow-up settlement action immediately. It preserves the Todo id and completion-turn identity and does not send the agent back through quota/read/refresh loops. Strict succession and settlement matching remain enforced.
- **Bounded TypeScript refactor:** lifecycle-trigger interpretation, causal identity checks, shell-safe action rendering, and effect registration now have one TypeScript owner. Python remains the Todo persistence/CLI adapter and validates the TypeScript result shape and identities. I deliberately did not widen this PR into a Todo persistence rewrite because that is a separate state-ownership migration without a second active call site.
- **Direct validation:** TypeScript typecheck passed; full TypeScript control-plane suite passed (87/87); broad related Python suite passed (194); final focused Python suite passed (78); Ruff undefined-name/import checks and `git diff --check` passed.
- **Risk-based premerge canary:** standard tier passed 17/17: catalog canaries 9/9 and risk-profile smokes 8/8. No failures, skips, or manual holds.
- **GitHub checks:** DCO, dependency review, release build, Linux pytest, and Windows PowerShell all passed. `upload-release` and `publish-pypi` are the expected non-release skips.
- **Boundary review:** candidate paths contain no credentials, private/local state, raw logs, local paths, or generated validation artifacts. The change is domain-neutral and does not change permissions, scoring, releases, or benchmark semantics.

Coverage is sufficient because it exercises both the TypeScript authority and Python adapter, the original `refresh-state` failure path, causal-identity mismatch negatives, default human output, JSON projection, and cross-platform CI. No unresolved review findings remain. Admin self-merge is appropriate under the owner's explicit authorization and the repository's runtime-helper self-merge policy.

PR #3489, opened 2026-08-22
fix(ssh-source): quick-add establishes the tunnel and avoids the local port

huangruiteng:
## Summary

The reported "切回本机又自动切回 ark-devbox" is caused by the dropdown quick-add path for configured SSH hosts:

- `quickAddConfiguredHost` never called `ensureSshSource`, so no tunnel was opened.
- It picked port `8876` as "free" — but `8876` is already the local source's URL, so the added "ark-devbox" source pointed at `127.0.0.1:8876` (the local feed). "本机" and "ark-devbox" then shared the same URL and the selector label flipped between them.

Fix:

- `quickAddConfiguredHost` now skips every port used by any source (including the local source), picks a genuinely free tunnel port, calls `ensureSshSource` (open tunnel + start remote `serve-status`), then adds and loads the source.
- Selecting an existing SSH source also best-effort calls `ensureSshSource` before loading, so a dropped tunnel is re-established on switch.
- Refresh the packaged `/chat/` bundle.

## Validation

- `npx tsc --noEmit` (dashboard): clean.
- `git diff --check`: clean.
- Repro before: selecting `configured:ark-devbox` produced an `ssh-<hash>` source on the local port and "不可用". After: it should create a distinct source on a free port with a working tunnel.


huangruiteng (commented):
**Approval conclusion (author-owned PR; GitHub blocks formal self-approval)**

### 详细中文评审（head `HEAD`）

**动机**：用户“切回本机又自动切回 ark-devbox”。复现定位：下拉“已配置 SSH Host（点击快速添加）”的 `quickAddConfiguredHost` 从不调用 `ensureSshSource`，且把 8876（本机同一端口）当作“空闲”端口，导致新增的“ark-devbox”源实际指向本机 8876，两个源 URL 相同，选择器标签来回闪。

**改动**：quick-add 跳过所有已用端口（含本机）、选真正空闲端口、先 `ensureSshSource`（建隧道+远端 serve-status）再 add+load；对已存在的 SSH 源，`onSelect` 也先 best-effort ensure 再加载；刷新打包 bundle。

**风险**：低，前端源切换逻辑；后端/控制面不变。

**结论**：命中根因，逻辑修复 + bundle 刷新，CI 全绿。建议合并。

### English verdict

`English verdict: APPROVE (self-merge). This fixes the SSH quick-add to open a real tunnel on a free port (not the local port), so selecting a configured host no longer flips between 本机 and the SSH label on the same URL; full CI green.`

PR #3490, opened 2026-08-22
fix(ssh-source): stop the route effect from re-triggering source switches

huangruiteng:
## Summary

The "切回本机又自动切回 ark-devbox" back-and-forth was reproduced in the browser: selecting an SSH source started its load, while the route `useEffect` kept firing a competing `loadFromUrl` for the previously loaded (or default local) source. Both navigated the URL, so the active source and `statusUrl` alternated every ~3s.

Fix in `dashboard-page.tsx`:

- The route effect now skips when an in-flight selection targets a different URL, and skips loading the default local source while a selection is pending.
- `loadFromUrl` applies the projection-revision guard to foreground responses too (it already did for background), so a stale load can no longer navigate/setSource after a newer selection started.
- Refreshed the packaged `/chat/` bundle.

## Validation

- `npx tsc --noEmit`: clean.
- Browser repro before: active source and URL alternated between 本机(8766) and ark(8877) every ~3s. After: after selecting a configured SSH host, the active source stays stable for 30s+.
- `git diff --check`: clean.


huangruiteng (commented):
**Approval conclusion (author-owned PR; GitHub blocks formal self-approval)**

### 详细中文评审（head `HEAD`）

**动机**：浏览器复现确认“切回本机又自动切回 ark-devbox”是路由 `useEffect` 与用户选择竞争：选 SSH 源后，effect 仍会对“上一个/默认本机源”再发一次 `loadFromUrl`，两者互相导航，active/URL 每 ~3s 来回闪。

**改动**：effect 在“已有进行中且 URL 不同的选择”时跳过、且有选择进行中时不加载默认源；`loadFromUrl` 对前台响应也套用 revision 守卫（原本仅 background），过期响应不再导航/报错；刷新打包 bundle。

**风险**：低，仅前端源切换状态机；不触碰后端/控制面。

**结论**：修复实测有效（选择 SSH 源后 30s+ 稳定），CI 全绿。建议合并。

### English verdict

`English verdict: APPROVE (self-merge). This stops the route effect from re-triggering competing source loads; browser repro confirms the active source stays stable after switching; full CI green.`

PR #3493, opened 2026-08-22
fix(todo): make rejected replan reentry executable

### LINKED ISSUE #3473: bug(goal): terminal Todo closeout can bypass direct settlement projection

huangruiteng:
## Summary

Fixes #3473 by making the Todo lifecycle actions projected after a rejected `refresh-state` executable by the real CLI and completion fence.

- separates local Todo completion identity from quota turn settlement identity
- derives a stable TS-owned `local_completion_<digest>` identity for unscoped completions
- adds a narrow `--completion-identity-key` CLI path for projected `--no-follow-up` lifecycle reentry
- permits legacy null-identity repair only for the exact TS-derived key on an already completed Todo
- preserves receipt-backed `--turn-instance-id` behavior for quota-scoped settlement
- executes projected actions end-to-end in regression tests, then verifies terminal quota exit
- adds a real Ark/Doubao model-behavior scenario for rejection -> two lifecycle settlements -> terminal quota -> stop

## Ownership and refactor boundary

Completion identity derivation, fence legality, and reentry action projection remain TypeScript-owned. Python adapts the typed contract and persists the resulting identity; it does not recreate the state rule. A bounded companion refactor moves identity normalization and decision-target lookup out of the hot `loopx/todos.py` facade, keeping that module at its maintainability ceiling instead of adding new debt.

This remains a domain-local Todo reducer/ACK contract rather than a new generic Effect Program abstraction: no second external effect or ordered receipt chain is introduced.

## Validation

- 178 focused Python control-plane tests passed
- TypeScript control-plane typecheck passed
- 92 TypeScript control-plane tests passed
- live weak-model qualification (`doubao-seed-2-1-turbo-260628`) passed 2/2 attempts on commit `6b5714373`
  - exact sequence each time: two Todo lifecycle settlements, terminal quota read, then stop
  - completion identity preserved; no repeated refresh or quota spend
- `loopx canary premerge --from-git-diff --tier standard` passed
  - 18/18 selected checks executed
  - 0 failures, 0 warnings, 0 manual holds
  - maintainability ratchet and public/private boundary scan passed

### DIFF
--- a/docs/project-agent-todo-contract.md
+++ b/docs/project-agent-todo-contract.md
@@ -360,10 +360,17 @@ the agent should do one of two things:
 
 This succession decision is durable Todo state. A later progress observation,
 vision ACK, coverage-exhausted result, or rewritten rationale cannot substitute
-for it. If the Todo is already complete without either branch, use the supported
-same-turn `todo complete --no-follow-up` recovery when its completion identity
-is still current; otherwise add/link a real successor. Do not create a user
-gate merely to silence a succession warning.
+for it. Every new completion therefore retains an opaque completion identity.
+A quota-bound completion still permits only the receipt-backed same-turn
+`todo complete --no-follow-up` recovery. An ordinary unscoped completion gets a
+stable `local_completion_*` identity; if a later `refresh-state` discovers that
+the finished Goal has no real successor, its typed rejection may project
+`--completion-identity-key` for one direct lifecycle reentry. That command is
+valid only for the exact already-completed Todo, its matching local identity,
+`active_goal` continuation, no successor, and the authorized lifecycle actor.
+It cannot be supplied for an open Todo or used as a quota turn identity.
+Otherwise add/link a real successor. Do not create a user gate merely to
+silence a succession warning.
 
 This keeps the active checklist honest without making LoopX a heavyweight
 project-management state machine.
--- a/docs/reference/protocols/loopx-turn-v0.md
+++ b/docs/reference/protocols/loopx-turn-v0.md
@@ -377,17 +377,26 @@ Turn journal records that outcome before quota spend. A Todo completion alone
 never terminates the goal; a fresh decision owns successor selection and goal
 termination.
 
-Every newly completed Todo persists `completion_continuation` explicitly. The
+Every newly completed Todo persists `completion_continuation` and an opaque
+completion identity explicitly. The
 value must agree with its durable relations: `successor` requires at least one
 `successor_todo_id`, `no_followup` requires `no_followup=true`, and
 `active_goal` requires neither. A completed record that omits the field is not
 interpreted as `active_goal`; it fails closed until an agent explicitly repairs
 it by replaying `loopx todo complete`. The only post-completion transition is
-the narrow #3261 recovery seam: during the original `completion_turn_key`, an
-explicit `active_goal` may be upgraded to `no_followup` after the matching
-writeback and spend receipts exist. The recovery records
-`completion_recovery=same_turn_terminal_closeout`; it cannot cross a Turn or
-replace a successor.
+the narrow #3261 recovery seam: during the original quota-bound
+`completion_turn_key`, an explicit `active_goal` may be upgraded to
+`no_followup` after the matching writeback and spend receipts exist. The
+recovery records `completion_recovery=same_turn_terminal_closeout`; it cannot
+cross a Turn or replace a successor. A completion made outside a quota Turn
+instead persists a TS-derived `local_completion_*` identity. When a strict
+`refresh-state` rejection later proves that only Todo lifecycle settlement is
+missing, it may project that exact key through `--completion-identity-key`.
+The completion fence accepts this distinct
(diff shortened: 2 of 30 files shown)


huangruiteng:
Maintainer self-review / exact-head qualification (`6b5714373`)

Changed surfaces
- TS-owned Todo completion identity, completion fence, completion state, and replan reentry projection
- Python CLI/adapter/event-writeback persistence boundary
- public Todo/Turn/model-behavior protocol docs
- deterministic and live model-behavior qualification

Review lenses
- Typed state: `turn_settlement`, `unscoped_completion`, and `lifecycle_reentry` are distinct; legacy repair accepts only the exact goal/Todo-derived identity on an already-done Todo. A local-shaped but non-derived turn key remains turn-scoped.
- Domain neutrality: the contract is Todo lifecycle-local and contains no benchmark/product-specific semantics.
- Behavior disclosure: ordinary unscoped completions now persist a stable local completion identity; docs and CLI help disclose the new reentry-only argument.
- Guidance vs obligation: projected settlement identity and terminal fence checks are machine-enforced; the human-readable no-successor clause remains an explicit branch, not a receipt.
- Effect Program scope: no generic Effect Program expansion; this has one local durable mutation/ACK authority and no ordered multi-provider effect chain.
- Reuse/simplification: the existing TS effect runtime owns identity projection. The first canary run found `loopx/todos.py` over its module budget; a bounded same-domain refactor moved identity normalization and decision-target lookup to their owning modules. Final maintainability ratchet passes and the facade stays at its reviewed ceiling.

Validation
- 178 focused Python tests: passed
- TS typecheck: passed
- 92 TS control-plane tests: passed
- weak-model live qualification: 2/2 passed, 8 provider calls total; each attempt observed `todo_lifecycle_settlement`, `todo_lifecycle_settlement`, `terminal_quota_reentry`, then stopped
- standard premerge canary: 18/18 checks executed, 0 failures, 0 warnings, 0 manual holds
- DCO trailers present on all commits

Non-gating tooling note
- An exploratory `uvx ruff check` was not used as a project gate: the repository has no Ruff configuration and the command reports existing broad legacy style/policy findings across touched modules. No formatter or auto-fix output was retained. Repository-native compile, tests, maintainability ratchet, canary, and boundary checks all pass.

Future-facing pass: applied only at the adjacent Todo ownership boundary; broader facade or Effect Program migration was explicitly deferred as unrelated.

PR #3501, opened 2026-08-22
fix(dashboard): fence local and SSH source switches

huangruiteng:
## Summary

- fence the complete status-source transaction from user selection through SSH preparation, fetch, payload commit, and route commit;
- make the dashboard the single owner of SSH source selection and remove duplicate child-side ensure state;
- reject stale foreground/background responses and prevent an older route commit from clearing a newer pending selection;
- add deterministic development and packaged-browser regressions, then refresh the shipped `/chat/` bundle.

## Root cause

The foreground loader captured `statusProjectionRevisionRef.current` before incrementing it, so every successful foreground response immediately failed its own freshness check. That left the previous Goal projection visible and the request pending.

Freshness also began only after asynchronous SSH preparation. A slower earlier selection could therefore start its fetch after a later local/SSH selection. Finally, an older request could finish `navigate()` and clear the pending state belonging to a newer selection. The fix binds all of those phases to one monotonic typed request fence and keeps projection revisions for payload/background invalidation.

## Validation

- dashboard TypeScript compile plus desktop and packaged chat production builds;
- status-source catalog smoke;
- deterministic status-source race browser smoke in development and packaged modes;
- personal-workspace browser acceptance smoke;
- packaged PWA bundle smoke;
- repository TypeScript typecheck and 97 control-plane tests;
- changed-file public/private boundary scan: 9 files, 0 findings;
- `loopx canary premerge --from-git-diff --goal-id loopx-meta`: 10/10 selected checks passed, 0 failures, 0 warnings, 0 manual holds.

Change-quality receipt: `cqr_474bf80e7e2a6fa68335` (`pass`, exact current base and committed scope).

## Scope review

The related future-facing pass was applied in this PR: SSH preparation and selection share one parent-owned transaction boundary, while the switcher only gathers typed input. The race smoke reuses the repository browser support. No broader source store or router abstraction was added.


huangruiteng:
## Maintainer review

### Findings

No remaining correctness, compatibility, privacy, or merge blockers found in the final head.

None. The fix keeps existing status-source URLs and catalog persistence compatible; it only changes which asynchronous request generation may commit UI and route state.

- **Motivation:** local/SSH switching could strand the Goal projection or let a slower earlier selection overwrite a newer user choice.
- **Solved:** the final diff binds selection, optional SSH preparation, fetch, payload projection, and route commit to one typed request fence. Background reconciliation may only commit to the currently loaded source.
- **User impact:** switching to local, Remote A, or Remote B now converges on the last explicit selection without stale Goal lists or A↔B bounce-back.
- **Main risk:** async ordering in the dashboard hot path and parity between development and the packaged `/chat/` bundle.
- **Design judgment:** the page is the correct transaction owner; removing switcher-owned SSH effects avoids a second authority. The bounded future-facing pass was applied by extracting the typed fence and sharing existing browser-smoke infrastructure rather than adding a new source-store framework.

### Validation

- dashboard TypeScript compile plus desktop and packaged chat production builds;
- deterministic source-switch browser smoke in development and packaged modes;
- status-source catalog, personal-workspace browser, and PWA bundle smokes;
- repository control-plane typecheck and 97 local tests;
- exact public-boundary scan: 9 changed files, 0 findings;
- exact change-quality receipt `cqr_474bf80e7e2a6fa68335`: pass, 0 blockers/warnings/advisories;
- premerge: 10/10 selected checks passed, 0 failures, 0 warnings, 0 manual holds;
- remote DCO, dependency review, Frontstage, release build, macOS/Windows desktop, Windows PowerShell, pytest, and SonarCloud all passed;
- SonarCloud reports 0 new issues and 0.0% duplication on new code;
- the first Frontstage run was cancelled by workflow concurrency after its earlier steps passed; the failed/cancelled jobs were rerun and passed. Publish/deploy jobs skipped as expected for a PR.

Coverage is sufficient because the regression test controls both race boundaries directly: a delayed status response cannot overwrite a newer source, and a delayed SSH ensure cannot start a superseded status request. The same assertions run against source and shipped bundle.

### Merge decision

`self-merge`: allowed by the owner request and the repository gate (`self_merge_allowed=true`); all required checks are green.

PR #3522, opened 2026-08-23
Keep active task leases when the registered-agent roster changes

lightencc:
## Motivation

TaskForge batch 3a: LoopX cannot yet serve as a multi-agent execution kernel while a running roster change silently disarms in-flight locks. `configure-goal --registered-agent` currently replaces `coordination.registered_agents` wholesale, and an active task lease stopped protecting its todo as soon as the owner left that list. A second executor could then acquire the same todo.

## Behavior change

- An active task lease keeps protecting its todo until release or expiry even if the owner is later removed from `registered_agents`.
- New `task-lease acquire` still requires the caller to be on the roster.
- Write-scope conflicts, `task-lease inspect`, and completion fencing follow the same rule: roster membership no longer self-disarms a time-active lease.
- `configure-goal` keeps `--registered-agent` as a full replace, and adds `--add-registered-agent` / `--remove-registered-agent` for in-place roster edits.
- If a replace, remove, or clear would drop an owner who still holds an active lease, the command emits an explicit `active_lease_owner_warnings` payload (and markdown warning). It does not refuse the roster change and does not wipe the lease.

## Compatibility

- `--registered-agent` remains a wholesale replacement.
- `--clear-registered-agents` is unchanged except for the new warning when it would drop an active-lease owner.
- `--registered-agent` cannot be combined with the new incremental flags.
- Existing callers that only add agents by repeating `--registered-agent` still replace the whole list; hosts that want a true add should switch to `--add-registered-agent`.
- HTTP `configure-goal` compact responses now include `active_lease_owner_warnings`. Incremental add/remove is CLI/Python only in this PR.

## Tests

Worktree validation:

```text
python -m pytest -q tests/control_plane/test_roster_safe_leases.py
# 9 passed
python -m ruff check <touched paths>
# All checks passed
loopx canary premerge --from-git-diff
# status: passed; selected=18 failures=0
```

Core scenes in `tests/control_plane/test_roster_safe_leases.py`:

1. A holds a lease, roster is replaced with only B, B's acquire of the same todo is rejected (`todo_lease_conflict`); the lease stays active.
2. After A releases, or after the lease expires, B can acquire.
3. Incremental add of C does not change A/B roster membership or either lease record.
4. Wholesale replace or incremental remove that would drop an active-lease owner emits `active_lease_owner_warnings`.
5. New acquire still requires the owner to be registered.

Discriminant proof (restore the roster-coupled `existing_is_effective` check): the same-todo steal tests fail with `DID NOT RAISE TaskLeaseError`; restoring the time-active check returns them to green.

## Future-facing pass

Applied: lease protection now keys off time-active status rather than roster membership, and the drop-warning helper lives next to `list_active_task_leases` so `configure_goal.py` stays under the 1500-line module budget. Deferred: HTTP/dashboard incremental add/remove, and whether renew should stay roster-gated (this PR keeps renew as a new grant and still requires registration).

Do not self-merge: this is a runtime behavior change.

### DIFF
--- a/docs/reference/protocols/peer-agent-runtime-v1.md
+++ b/docs/reference/protocols/peer-agent-runtime-v1.md
@@ -29,7 +29,11 @@ rank-bearing `role` field, including null-valued placeholders.
 
 ## Work Ownership
 
-1. An explicit todo `claimed_by` or active task lease wins.
+1. An explicit todo `claimed_by` or active task lease wins. An active task
+   lease keeps protecting its todo until release or expiry even if that owner
+   is later removed from `registered_agents`. New acquires still require a
+   registered owner. Use `configure-goal --add-registered-agent` /
+   `--remove-registered-agent` to change the roster without replacing it.
 2. An unclaimed todo must be claimed or leased before delivery.
 3. An explicitly agent-scoped replan obligation stays with that agent.
 4. An unscoped replan obligation is assigned to exactly one registered peer by
--- a/examples/cli-registry-admin-command-modularization-smoke.py
+++ b/examples/cli-registry-admin-command-modularization-smoke.py
@@ -232,7 +232,13 @@ def main() -> None:
     require("handle_registry_admin_command" in init_source, "__init__ did not export registry admin handler")
 
     for command, options in {
-        "configure-goal": ("--quota-compute", "--registered-agent", "--execute"),
+        "configure-goal": (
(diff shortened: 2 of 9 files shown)


lightencc:
Withdrawing: this capability will be maintained in a long-lived fork tailored for a downstream project, not proposed upstream. Thanks!

PR #3524, opened 2026-08-23
fix(todos): rebind stale generated Next Action

huangruiteng:
## Problem

A machine-bound, low-priority startup route could remain in the durable goal-level Next Action after an agent authored a strictly higher-priority executable Todo. Quota selection already chose the higher-priority work, but status and refresh projections could still present the stale route to clients.

## Change

- add a typed Todo transition that rebinds Next Action only on a strict priority improvement;
- preserve manual or untyped operator text, equal-priority tie-breaking, multi-entry routes, and routes claimed by another agent;
- keep recommended startup checks below concrete task work while requiring real connection, permission, or runtime gates to use explicit prerequisites or blockers;
- document the goal-level and per-agent-lane precedence contract.

This changes default behavior only when an agent Todo is added while Next Action has exactly one typed binding to compatible lower-priority work.

## Validation

- `pytest -q tests/control_plane/test_todo_next_action_settlement.py tests/control_plane/test_quota_cli_projection.py tests/control_plane/test_capability_gate.py tests/control_plane/test_status_agent_lane_projection.py` (35 passed)
- `npm run typecheck:control-plane`
- `npm run test:control-plane` (110 passed)
- `python examples/project/onboarding-no-scan-projection-smoke.py`
- `python3 examples/control_plane/control-plane-maintainability-ratchet-smoke.py`
- `loopx canary premerge --from-git-diff --git-diff-base origin/main --tier standard` (17 selected checks passed; public boundary passed)

Review requested; this PR will not be self-merged.

### DIFF
--- a/docs/reference/protocols/long-horizon-agent-state-protocol-v0.md
+++ b/docs/reference/protocols/long-horizon-agent-state-protocol-v0.md
@@ -123,6 +123,15 @@ timeline. Use an independent state file only when the work has a distinct
 `goal_id`, durable objective, protected boundary, owner policy, or backlog that
 would be misleading if compressed into the parent goal's `Next Action`.
 
+The goal-level `Next Action` remains a stable tie-breaker among equally ranked
+work. When it carries a typed todo binding, adding an executable agent todo may
+rebind it only if the new todo has strictly higher priority and belongs to the
+same agent lane (or replaces an unclaimed generated route). Manual or untyped
+operator text and routes already owned by another agent remain authoritative.
+Startup checks that are merely recommended stay below concrete task work;
+required connection, permission, or runtime setup must instead be represented
+as an explicit high-priority prerequisite or blocker.
+
 Every new rollout event that should participate in concurrent-agent views should
 try to include at least one stable join key: `agent_id`, `todo_id`, `run_id`,
 `lane.lane_id`, or `case_id`. Events that cannot expose a join
--- a/loopx/control_plane/todos/next_action.ts
+++ b/loopx/control_plane/todos/next_action.ts
@@ -31,6 +31,7 @@ export interface TodoNextActionSnapshot {
   status: string;
   task_class: string | null;
   priority: TodoPriority | null;
+  claimed_by: string | null;
   text: string;
   index: number;
   completion_continuation: string | null;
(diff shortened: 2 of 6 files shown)


huangruiteng:
Maintainer self-review/refinement report

No remaining correctness or architecture blocker found on the exact final head `fb3051710`.

### Product and ownership judgment

- The PR fixes a real projection split: after a higher-priority same-lane advancement Todo is added, quota could select it while the durable goal-level `Next Action` remained bound to the lower-priority generated startup route.
- The transition rule remains single-owned by the typed TypeScript Todo Next Action boundary. Python only supplies the final Todo snapshot and applies the returned projection; it does not recreate priority or binding semantics.
- I did not expand this into migration of the full Todo-add transaction. That path also owns metadata, lease, capability, and gate semantics and would make this focused consistency fix harder to review and revert.

### Bounded future-facing refinement applied

- Reused one TypeScript `Next Action` section renderer for both add reconciliation and completion settlement, removing duplicate reconstruction logic.
- Added a Python adapter guard so unchanged additions and non-advancement Todos do not cross the TypeScript runtime boundary.
- Added a negative integration test proving those non-candidates stay outside the runtime call.

### Exact-head validation

- TypeScript: `npm run typecheck:control-plane` passed; `npm run test:control-plane` passed, 110/110.
- Python: focused Next Action/projection suite passed, 30/30; `mypy` passed; changed-path `ruff check` passed.
- Risk gate: `loopx canary premerge --from-git-diff --goal-id loopx-meta --git-diff-base origin/main --tier standard` passed 18/18 selected checks, with 0 failures and 0 manual holds; public/private boundary passed.
- Exact-scope change-quality receipt: `cqr_d96c2e33ee549d68996c`, decision `pass`.
- Performance characterization: warm typed transition p50 ~0.42 ms / p95 ~1.06 ms; guarded no-op is effectively zero. Matched full CLI runs showed no end-to-end regression; cold full-add p50 changed from ~210.6 ms to ~213.7 ms (~1.5%).

There were no product validation failures. One initial focused pytest invocation named a nonexistent test path; it was discarded and rerun against the repository's actual focused files. A model-behavior test is not applicable here because the change does not alter model-visible instructions or interpretation; typed transition tests plus Python integration tests are the authoritative behavior boundary.

With owner authorization to self-merge, I will merge after the refreshed remote checks are green.

PR #3548, opened 2026-08-24
fix(quota): settle repaired todo delivery contracts

huangruiteng:
## Summary
- re-resolve an unknown heartbeat receipt causality from the exact current Todo contract
- preserve the immutable turn identity while accepting only typed repository/write or explicit non-delivery repairs
- keep unknown or missing repairs fail-closed and add focused regression coverage

## Root cause
A turn receipt intentionally freezes the original Todo causality. When a legacy or underspecified Todo was repaired after the guard, `quota spend-slot` kept reading the immutable `unknown` snapshot and could never observe the accepted contract repair named in its own error guidance.

## Validation
- 63 focused quota settlement/workspace causality tests passed
- control-plane maintainability and peer runtime smokes passed under Python 3.11
- risk-based premerge canary passed all selected checks with no manual holds
- exact change-quality receipt: `cqr_01ee11c549d91adcae26`
- live exact-turn retry validated the independent worktree and committed one quota spend receipt

## Scope
The repair stays in the typed quota settlement boundary. It does not rewrite heartbeat receipts, weaken workspace validation, or change scheduling, scoring, permissions, or production behavior outside settlement re-entry.

### DIFF
--- a/loopx/control_plane/quota/settlement_workspace_causality.py
+++ b/loopx/control_plane/quota/settlement_workspace_causality.py
@@ -292,8 +292,9 @@ def completed_todo_workspace_causality(
     *,
     goal_id: str,
     todo_id: str,
+    source: str = "completed_todo_contract_fallback",
 ) -> dict[str, str] | None:
-    """Recover a legacy effect's causality from its exact completed Todo."""
+    """Recover an effect's causality from its exact projected Todo contract."""
 
     queue = status_payload.get("attention_queue")
     queue_items = queue.get("items") if isinstance(queue, Mapping) else []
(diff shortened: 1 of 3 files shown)


huangruiteng:
Self-review: the typed quota settlement boundary remains the single authority. The change accepts a repaired current Todo contract only when the frozen receipt was missing/unknown, and only when the repaired typed contract resolves to required or not_required; unresolved repairs remain fail-closed. The existing delivery-workspace snapshot guard still validates the exact repository/worktree before spend. Focused pytest (63), Ruff, compile, maintainability, peer runtime, and risk-based premerge canaries passed. Future-facing pass applied: the repair lookup was extracted from the already-hot slot decision function, avoiding further growth without adding a compatibility seam. No scoring, scheduler, permission, public/private evidence, or default lane behavior changes.

PR #3583, opened 2026-08-24
fix(quota): bind SSH guided start to one turn

### LINKED ISSUE #3569: bug(control-plane): guided start emits an unbindable Todo selection command
## Summary

The direct guided-start packet can emit a Todo selection command that cannot be executed. When more than one runnable Todo exists, the first quota guard correctly asks the caller to choose one, but the suggested follow-up carries `--todo-id` without a stable `--turn-instance-id`. The CLI then rejects that follow-up, so a product-native guided start cannot bind the selected Todo through the formal path.

## Impact

- A fresh guided agent can enter `selection_required` and then fail before useful work begins.
- Hosts must invent an out-of-band turn id or bypass the generated command, weakening reproducibility.
- The explicit Todo-selection behavior introduced for quota admission is not reachable from the direct guided-start packet when several Todos are runnable.

## Minimal synthetic reproduction

1. Register a goal with at least two runnable Todos for one agent.
2. Generate the direct guided-start or onboarding packet.
3. Run its emitted quota guard.
4. Observe a `selection_required` result whose suggested selection command includes `--todo-id` but no stable `--turn-instance-id`.
5. Execute the suggested command.

Actual result:

```text
loopx quota should-run --todo-id requires --turn-instance-id
```

## Expected behavior

The direct guided-start packet should establish and propagate one stable turn identity across its initial quota guard and any emitted explicit-selection follow-up. The fix must preserve fail-closed Todo selection and must not weaken quota, lineage, or settlement gates.

## Reproduction revision

Reproduced from a clean install of `main` at `d2c959f7a1a1fe753242bc2e281fce6d69372ef9`.

## Suspected surface

This appears to be a command-construction and host-packet binding gap: heartbeat-owned quota guards receive a turn id, while the direct guided-start/bootstrap path renders the guard without one. The exact fix location should be chosen after tracing the generated packet and selection action end to end.

## Acceptance criteria

- A focused synthetic regression creates multiple runnable Todos and executes the generated selection action successfully.
- The same stable turn identity is used for initial admission, explicit Todo binding, subsequent writeback, and settlement.
- Single-Todo guided start remains unchanged.
- Missing or stale turn identity still fails closed.
- Public-boundary validation contains no private state or provider-specific context.

huangruiteng:
## Summary

- let direct `codex-app-ssh` guided starts mint a stable Turn identity at quota admission
- propagate that identity through the generated multi-Todo selection command
- honor the selected Todo for both supported Codex App guided runtime profiles
- keep ordinary visible-Goal polling unchanged and keep generic profiles fail-closed

Closes #3569.

## Root cause

The guided-start generator and quota validator shared a heartbeat-only allow rule, while `codex-app-ssh` uses the distinct `codex_app_ssh_goal` runtime profile. Even after admitting that profile, explicit Todo selection was still classified as heartbeat-only, so the receipt identity and projected selected Todo could diverge.

## Validation

- 182 adjacent quota, start-goal, and host activation tests passed
- focused post-rebase regression: 5 passed
- onboarding host-loop activation smoke passed
- Ruff passed on every changed Python file
- premerge canary passed 17/17 selected checks with no failures, warnings, skips, or manual holds
- mypy 2.3.0 hit its own `INTERNAL ERROR` before emitting any type diagnostic; recorded as an infrastructure failure, not a passing check

## Public/private boundary

The diff contains only product runtime code and synthetic public regression fixtures. Added lines were scanned for credentials, private paths, internal links, raw benchmark evidence, and provider-specific context; none were present.

## Merge decision

The change is a narrow runtime repair with executable end-to-end regression coverage. It does not change benchmark scoring, task semantics, permissions, or ordinary non-guided quota calls.

### DIFF
--- a/examples/control_plane/agent-onboard-host-loop-activation-smoke.py
+++ b/examples/control_plane/agent-onboard-host-loop-activation-smoke.py
@@ -429,6 +429,7 @@ def main() -> int:
         assert "--runtime-profile codex_app_ssh_goal" in (
             app_ssh_prompt["quota_guard_command"]
         ), app_ssh_prompt
+        assert "--begin-turn" not in app_ssh_prompt["quota_guard_command"], app_ssh_prompt
         assert "--turn-instance-id" not in app_ssh_prompt["quota_guard_command"], app_ssh_prompt
         assert "--source visible-goal" in app_ssh_prompt["quota_spend_command"], app_ssh_prompt
         spend_args = build_parser().parse_args(
--- a/loopx/bootstrap_command_pack.py
+++ b/loopx/bootstrap_command_pack.py
@@ -18,6 +18,9 @@ from .control_plane.goals.start_contract import (
     build_goal_start_contract,
     build_goal_start_prompt,
 )
+from .control_plane.scheduler.execution_context import (
+    GUIDED_START_TURN_RUNTIME_PROFILES,
+)
 from .host_loop_activation import (
     agent_type_for_host_surface,
     build_host_loop_activation_packet,
@@ -808,7 +811,7 @@ def build_loopx_bootstrap_command_pack(
         explicit_goal_start
         and selected_agent_id
         and scheduler_command_binding.get("runtime_profile")
-        == "codex_app_heartbeat"
+        in {profile.value for profile in GUIDED_START_TURN_RUNTIME_PROFILES}
     )
     quota_guard_command = (
         render_quota_guard_command(
--- a/loopx/cli_commands/quota.py
+++ b/loopx/cli_commands/quota.py
@@ -5,16 +5,17 @@ from collections.abc import Callable, Mapping
 from dataclasses import dataclass
 from pathlib import Path
 
-from ..capabilities.repository_change_window import (
-    repository_delivery_interaction_hook,
-)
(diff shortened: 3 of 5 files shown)


huangruiteng:
Self-review / pre-merge validation update:

- Changed surfaces: guided start command generation, quota runtime-profile validation, explicit Todo selection binding, and the visible-Goal onboarding boundary.
- Direct checks: final rebased diff/compile checks passed; focused guided-start regression passed 5/5; adjacent quota/start-goal/host suite passed 182/182; onboarding activation smoke passed; Ruff passed on changed product/test files.
- Catalog canaries: 9/9 passed.
- Risk-profile smokes: 8/8 passed.
- Public/private boundary: 1/1 passed; added-line credential/private-context scan was clean.
- Failures/skips: no product test failure or canary skip. Standalone mypy 2.3.0 terminated with its own INTERNAL ERROR before producing a type diagnostic.
- Manual holds: none.
- Self-review refinement: added a negative regression proving ordinary SSH visible-Goal polling does not gain `--begin-turn`; only direct guided start opts in.
- Merge decision: coverage is sufficient for this narrow control-plane repair; self-merge after the final remote checks pass.

PR #3611, opened 2026-08-25
fix(quota): restore CLI composition seams

Alicecooo:
## Summary
- restore the quota CLI composition boundary after the quota command-module split
- keep Lark urgency provider ownership in `loopx.cli_commands.quota`
- inject the status collector through the CLI boundary so scheduler and diagnostic paths use the composed collector
- normalize missing or invalid quota limits before scheduler lookback calculation
- preserve the existing Python CLI contract and the TypeScript control-plane authority boundary
## Issue Or Task
-  N/A
## Root cause
The quota command was split into `quota.py` and `quota_context.py`, but the extracted context module retained ownership of the Lark urgency provider. This left the CLI composition boundary inconsistent with the architecture contract.
The extracted context builder also called its own imported `collect_status`, so replacing the collector at the CLI command boundary did not affect scheduler acknowledgement and quota diagnostic paths. In addition, scheduler commands could receive a missing or invalid `limit` value before applying the periodic lookback floor.
## Safety and compatibility
- Lark urgency remains composed by the quota CLI and is injected into the context builder.
- `quota_context.py` no longer owns the Lark provider dependency.
- Existing quota commands, Python CLI arguments, and output payload shapes remain unchanged.
- Scheduler commands continue to enforce the periodic lookback minimum.
- Invalid or missing limits fail closed to a bounded non-negative value.
- No new capability, extension, provider, or product-specific routing authority is introduced.
- No TypeScript control-plane authority is duplicated in Python.
## Validation
- [x] related pytest suites
- [x] `loopx check --scan-root .`: 0 errors
- [x] TypeScript/control-plane premerge canary
- [x] Python compilation for the changed quota modules
- [x] CLI output boundary and import-boundary checks

## Future-facing pass
The change keeps provider composition at the CLI boundary and makes the extracted quota context consume injected capabilities instead of importing provider implementations itself. This removes duplicate ownership and leaves the next control-plane migration step localized to the existing typed runtime boundary.
No product-specific vocabulary, provider implementation, or second business-rule authority was added to the generic quota context or ordinary Goal path. 翻译

### DIFF
--- a/loopx/cli_commands/quota.py
+++ b/loopx/cli_commands/quota.py
@@ -72,6 +72,7 @@ from ..quota import (
 from ..status import collect_status
 from ..upgrade import resolve_codex_app_automation_rrule
 from .quota_context import QuotaCommandContext, prepare_quota_command_context
+from .lark_inbox import build_lark_operator_inbox_urgency_projector
 from .quota_host_poll import attach_host_poll_receipt
 from .quota_monitor_poll import record_quota_monitor_poll_for_cli
 from .quota_registration import (
@@ -416,6 +417,10 @@ def handle_quota_command(
             args,
             registry_path=registry_path,
             runtime_root_arg=runtime_root_arg,
+            status_collector=collect_status,
+            operator_inbox_urgency_projector=build_lark_operator_inbox_urgency_projector(
+                runtime_root_arg=runtime_root_arg,
+            ),
         )
         heartbeat_turn_id = context.heartbeat_turn_id
         detail_sections = context.detail_sections
--- a/loopx/cli_commands/quota.py
+++ b/loopx/cli_commands/quota.py
@@ -418,8 +418,8 @@ def handle_quota_command(
             registry_path=registry_path,
             runtime_root_arg=runtime_root_arg,
             status_collector=collect_status,
-            operator_inbox_urgency_projector=build_lark_operator_inbox_urgency_projector(
-                runtime_root_arg=runtime_root_arg,
+            operator_inbox_urgency_projector_factory=(
+                build_lark_operator_inbox_urgency_projector
             ),
         )
         heartbeat_turn_id = context.heartbeat_turn_id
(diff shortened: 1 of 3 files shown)


huangruiteng (approved):
Findings: none after refinement.

The original patch correctly restores the quota CLI as the Lark provider-composition owner and injects the CLI-visible status collector into the extracted context. I added one bounded refinement: the CLI now injects a projector factory, and `quota_context` instantiates it only after resolving the registry-aware runtime root. This preserves custom `common_runtime_root` behavior without moving Lark implementation authority back into the generic context.

- Motivation: restore the composition seam lost in the quota command split; this is also the root cause of the current main/stacked-PR fast-test failures around import ownership, scheduler lookback, CLI diagnostics, and the Codex CLI transaction smoke.
- Solved: yes. Provider ownership stays at the CLI boundary, the collector is injectable where tests and host composition expect it, missing limits remain bounded, and registry-derived extension state is preserved.
- User impact: scheduler ACK and quota error paths again observe the composed collector; custom runtime-root installations no longer risk reading the wrong Lark activation state.
- Main risk: quota is a hot control-plane path, but the change is a narrow dependency-injection repair with no output-schema, selection, permission, or TypeScript-authority change.
- Future-facing pass: applied at the existing CLI/context boundary; no broader abstraction is warranted.

Validation:

- 116 related tests passed across import boundaries, scheduler ACK, CLI diagnostics, Codex CLI E2E, and Lark urgency composition.
- Python compile and Ruff passed on the changed files.
- The six failures reproduced by the current stacked PRs all pass on this head, including the three-turn Codex CLI transaction E2E.
- Premerge selected 17 checks: 16 passed in one run; the hot-path interface smoke failed once transiently and passed on immediate direct rerun. The CLI differential initially could not create its temporary detached worktree because of the repository's local reference hook; rerunning with hooks disabled only for that validation process passed.
- Direct broad mypy remains non-actionable because the repository baseline reports pre-existing errors outside this diff; no new targeted diagnostic points to the refined factory boundary.
- Public/private scan is clean; only an existing synthetic fixture reset token was matched.

Merge decision: approved; merge after the refreshed required checks complete.

Issue #3560, opened 2026-08-24
Dashboard Personal Workspace hides completed Todo progress and shows stale next action


bubble66:
## Symptom and impact

The Personal Workspace Goal view can show an active current Todo while reporting `已完成 0`, even though the status payload reports a non-zero completed Todo count. The Goal header can also remain on an older `recommended_action` after `agent_management_projection.current_todo` has advanced.

This makes a healthy long-running Goal look stalled and prevents operators from seeing cumulative execution progress.

## Expected behavior

- The Goal header should prefer the current projected Todo when one exists.
- The completed column should display the summary's completed count.
- If completed Todo details are available in a fallback projection, the UI should merge or read them instead of treating a compact open-only item list as the complete history.

## Actual behavior

- The header prefers an older queue or latest-run `recommended_action`.
- `project_asset.agent_todos` can correctly report `done > 0` while its compact `items` contains only open Todos.
- Personal Workspace maps only those compact `items`, so `doneAgentTodos.length` becomes zero and the page renders `已完成 0`.

## Minimal synthetic reproduction

Serve a valid status payload containing the following public-safe shape:

```json
{
  "attention_queue": {
    "items": [{
      "goal_id": "example-goal",
      "recommended_action": "Older Todo",
      "agent_todos": {
        "total_count": 4,
        "open_count": 1,
        "done_count": 3,
        "items": [
          {"todo_id": "open-1", "done": false, "title": "Current Todo"},
          {"todo_id": "done-1", "done": true, "title": "Completed A"},
          {"todo_id": "done-2", "done": true, "title": "Completed B"},
          {"todo_id": "done-3", "done": true, "title": "Completed C"}
        ]
      },
      "project_asset": {
        "agent_todos": {
          "total": 4,
          "open": 1,
          "done": 3,
          "items": [
            {"todo_id": "open-1", "done": false, "title": "Current Todo"}
          ]
        }
      }
    }]
  },
  "agent_management_projection": {
    "agents": [{
      "agent_id": "example-agent",
      "current_todo": {
        "goal_id": "example-goal",
        "todo_id": "open-1",
        "title": "Current Todo"
      }
    }]
  }
}
```

Open the Personal Workspace Goal Tasks view. The current implementation can show the older header text and `已完成 0`, despite `done=3` and the fallback completed items.

## Suspected product surface

This is a hypothesis based on source inspection:

[...]

PR #3689, opened 2026-08-26
fix(dashboard): project completed Todos into the Personal Workspace board

now-ing:
## Summary

- `project_asset_todo_summary` now forwards the bounded `recent_completed_advancement_items` lane alongside the existing `done` count, so the presentation surface can see finished work without growing the payload (lane stays capped).
- The dashboard status schema and Personal Workspace Goal projection surface `doneTodoCount` plus the recent completed items, merge them into the Goal board, and the completed column counts payload completions instead of deriving a false zero from the open-only item list.
- The Goal header now prefers the current projected Todo text over stale `queueItem`/`latestRun` `recommended_action` strings.
- The contract smoke's `statusRequestActive` assertion predates the example-source guard on current main and fails without this change; realigned it (separate commit).

## Issue Or Task

- Fixes #3560

## Validation

- [x] `python3 -m py_compile loopx/control_plane/work_items/project_asset.py`
- [x] `python3 -m pytest tests/control_plane/ -k todo -x -q` (393 passed)
- [x] `python3 -m pytest tests/control_plane/test_goal_vision_succession.py` (34 passed)
- [x] `python3 examples/control_plane/todo-first-open-summary-smoke.py`
- [x] `node apps/presentation/dashboard/src/features/personal-workspace/personal-workspace-contract.test.mjs`
- [x] `tsc --noEmit` in apps/presentation/dashboard

## Type of Change

- [x] Bug fix

## LoopX Area

- [x] Public docs or presentation surface (dashboard)

## Technical Direction

- [x] Core control-plane hardening

## Boundary Checklist

- [x] No `.loopx/`, credentials, private traces, or local machine paths committed
- [x] Change scoped to the linked issue
- [x] Every commit includes a DCO `Signed-off-by` trailer

PR #3695, opened 2026-08-26
test(ci): run the Personal Workspace contract smoke on Python Tests

now-ing:
## Summary

- The Personal Workspace contract smoke (`personal-workspace-contract.test.mjs`) only ran locally: when its `statusRequestActive` assertion drifted behind the example-source guard in `dashboard-page.tsx`, nothing on CI caught it.
- This aligns the stale assertion with the shipped condition and runs the smoke as a step of the Python Tests workflow (node 22.6 is already set up there), so drawer/board contract drift fails fast.

## Issue Or Task

- No public issue yet — this came out of a brainstorm audit while fixing #3560-adjacent dashboard work; happy to file one if maintainers want it tracked.

## Validation

- [x] `node apps/presentation/dashboard/src/features/personal-workspace/personal-workspace-contract.test.mjs` passes on this branch (it fails on current `main` before the assertion fix)
- [x] workflow YAML parses (`python -c "import yaml; yaml.safe_load(...)"`)

## Type of Change

- [x] Test update

## LoopX Area

- [x] Build, packaging, installer, or CI

## Technical Direction

- [x] Core control-plane hardening

## Boundary Checklist

- [x] No `.loopx/`, credentials, private traces, or local machine paths committed
- [x] Change scoped to the contract smoke + CI wiring
- [x] DCO `Signed-off-by` trailer

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -65,6 +65,10 @@ jobs:
           npm run typecheck:control-plane
           npm run test:control-plane
 
+      - name: Qualify the Personal Workspace contract smoke
+        run: >-
+          node apps/presentation/dashboard/src/features/personal-workspace/personal-workspace-contract.test.mjs
+
       - name: Lint test suite
         run: >-
           python -m ruff check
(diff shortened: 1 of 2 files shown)


now-ing:
**Note on the missing CI runs:** this PR touches `.github/workflows/python-tests.yml`, so GitHub requires maintainer approval before workflows run on this fork PR. The new step is exercised by the workflow change itself; locally validated as:

```
node apps/presentation/dashboard/src/features/personal-workspace/personal-workspace-contract.test.mjs
# personal workspace drawer contract smoke passed
```

(and the smoke fails on current `main` before the assertion fix in this PR). Happy to split the assertion fix from the workflow wiring if maintainers prefer reviewing them separately.

now-ing:
Understood — thanks for taking a look. If the CI wiring is something you'd rather own or approach differently, happy to drop this; the assertion drift fix itself (aligning `statusRequestActive` with the example-source guard) is also carried in #3689 if that's a more convenient landing spot.

PR #3708, opened 2026-08-27
fix(lark): acknowledge inbox reads at turn start

huangruiteng:
## Summary

This PR makes the Lark Inbox `Get` reaction an Agent read acknowledgement with one precise lifecycle boundary:

- the realtime collector only validates and durably captures events; it never creates `Get`;
- the Agent's turn-start Lark Inbox hook creates `Get` after it reads and confirms a still-pending human message;
- mention, verified reply, question, and material-review classification affect attention and response policy, not whether a read message is acknowledged;
- every reply-enabled Inbox defaults to `Get`; an explicit empty `received_reaction_emoji` disables the provider write;
- reaction failure cannot discard the Inbox event or falsely report successful reading.

The PR also keeps the exact typed-profile self-message filter: a verified Bot-authored message is excluded, while identity lookup failure remains fail-open and captures the message.

## Final semantic contract

`Get` means:

> This pending human message was read into the Agent's turn-start processing chain.

It does **not** mean:

- the background collector stored the event;
- the Bot was mentioned;
- a reply is required or has been sent;
- domain processing is complete;
- the message has been settled.

The hook's public `observation_count` counts messages newly read by that turn. This includes a message already persisted by the realtime collector when the hook performs its first read-ack attempt, without double-counting a message that was both newly persisted and acknowledged in the same hook invocation.

## Recommended review order

1. `d478992ec` — introduces the receipt-backed reaction and exact typed self-message boundary.
2. `db47e91c2` — documents the original provider-neutral authority boundary.
3. `daabf87ca` — covers cross-ingress identity, failure, and idempotency behavior.
4. `67f30eb88` — applies the final semantics: collector-only capture, default `Get`, attention-independent turn-start read ACK, correct observation projection, and focused regressions.
5. `3263d1593` — updates the architecture and Lark operator contract to the final semantics.

Reviewers should treat commits 4–5 as the semantic correction to the earlier shared collector/history implementation.

## Ownership and authority

- **Core control plane** owns the provider-neutral turn-start hook contract, validates `external_writes_performed`, and admits only the narrow `provider_message_reaction` write scope.
- **Lark turn-start sync** owns bounded provider-history reading, Inbox/cursor readback, pending-message confirmation, reaction creation, and private reaction receipts.
- **Realtime collector** owns provider event consumption, routing, exact typed self-message filtering, and durable Inbox capture. It no longer consumes reaction authority or requires reply permission.
[...]

PR #3713, opened 2026-08-27
fix(lark): make turn-start read acknowledgements durable

huangruiteng:
## Summary

Follow up on the two P1 findings raised after #3708 was merged:

- decouple the first Agent-owned turn-start read receipt from optional provider reactions, so collector-captured messages still require Agent reading when `received_reaction_emoji` is explicitly disabled;
- retry failed reactions from a durable owner-private pending-read set rather than relying on the bounded provider overlap window;
- fence reaction creation with prepared/created private operation state, recovering a known reaction id after receipt persistence failure and failing closed when the provider outcome is uncertain instead of creating a duplicate;
- update the provider-neutral RFC and Lark inbox operator docs.

Addresses review findings:
- https://github.com/huangruiteng/loopx/pull/3708#pullrequestreview-5043683929
- https://github.com/huangruiteng/loopx/pull/3708#pullrequestreview-5043714094

## Validation

- `pytest -q tests/extensions`: 530 passed
- focused mypy on both changed runtime modules: passed
- Ruff check and format check on changed Python files: passed
- `loopx check` public/private boundary scan: 6 files clean
- `loopx canary premerge --from-git-diff --goal-id loopx-meta`: passed, 0 failures/skips/manual holds
- change-quality receipt: `cqr_490ba94256652bceeb7f`, exact-scope valid
- `git diff --check`: passed

## Scope and risk

This remains inside the existing Lark extension owner. It adds no outbound-message authority, no new provider capability, and no generic queue. The realtime collector still only captures; the Agent turn-start hook remains the sole reaction-effect owner. No private state, credentials, local paths, raw logs, or generated evidence are included.

### DIFF
--- a/docs/architecture/rfcs/provider-neutral-turn-start-inbox-hook-v0.md
+++ b/docs/architecture/rfcs/provider-neutral-turn-start-inbox-hook-v0.md
@@ -45,8 +45,9 @@ to a durable effect receipt before inbox ACK.
 
 ```text
 provider-neutral turn_start dispatch
-  -> provider read + owner-private commit/readback
-  -> optional pending-message read acknowledgement + private reaction receipt
+  -> provider read + owner-private inbox commit/readback
+  -> first Agent-read receipt independent of optional provider reaction
+  -> retry optional reaction from durable pending reads
   -> fresh status + quota projection
   -> inbox lane preempts ordinary work when agent_read_required
   -> private drain into the active Agent turn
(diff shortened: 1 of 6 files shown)


huangruiteng:
Self-review completed against exact head `f223f4f78352047f08343ba78ecd63af0f6e54f9`.

## Review result

No remaining blockers found. The change is positive and appropriately scoped: it closes both P1 correctness gaps from #3708 without broadening the Lark provider authority boundary.

### Changed surfaces

- owner-private Lark turn-start first-read receipts and pending retry projection;
- received-reaction prepared/created effect journal and replay behavior;
- bounded history sync observation accounting;
- focused Lark extension tests and the canonical RFC/operator docs.

### Review lenses

- **Typed state rules:** private schemas explicitly bind message ids and allow only `prepared`/`created` effect phases; malformed state fails closed.
- **Domain neutrality:** core hook receipt semantics remain provider-neutral; Lark-specific reaction mechanics stay under `loopx/extensions/lark`.
- **Behavior disclosure:** docs now state that reaction disablement does not cancel Agent reading, retries do not depend on overlap, and uncertain provider effects block duplicate create.
- **Guidance vs obligation:** `agent_read_required` is derived only from the first owner-private Agent-read receipt, not optional provider-write attempts.
- **Authority/privacy:** no outbound reply or new capability authority is added; all ids and effect state remain owner-private and public results stay content-free.

### Validation

- `pytest -q tests/extensions`: **530 passed**
- focused mypy on both changed runtime modules: **passed**
- Ruff lint/format and Python compile: **passed**
- public/private boundary scan: **6 changed files clean**
- risk-based premerge: **4 direct checks + 10 catalog/risk/boundary checks passed**
- failures/skips/manual holds: **0 / 0 / 0**
- exact-scope quality receipt: `cqr_490ba94256652bceeb7f` (**valid**)

The coverage is sufficient because it tests both original counterexamples directly: collector-first + disabled reaction, and a failed reaction whose message is outside the next real provider window. It also covers provider-create success followed by receipt failure, as well as cleanup/operation uncertainty without duplicate create.

### Future-facing pass

Applied narrowly: the durable first-read set also owns retry discovery and prunes settled ids, so the fix does not add a generic outbound queue or a second cursor authority. A broader reaction reconciliation framework was intentionally not introduced.

Self-merge is appropriate once required GitHub checks are green; the local premerge gate returned `self_merge_allowed=true` with no manual holds.

PR #3732, opened 2026-08-29
fix(dashboard): keep the default status source same-origin

huangruiteng:
## Summary

- keep the built-in Dashboard status source on the current origin as `/status.json`
- preserve absolute loopback URLs for explicitly configured SSH tunnel sources
- make the packaged chat entry point provide the same workspace i18n context as the main entry point
- add request-level browser coverage proving a bare Dashboard never bypasses the Vite proxy for browser-local port `8766`

## Why

VS Code Remote SSH users normally forward only the Vite port (`5173`). The previous built-in source used `http://127.0.0.1:8766/status.json`, so the browser resolved that loopback address on the user's local machine and bypassed Vite's remote `/status.json` proxy. Restoring the same-origin default lets local development, Remote SSH, and the packaged app use their owning same-origin status route, while explicit SSH tunnel sources remain absolute and read-only.

Rebuilding the packaged chat surface also exposed that `chat-main.tsx` lacked the `WorkspaceI18nProvider` already used by the main entry point. Without that parity fix, the packaged surface crashed before issuing any status request, so the focused packaged regression could not validate the shipped bundle.

## Validation

- `npm run smoke:status-sources`
- `npm run smoke:pwa-bundle`
- `npm run build`
- `npm run smoke:status-source-switch-browser`
- `npm run smoke:status-source-switch-packaged`
- `git diff --check`
- `loopx canary premerge --from-git-diff`: 9/10 selected checks passed; the unrelated `todo-readmodel-boundary-smoke.py` failure was reproduced unchanged on a clean archive of `origin/main` (`task_lease_acquire_adapter.py` already imports from `loopx.status`)

## Scope and risk

Changed surfaces are limited to the typed status-source catalog, its focused smoke, the browser switch regression, packaged-entrypoint provider parity, and regenerated chat assets. This does not change permissions, status schemas, control-plane state transitions, or SSH tunnel registration semantics.

The future-facing scope pass found no additional abstraction worth adding: the existing catalog and Vite proxy are the correct owners, so this change restores their contract instead of introducing another routing layer.


huangruiteng:
Self-review for exact head `f17212cb51130ecbcf10a3d1fbbdf6d435b539d5`:

- **Positive value / right-sized scope:** fixes the confirmed Remote SSH regression with one owning default (`/status.json`) instead of adding another proxy or environment detector. The browser now resolves the status route through the same origin that served the Dashboard.
- **Local + SSH coverage:** local Vite, Remote SSH with only `5173` forwarded, and packaged `/chat/` all use the owning same-origin route. Only explicitly registered SSH tunnel sources retain absolute loopback URLs.
- **Packaged-entrypoint parity:** rebuilding current source revealed that `chat-main.tsx` omitted the existing `WorkspaceI18nProvider`, causing the packaged app to crash before any status request. The PR adds only the provider already used by `main.tsx`.
- **Future-facing pass:** no new abstraction is warranted. The existing typed source catalog and Vite/package status routes are the correct boundaries; additional routing machinery would be speculative.

### Review lenses

- Typed state rules: source kinds and read-only authority remain typed; relative sources map to the existing `local` source, while `normalizedTunnelUrl` continues to reject relative tunnel URLs.
- Domain neutrality: no product-specific wording enters generic control-plane contracts.
- Behavior disclosure: the default changes from browser-local absolute loopback to same-origin, and the affected local/Remote SSH/packaged lanes are named in the PR body.
- Guidance vs obligation: no machine-enforced work or lifecycle obligation changes.
- Public/private boundary: passed; generated assets contain no local paths, credentials, or private operating evidence.

### Validation

- PASS `npm run smoke:status-sources`
- PASS `npm run smoke:pwa-bundle`
- PASS `npm run build`
- PASS `npm run smoke:status-source-switch-browser`
- PASS `npm run smoke:status-source-switch-packaged`
- PASS `git diff --check`
- PASS public/private boundary scan
- PARTIAL standard premerge canary: 9/10 selected checks passed. The sole failure, `todo-readmodel-boundary-smoke.py`, is unrelated to this PR and reproduces on a clean `origin/main` archive because existing `task_lease_acquire_adapter.py` imports from `loopx.status`.

The request-level regression starts from the bare Dashboard URL, requires a same-origin `/status.json` request, and rejects any request to browser-local `http://127.0.0.1:8766/status.json`. This coverage is sufficient for the changed source-selection and packaged surfaces.

PR #3745, opened 2026-08-29
fix(dashboard): retain startup error context

huangruiteng:
## Summary

- detect when a dashboard status or Chat child exits before readiness
- keep the child's original stderr on the inherited terminal stream, then add service name and exit-status context
- fail immediately instead of waiting for the 15-second HTTP readiness timeout
- remove the duplicate Lark CLI argument initialization left adjacent to #3738

PR #3738 fixed the immediate unbound-variable crash, but it also exposed a supervision gap: an early child failure printed its raw error without a durable connection to the launcher's later generic timeout. This follow-up keeps lifecycle authority in the existing `wait_for_service` boundary and reuses inherited stderr rather than adding log storage or a second error transport. The operator now sees the original failure plus the service identity and exit status together. No API, permission, persistence, or UI behavior changes.

The adjacent future-facing boundary was reviewed. Passing the existing child PID into the current readiness owner is sufficient; a new process supervisor or logging abstraction would be disproportionate here.

## Validation

- `uv run --project . --extra test pytest tests/test_dashboard_command.py -q` — 26 passed
- `uv run --project . --extra test ruff check tests/test_dashboard_command.py` — passed
- `bash -n scripts/dashboard-dev.sh` — passed
- `git diff --check` — passed
- `loopx canary premerge --from-git-diff --goal-id loopx-meta` — passed; 3 selected canaries, no failures or manual holds; self-merge allowed
- change-quality receipt `cqr_f5ee6b39882fb8035694` — valid for the exact committed diff
- ShellCheck was unavailable in the current environment

## Boundary review

The two changed public paths contain no credentials, private state, local absolute paths, raw logs, or internal links. Original service output remains transient on the user's existing terminal stream and is not copied or persisted.

### DIFF
--- a/scripts/dashboard-dev.sh
+++ b/scripts/dashboard-dev.sh
@@ -62,16 +62,25 @@ fi
 wait_for_service() {
   local service_name="$1"
   local service_url="$2"
-  "${PYTHON_BIN}" - "${service_name}" "${service_url}" <<'PY'
+  local service_pid="$3"
+  local probe_status=0
+  local service_status=0
+  "${PYTHON_BIN}" - "${service_name}" "${service_url}" "${service_pid}" <<'PY'
+import os
 import sys
 import time
 import urllib.error
 import urllib.request
 
-name, url = sys.argv[1:]
+name, url, pid_text = sys.argv[1:]
+pid = int(pid_text)
 deadline = time.monotonic() + 15
 last_error = "service did not respond"
 while time.monotonic() < deadline:
+    try:
+        os.kill(pid, 0)
+    except ProcessLookupError:
+        raise SystemExit(2)
     try:
         with urllib.request.urlopen(url, timeout=1) as response:
             if response.status == 200:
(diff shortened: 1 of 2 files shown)


huangruiteng:
No actionable findings on exact head `69a0482bec8a79a067d0af628657419f99f6aebd`.

Assumption checked: the launcher already intentionally exposes each child service's stdout/stderr on the invoking terminal. This change preserves that stream and only adds causal context after the child exits; it does not copy raw output into a public API, file, fixture, or durable state.

Product/architecture judgment: the patch closes the supervision gap exposed by #3738. It makes an early status or Chat failure immediate and actionable by naming the service and exit status while retaining the original error above. Extending the existing readiness owner with the child PID is the smallest reusable design; no additional supervisor or logging layer is warranted. Capability risk is low and limited to dashboard development startup failure handling. Existing successful startup and timeout behavior remain unchanged.

Validation reviewed:

- dashboard command tests: 26 passed, including distinctive raw stderr, exit status 42, and bounded latency
- focused Ruff check, Bash syntax check, and diff hygiene: passed
- LoopX premerge gate: passed with 3 selected canaries, no failures or manual holds, and self-merge allowed
- exact-scope quality receipt: `cqr_f5ee6b39882fb8035694`, valid
- public/private scan: clean; no logs or secrets are persisted
- ShellCheck: unavailable locally and disclosed in the PR body

Future-facing pass: applied only as a bounded cleanup of duplicated Lark argument initialization; broader supervision abstraction was considered and found unnecessary.

Merge decision: approved for self-merge after required GitHub checks complete successfully.

PR #3749, opened 2026-08-30
fix: consume pending periodic report intents

huangruiteng:
## Summary

- project durable, unconsumed capability intents into quota through a TypeScript-validated interaction slot
- let the periodic-report capability consume its own exact Goal/Agent sidecars into checked local HTML and Markdown artifacts
- create one generation-digest-bound user approval Todo while keeping external publication and delivery unauthorized
- recover idempotently from replay and from a crash between approval-Todo persistence and consumption-receipt persistence

## Root cause

The post-writeback producer persisted a valid periodic-report trigger intent, but quota never read the post-writeback sidecar journal. Terminal or quiet lifecycle state therefore hid mandatory local capability work indefinitely.

## Architecture

TypeScript owns admission of the generic pending-capability-intent projection and the canonical interaction contract. The periodic-report capability owns sidecar validation, report generation, content checks, receipts, and its exact approval boundary. No new Stage entity or premarked stage Todo is introduced.

## Validation

- 186 focused Python tests passed
- 288 TypeScript control-plane tests passed
- TypeScript typecheck passed
- Ruff passed
- periodic-report HTML smoke passed
- premerge canary: 18/18 checks passed, zero failures, zero manual holds
- source smoke on a real pending intent projected `governed_capability_intent`, `must_attempt=true`, and the exact local consumer command

## Authority boundary

The consumer writes only local report artifacts, a local receipt, and one exact-payload approval Todo. It does not publish to Miaoda, send a group message, or perform any external write.

## Excluded

Generated dependency lock output and local Goal/runtime state are not included.


huangruiteng:
No blocking findings after final diff review.

- Motivation: a durable post-writeback report intent was invisible to quota, so terminal/quiet lifecycle state permanently suppressed required local generation.
- Solved: the real pending-intent source smoke now selects `governed_capability_intent`, requires an attempt, and projects the exact consumer command. The consumer freezes checked local HTML/Markdown and creates one digest-bound approval gate.
- User impact: meaningful stage completion can now reach a reviewable local report even when normal Goal work has ended, without a premarked Todo or a second Stage state machine.
- Main risk: quota precedence and replay safety. The implementation preserves existing user notices, keeps canonical interaction-contract construction in the existing owner, validates exact supported commands in TypeScript, and recovers when a gate write precedes receipt persistence.
- Design: the generic seam is a TypeScript-authoritative pending-intent projection slot; periodic-report-specific journal reads and effects remain capability-owned. The future-facing pass removed the initial hand-built duplicate interaction contract; no broader refactor is needed for this PR.

Validation:

- 186 focused Python tests passed.
- 288 TypeScript control-plane tests passed.
- TypeScript typecheck, Ruff, Python compile, diff hygiene, and periodic-report HTML smoke passed.
- Premerge canary passed 18/18 selected checks with zero failures and zero manual holds.
- No Miaoda publication, group delivery, provider call, or other external write was performed.

Merge decision: self-merge is authorized by the owner and admitted by the premerge gate. I will merge after GitHub checks are green.

PR #3751, opened 2026-08-30
Fix periodic report retry after rejection

huangruiteng:
## Summary

- reopen a periodic-report intent after its exact approval gate is durably rejected
- freeze every revision in a separate retry directory instead of overwriting an earlier draft
- support repeated reject → revise cycles while suppressing replay and cross-agent matches

## Motivation and architecture

The existing consumer treated `approval_pending` as permanently terminal for an intent. Once an owner rejected the exact frozen payload, the approval Todo and the receipt diverged: the Todo was closed with `reject`, but the receipt still suppressed regeneration. This made editorial review a one-shot workflow.

The fix keeps approval authority in the existing Todo lifecycle. The consumer now reads the exact approval scope and bound agent from durable Todo state, derives an immutable rejection revision, and routes the next draft into `retry-<revision>/`. A still-open, approved, malformed, cross-agent, or otherwise non-rejected gate remains fail-closed. Repeated rejection is handled as the same general attempt lifecycle rather than as a special second-draft branch.

This is intentionally inside the built-in `periodic-report` capability: it changes that capability's review lifecycle and does not introduce a second state machine or provider abstraction. The nearby future-facing pass was applied by centralizing receipt validation and attempt selection; no broader refactor is needed.

## Validation

- focused Python tests: 15 passed
- Ruff on changed Python paths: passed
- Python compile on changed modules: passed
- TypeScript control-plane typecheck: passed
- TypeScript control-plane tests: 288 passed
- `loopx canary premerge --from-git-diff --goal-id ark-4-0-goal`: 17 selected checks, 0 failures, 0 manual holds

## Boundaries

- no report content, private case evidence, credentials, local paths, or external-delivery behavior is included
- the generated lockfile from local dependency setup is intentionally excluded
- approval still grants authority only for the exact frozen payload; this change performs no external publication

### DIFF
--- a/loopx/capabilities/periodic_report/pending_intent.py
+++ b/loopx/capabilities/periodic_report/pending_intent.py
@@ -76,27 +76,191 @@ def _intent_key(intent: Mapping[str, Any]) -> str:
     ).hexdigest()[:24]
 
 
-def _receipt_path(runtime_root: Path, goal_id: str, intent: Mapping[str, Any]) -> Path:
-    return (
+def _attempt_dir(
+    runtime_root: Path,
+    goal_id: str,
+    intent: Mapping[str, Any],
+    *,
+    rejection_revision: str | None = None,
+) -> Path:
+    base = (
         runtime_root
         / "goals"
         / goal_id
         / "periodic_reports"
         / _intent_key(intent)
-        / "receipt.json"
     )
+    return base / f"retry-{rejection_revision}" if rejection_revision else base
+
+
+def _receipt_path(
+    runtime_root: Path,
+    goal_id: str,
+    intent: Mapping[str, Any],
+    *,
+    rejection_revision: str | None = None,
+) -> Path:
+    return _attempt_dir(
+        runtime_root,
+        goal_id,
+        intent,
+        rejection_revision=rejection_revision,
+    ) / "receipt.json"
 
 
 def _editorial_request_path(
-    runtime_root: Path, goal_id: str, intent: Mapping[str, Any]
+    runtime_root: Path,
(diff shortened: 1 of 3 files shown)


huangruiteng:
No blocking findings.

- **Motivation:** rejection previously closed the user Gate but left the consumption receipt permanently suppressing the intent, so an owner could not request a corrected draft through the governed path.
- **Solved:** the exact durable reject now reopens only its bound intent/agent, every revision is immutable, replay remains idempotent, and the test covers two consecutive reject cycles.
- **User/operator impact:** review becomes iterative without manual receipt edits, while approval continues to bind only one frozen payload.
- **Main risk:** reading active Todo state from the pending-intent path adds lifecycle coupling. The implementation bounds it to valid `approval_pending` receipts and exact action kind, decision scope, outcome, and agent identity; malformed or unresolved state stays suppressed.
- **Design judgment:** reusing Todo rejection as the authority source is cleaner than inventing a report-specific review state machine. Centralized attempt selection also avoids one-off “second draft” logic and supports arbitrary bounded revisions.
- **Future-facing pass:** applied within scope by extracting exact receipt validation and a reusable attempt selector. A wider control-plane refactor would not improve this narrow lifecycle fix.

## Validation performed

- 15 focused Python tests passed
- Ruff and Python compile passed
- TypeScript typecheck and 288 control-plane tests passed
- premerge gate passed 17 selected checks with 0 failures and 0 manual holds
- public/private boundary scan passed; local generated lockfile excluded

Merge decision: self-merge when required GitHub checks are green. No external publication authority is introduced.

PR #3754, opened 2026-08-30
fix(periodic-report): use actual work interval

huangruiteng:
## Summary

- derive the periodic-report window from the selected agent's durable run history and report facts instead of subtracting seven days from stage completion
- freeze the computed human-readable period in the editorial request
- reject an authored `period_label` that disagrees with the computed interval

## Motivation

A stage report covering less than one day could display a natural-week range. That made the artifact factually wrong even though its analysis and evidence lineage were valid.

## Validation

- `uv run pytest -q tests/capabilities/test_periodic_report_pending_intent.py tests/capabilities/test_periodic_report_adapters.py tests/presentation/test_periodic_report_html.py` — 45 passed
- Ruff and Python compile — passed
- `git diff --check` — passed
- `loopx canary premerge --from-git-diff --goal-id ark-4-0-goal` — passed; one selected canary, zero failures, zero holds
- public/private boundary scan — clean

## Boundary

This changes only local report generation semantics. It grants no publication or messaging authority and includes no project-specific evidence.

### DIFF
--- a/loopx/capabilities/periodic_report/pending_intent.py
+++ b/loopx/capabilities/periodic_report/pending_intent.py
@@ -147,7 +147,7 @@ def _decision_scope_text(value: object) -> str:
     return str(value or "")
 
 
-def _rejected_approval_revision(
+def _superseding_approval_revision(
     *,
     registry_path: Path,
     goal_id: str,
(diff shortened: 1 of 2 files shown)


huangruiteng:
No blocking findings.

- **Motivation:** the consumer conflated a weekly cadence with the report's factual work interval, so a valid stage analysis could publish a false date range.
- **Solved:** the frozen request now owns one computed interval from durable agent run history plus cited fact timestamps; the renderer receives that interval, and authored labels cannot contradict it.
- **User/operator impact:** short or irregular work stages show when the reported work actually happened while preserving the existing overview-to-depth editorial contract.
- **Main risk:** run history can be absent or contain malformed rows. The implementation ignores malformed or cross-agent rows and falls back to the cited facts, while preserving a non-empty protocol window for a single timestamp.
- **Design judgment:** keeping time authority in the consumer is more reusable than asking the editor to infer a date. The narrow helper centralizes computation without introducing a second report state machine.
- **Future-facing pass:** applied within scope by freezing the interval and validating the human label from the same source of truth; no wider renderer refactor is needed.

## Validation performed

- 45 focused periodic-report tests passed
- Ruff, Python compile, and diff checks passed
- premerge gate passed with one selected canary, zero failures, and zero manual holds
- public/private boundary scan passed; local generated lockfile excluded

Merge decision: self-merge when required GitHub checks are green. No external publication authority is introduced.

huangruiteng:
No blocking findings after the lifecycle follow-up.

The final diff now covers both halves of the root cause:

- factual time authority: derive and freeze the actual work interval from durable same-agent run history plus cited fact timestamps, and reject authored labels that disagree;
- revision lifecycle: a later same-scope `reject` or `cancel` supersedes the frozen generation and opens a fresh immutable retry, including the real sequence where an approval was recorded before the owner corrected the date.

Focused validation after the amendment:

- 10 pending-intent lifecycle tests passed, including later cancel after approve;
- Ruff passed;
- Python compile passed;
- diff hygiene and public/private boundary scan passed.

The earlier 45-test periodic-report suite and premerge canary remain the broader validation baseline; GitHub checks are rerunning for the amended commit. Self-merge remains conditional on required checks passing.

PR #3768, opened 2026-08-30
fix(chat): recover from legacy Codex model catalogs

### LINKED ISSUE #3767: Dashboard Chat: recover from legacy Codex model catalog schema
## Symptom

Dashboard Chat can route an ordinary conversational message correctly but fail before the Agent turn begins because Codex app-server rejects `thread/start`. The UI can only report a generic host failure, so the user receives no reply.

## Expected behavior

LoopX Chat should recover when the configured Codex model catalog is valid under an earlier schema but lacks a newly required model metadata field. Recovery must not edit, copy, expose, or permanently replace the user's catalog.

## Actual behavior

A catalog row that predates the required `base_instructions` field makes the current Codex app-server reject configuration loading at `thread/start`.

## Minimal synthetic reproduction

1. Point Codex `model_catalog_json` at a synthetic catalog whose model row follows the previous schema and omits `base_instructions`.
2. Start Dashboard Chat and create a Codex-backed Chat Session.
3. Observe that app-server initialization succeeds but `thread/start` fails before any model turn.

## Suspected product surface

`loopx/chat_agent.py` currently starts app-server with ambient configuration and collapses this schema mismatch into the generic host-tool gate. A narrow compatibility retry can classify only the known missing-field error, generate the same Codex binary's current built-in catalog in an isolated temporary profile, and override `model_catalog_json` for that child app-server process only.

The fallback should be bounded: no ambient catalog content is read or copied by LoopX, no permanent config changes occur, and unrelated configuration failures remain fail-closed.

## Validation already performed

A public-safe process probe reproduced the ambient `thread/start` failure and confirmed that a catalog generated from the same Codex binary's current built-in schema allows `thread/start` to succeed.

huangruiteng:
## Summary

- recognize only the Codex app-server bootstrap error for a legacy `model_catalog_json` missing `base_instructions`
- regenerate the selected Codex binary's current built-in catalog inside an isolated temporary `CODEX_HOME`, then retry only the LoopX Chat child process
- expose whether compatibility was applied and keep unrelated configuration failures fail-closed
- cover successful recovery, unrelated-error non-retry, and catalog-generation failure

LoopX never reads, copies, edits, or persists the user's configured catalog. The temporary override is scoped to one Chat app-server process and is removed after bootstrap.

The failure happens before the model turn, so intent routing cannot repair it. The compatibility owner belongs in the existing Chat app-server bootstrap boundary. Reusing the same Codex binary's built-in catalog avoids creating a second schema authority in LoopX, while the exact error classifier and one-retry limit keep the fallback narrow.

## Validation

- `python examples/loopx-chat-agent-smoke.py`
- focused pytest: 22 Chat session, attached-session broker, and CORS tests
- focused Ruff and `py_compile`
- real Codex app-server/model turn: returned the exact requested reply with `compatibility_applied=true`, no protected action, and no gate
- `loopx check` public/private scan: clean for both changed files
- exact-scope change-quality receipt `cqr_8117a4282cd955e73a11`
- `loopx canary premerge --from-git-diff --goal-id loopx-meta`: passed, no manual holds
- full pytest sampling reached 298 passes before a manual interruption; required remote checks remain authoritative for the full matrix

Future-facing pass: no broader refactor was needed; the adjacent session lifecycle remains the single runtime owner.

Closes #3767

### DIFF
--- a/examples/loopx-chat-agent-smoke.py
+++ b/examples/loopx-chat-agent-smoke.py
@@ -4,12 +4,14 @@
 from __future__ import annotations
 
 import json
+import os
 import queue
 import stat
 import sys
 import tempfile
 import time
 from pathlib import Path
+from unittest.mock import patch
 
 
 REPO_ROOT = Path(__file__).resolve().parents[1]
@@ -25,8 +27,26 @@ from loopx.chat_agent import (  # noqa: E402
 
 FAKE_CODEX = r'''#!/usr/bin/env python3
 import json
+import os
 import sys
 import time
+from pathlib import Path
+
+if sys.argv[1:3] == ["debug", "models"]:
+    marker = os.environ.get("LOOPX_FAKE_DEBUG_MODELS_MARKER")
+    if marker:
+        Path(marker).write_text("invoked", encoding="utf-8")
+    if os.environ.get("LOOPX_FAKE_DEBUG_MODELS_FAIL") == "1":
+        raise SystemExit(2)
+    print(json.dumps({"models": [{"slug": "fake-current", "base_instructions": "Current built-in instructions."}]}))
+    raise SystemExit(0)
+
+catalog_override = None
+for index, value in enumerate(sys.argv):
+    if value == "-c" and index + 1 < len(sys.argv):
+        key, separator, encoded = sys.argv[index + 1].partition("=")
+        if separator and key == "model_catalog_json":
+            catalog_override = Path(json.loads(encoded))
 
 turn_count = 0
 assert "goals" not in sys.argv, sys.argv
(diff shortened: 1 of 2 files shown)


huangruiteng:
No blocking findings at exact head `aea7c570596e9d890db9e42b22efbd950d71330c`.

None. The retry depends only on the selected Codex binary exposing its current built-in catalog after reporting the known legacy-schema error. If that catalog cannot be generated or validated, the code stops at an actionable host gate.

- **Motivation:** installed Dashboard Chat can fail before a model turn when an older ambient Codex catalog lacks a newly required field.
- **Solved:** the real app-server/model regression confirms thread bootstrap and the user reply both complete after the isolated retry.
- **Impact:** existing users recover without editing, reading, copying, or replacing their catalog.
- **Main risk:** accidentally converting unrelated configuration errors into a permissive fallback. The implementation limits classification to thread bootstrap, three stable schema tokens, one retry, and validates every generated model row.
- **Design:** the existing Chat session bootstrap remains the single runtime owner; LoopX reuses the same Codex binary's built-in catalog instead of owning another catalog schema. No broader companion refactor is needed.

Review lenses: no generic typed control-plane state was duplicated; Chat-specific wording stays out of kernel contracts; the default-behavior change is disclosed as a narrow compatibility retry; and the one-retry/fail-closed obligations are machine-enforced rather than prose guidance.

## Validation

- focused Chat smoke including success and both negative paths: passed
- focused pytest: 22 passed
- real Codex app-server/model response: passed, exact reply, no protected action or gate
- Ruff, `py_compile`, diff hygiene, public/private scan: passed
- change-quality receipt `cqr_8117a4282cd955e73a11`: valid
- premerge canary: passed, no manual holds, self-merge allowed
- remote DCO, dependency review, pytest, Windows, build, and Sonar: passed
- release upload/publish jobs were expected skips on a pull request

Merge decision: self-merge authorized by the owner request and supported by exact-head validation.

PR #3795, opened 2026-09-01
fix(periodic-report): normalize delivery binding refs

huangruiteng:
## Summary

- make periodic-report delivery Todo binding refs valid for every hexadecimal generation digest
- prefix the digest segment with `g` so the namespaced value always begins with a letter
- add focused regression coverage for a real digit-leading generation id

## Motivation and architecture

The periodic-report consumer creates a blocked delivery Todo before its approval gate. Generation digests can begin with a digit, while Todo capability binding refs require the value after the namespace to begin with a letter. A digit-leading report therefore generated local artifacts and then failed before durable gate/successor settlement. The fix stays inside the existing periodic-report capability owner and preserves the public Todo contract instead of weakening its validator.

## Validation

- `python -m pytest tests/capabilities/test_periodic_report_pending_intent.py -q` — 13 passed
- `python -m ruff check` on both changed files — passed
- `git diff --check` — passed
- `loopx canary premerge --from-git-diff --goal-id ark-4-0-goal` — passed; maintainability ratchet and changed-file compile passed

## Risk

Low and localized. Existing binding refs gain a stable `g` prefix only for newly generated delivery Todos; approval scope and frozen artifact identity are unchanged.

## Future-facing pass

No broader refactor was needed. A small helper centralizes the binding rule at the owning call site and keeps the Todo validator authoritative.

### DIFF
--- a/loopx/capabilities/periodic_report/pending_intent.py
+++ b/loopx/capabilities/periodic_report/pending_intent.py
@@ -77,6 +77,13 @@ def _intent_key(intent: Mapping[str, Any]) -> str:
     ).hexdigest()[:24]
 
 
+def _periodic_report_delivery_binding_ref(generation_id: object) -> str:
+    digest_suffix = str(generation_id).split("_")[-1][:16]
+    # Todo capability binding refs require the namespaced value to start with
+    # a letter. Generation digests are hexadecimal and may start with a digit.
+    return f"periodic-report:g{digest_suffix}"
+
+
 def _attempt_dir(
     runtime_root: Path,
     goal_id: str,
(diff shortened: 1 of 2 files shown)


huangruiteng:
No blocking findings in the exact two-file diff.

- Motivation: a valid hexadecimal generation id could strand periodic-report settlement after artifact generation because Todo binding values must start with a letter.
- Solved: the owning periodic-report capability now produces a contract-valid binding for both digit- and letter-leading digests, with a focused regression for the failing shape.
- Operator impact: local generation can proceed atomically to the blocked delivery successor and approval gate instead of leaving a repeatable pending intent.
- Main risk: new delivery Todo binding values add a stable `g` prefix. Approval scopes, artifact digests, and report generation ids do not change.
- Design judgment: keeping the normalization at the periodic-report binding construction site preserves the generic Todo validator as the single authority. A wider contract relaxation would be less safe.

Validation performed:

- focused periodic-report test module: 13 passed
- Ruff on both changed files: passed
- diff hygiene: passed
- premerge canary: passed, including changed-file compile and maintainability ratchet

Future-facing pass: no broader refactor is needed; the small helper is the nearest durable owner for this invariant.

Merge decision: technically ready, but repository review protection remains authoritative.

PR #3838, opened 2026-09-02
Add SWE-Marathon codex×LoopX evaluation (report + harness)

BouwenZhou:
在 `benchmark/swe-marathon/` 下加入一份 codex×LoopX 评测：裸 `codex`、codex 原生 `goal`、以及三种 LoopX 模式（ssh-goal / codex-cli / heartbeat）在 15 个 SWE-Marathon 任务上的对照（75 trial，模型 gpt-5.6，预算 ~30%）。

本 PR 为**合并后的单一 PR**，含：
- `README.md` — 简洁正式的评测报告（两 baseline + 三 LoopX 模式；matched 15 任务分母；机制观察为假设）
- `data.json` — pinned public-safe 聚合产物
- `agents/` — Harbor 适配器（LoopX treatment + codex 原生 goal baseline）
- `scoring/` — 评分/聚合/可视化（口径见 `_common.py`）
- `skills/` — 两个五模式 benchmark skill（内网信息已清洗）
- `runtime/` — 模式框架 + turn 驱动，含 automation 唤醒循环 `loopx_turn_runner.py`

已按前次 review 处理：移除公开 raw 轨迹（仅发布聚合，raw 私有）、matched 分母（构建失败计入不单臂剔除）、因果表述改为假设、复现命令仓库相对、补血缘。原 companion 代码 PR #3842 已并入本 PR 并关闭。

交互式聚合页（public-safe，无 raw 轨迹）：https://bouwenzhou.github.io/loopx/

PR #3854, opened 2026-09-02
chore: restore co-author attribution for #3838

huangruiteng:
## Why

The squash merge of #3838 omitted the `Co-authored-by` trailers for two contributors. The original merge commit is no longer the tip of `main`, so amending it would rewrite subsequent mainline history.

## Change

Add one metadata-only empty commit that restores attribution for:

- Ran Wenxing (`@piaji-68`)
- Wanli Li (`@Wanli-Lee`)

## Safety and validation

- No repository files or tree contents change.
- `origin/main...HEAD` has an empty diff and both revisions have the same tree.
- Both `Co-authored-by` trailers and the maintainer DCO `Signed-off-by` trailer parse successfully.
- This approach preserves the existing `main` history.

Closes the attribution follow-up requested on #3838.


huangruiteng:
## Maintainer validation

- **Scope:** contributor metadata only; exact head `ac0818faabe754d1002fd950435c6e054f93c60b` changes zero files and has the same tree as its `main` parent.
- **Attribution:** GitHub resolves the two `Co-authored-by` trailers to `@piaji-68` and `@Wanli-Lee`; the DCO `Signed-off-by` trailer is present.
- **Checks:** DCO Sign-off passed; dependency-review passed; local exact-tree comparison and trailer parsing passed.
- **Risk:** no runtime, documentation, benchmark data, permissions, or public/private boundary changes. Rewriting the original squash commit would disturb three later mainline commits, so this additive metadata-only commit is the safe repair.
- **Future-facing review:** no code boundary or refactor is applicable to an attribution-only commit.

Decision: **self-merge** as a small contributor-metadata repair. The squash message will explicitly preserve both co-author trailers.

PR #3878, opened 2026-09-03
feat(benchmark): bind case insights to terminal runs

huangruiteng:
## Summary

- make `case_insight_projection` an explicit child of one active exact terminal experiment-board run
- preserve the run row as sole authority for arm, score, countability, integrity, and treatment fidelity
- reject orphan, pre-terminal, outcome-mismatched, or later-invalidated insight attachments
- document the bounded redaction/evidence contract in benchmark-toolkit, its skill, and RFC #3812
- validate two-arm, factorized four-arm, and five-mode study shapes, including local CLI preview/execute/readback/dashboard flow

This extends the provider-neutral study upload contract implemented after RFC #3812 and checks compatibility with the public five-mode study shape contributed in #3838. It does not add a remote provider, publish benchmark data, launch jobs, or change scoring.

## Validation

- `python -m ruff check loopx/capabilities/benchmark_toolkit/experiment_board.py loopx/capabilities/benchmark_toolkit/study_projection.py loopx/capabilities/benchmark_toolkit/catalog_entry.py tests/capabilities/test_benchmark_study_projection.py`
- `pytest -q tests/capabilities/test_benchmark_*.py` — 231 passed
- `python -m compileall -q loopx/capabilities/benchmark_toolkit loopx/cli_commands/benchmark_study.py`
- `git diff --check`
- `loopx canary premerge --from-git-diff --git-diff-base origin/main --goal-id loopx-deepswe-bench --tier standard` — 18/18 checks passed; public-boundary scan clean

## Review / merge hold

The generic canary classifies all benchmark-sensitive paths as requiring manual maintainer review. The owner has separately authorized self-merge for bounded benchmark helper changes. This PR changes no runner, scoring, leaderboard, submission, permission, or job-launch behavior, so I will perform exact-head review and use that authorization only if checks remain green.

PR #3812, opened 2026-09-01
docs(benchmark): propose study upload and dashboard RFC

huangruiteng:
## Summary

- keep benchmark-native scoring and experiment-board rows authoritative
- define a provider-neutral study manifest and upload/readback envelope
- specify campaign, arm, case, and run dashboard projections
- split follow-up delivery into a contract PR and a read-only dashboard PR

## Validation

- `git diff --check`
- local Markdown link readback
- public/private boundary scan
- `examples/docs-governance-smoke.py` currently fails on unchanged `origin/main` because it still expects the previous RFC index section names
- MkDocs build not run because MkDocs is not installed in this environment

No runner, scoring, upload, provider, or dashboard behavior changes in this PR.

PR #3886, opened 2026-09-03
Add SWE-Marathon case-insight projections

BouwenZhou:
Project the five-mode SWE-Marathon study's per-case observations into LoopX's **public-safe** `benchmark_case_insight_projection_v0` records (follows #3878 / RFC #3812).

- `scoring/case_insights.py` — derive insight records from the pinned `data.json`.
- `scoring/test_case_insights.py` — assert every record satisfies the upstream contract (field set, terminal outcome_status, public_safe, redaction-attested, ≤600-char text, handle-only evidence_refs).
- `case_insights.json` — 5 pinned records: build-failure gate-zero, codex-cli unattended continuation idle-churn, heartbeat stable continuation.

All are `public_safe` with `producer_redaction_attested=true`; `evidence_refs` carry only `run:<case>__<arm>` handles (no raw trajectories, no timestamps). Attaching each record to an experiment-board terminal run row is left as a follow-up (board run rows are not yet uploaded).


BouwenZhou:
改由 Wanli-Lee 身份重开（PR opener + commit author 一致），本 PR 关闭。

PR #3890, opened 2026-09-03
fix(dsh): require target-install compatible LoopX

### LINKED ISSUE #3867: [Bug]: PyPI 0.5.3 lacks packaged workflow skills required by DSH plugin
### Preflight

- [x] I searched existing issues and discussions for this behavior.
- [x] This report is not an unpatched security vulnerability.
- [x] The report and attachments contain no credentials or private/internal material.

### Issue origin

Observed or reproduced in a real environment

### LoopX version or commit

0.5.3

### Host or runtime surface

Other

### LoopX area

Control plane (goals, todos, quota, scheduler, registry, runtime)

### Problem

Follow-up to #3796: PyPI publication now succeeds, but the released 0.5.3 distribution does not contain the packaged workflow-skill payload required by the DSH-native plugin integration. As a result, DSH users cannot complete /loopx-init even though the CLI itself installs and reports the expected version.

### Minimal reproduction

1. Install dsh-loopx-plugin@0.1.1-beta.3 in DSH.
2. Use a clean Python 3.13 environment with the official PyPI index.
3. Run /loopx-init.
4. The bootstrap installs loopx>=0.5.3 successfully, then the workflow-skills compatibility probe fails.

### Expected behavior

The PyPI package for the required release should include the workflow-skill payload. workflow-skills should report an available source for host surface deepseek-harness-native, and /loopx-init should install the DSH-native skills.

### Actual behavior

The managed CLI reports loopx 0.5.3, then /loopx-init stops with: LOOPX_INIT_FAILED: stage=probe; kind=incompatible. The installed LoopX CLI does not support the DSH-native skill contract. The workflow-skills inspection returns ok: false; source.available: false; source.kind: missing; and says the active LoopX installation does not contain the packaged workflow skills. Its expected source revision is 0.5.3.

### Sanitized diagnostics

```shell
The official PyPI package was installed in a clean Python 3.13 virtual environment. No credentials, private data, raw session contents, or local paths are included here.
```

### Additional context

Related issue: #3796. That issue addressed the package not being published to PyPI; this is a separate artifact-completeness issue after successful PyPI installation.

huangruiteng:
## Summary

- require LoopX 0.5.4 from the DSH managed runtime because it is the first release that discovers packaged workflow skills after Linux `pip --target` installation
- publish the fix as DSH LoopX Plugin 0.1.1-beta.4 and update its pinned documentation
- verify both the locally built wheel and the published PyPI package through the real DSH managed-runtime install shape

## Root cause

LoopX 0.5.3 does carry the workflow-skill payload in its wheel. The failure is narrower: under the Linux `pip --target` layout used by the DSH plugin, `importlib.metadata.files()` does not expose the installed `share/loopx/skills` rows, so 0.5.3 reports the source as missing. LoopX 0.5.4 added an explicit distribution-root fallback and succeeds in the same environment.

The plugin still accepted `loopx>=0.5.3`, so an already installed 0.5.3 satisfied pip and could not self-repair. This change moves that compatibility floor to 0.5.4 and locks the exact target-install behavior into the release pipeline.

Fixes #3867

## Validation

- workflow-skill tests: 27 passed
- DSH TypeScript typecheck and Vitest: 170 passed
- DSH peer-range, package build/artifact/profile, and real GoalBar runtime smokes passed
- root TypeScript control-plane typecheck and tests: 483 passed, 1 PostgreSQL integration test skipped as expected
- Ruff and strict mypy passed
- release-artifacts smoke passed
- clean Debian/Python 3.13 Docker smoke passed with PEP 668, the current LoopX wheel, the beta.4 plugin tarball, real DSH, and first-session skill discovery
- public/private boundary scan and `git diff --check` passed
- risk-based LoopX premerge: 18/18 checks passed, no manual hold
- exact change-quality receipt: `cqr_043546076f41664d77c0` for scope fingerprint `043546076f41664d77c0754a556c4790658379288ad7083dc6f4961ea84be08a`

## Risk and compatibility

This does not add a second workflow-skill installer or change LoopX authority boundaries. It narrows the plugin's minimum compatible LoopX version to the first version known to support its existing Linux managed-runtime layout. The new release checks prevent a future wheel from publishing successfully while remaining unusable through that layout.


huangruiteng:
Findings: none on exact head `648fcf0bfb783f9a406db3b6be754a60021c812c`.

Open questions / assumptions: none. The compatibility claim was reproduced against the released 0.5.3 package and rechecked against 0.5.4 in the same Linux/Python 3.13 target-install shape.

Product / architecture judgment: the user-facing failure is solved at the existing compatibility boundary. The plugin now refuses the one LoopX version that cannot discover workflow skills in its managed `pip --target` layout, while keeping workflow-skill source and installation authority in LoopX. This avoids introducing a second installer or copying packaged skills into another distribution. The main compatibility risk is future wheel layout drift; the release workflow now exercises the real managed-runtime shape before and after PyPI publication.

Validation: 27 workflow-skill tests; DSH typecheck and 170 Vitest tests; peer-range, package, artifact, real-profile, and GoalBar smokes; root control-plane typecheck and 483 tests (1 expected PostgreSQL integration skip); Ruff; strict mypy; release-artifacts smoke; clean Debian/Python 3.13 PEP 668 end-to-end install; public/private scan; diff check; and risk-based LoopX premerge 18/18 with no manual hold. Exact change-quality receipt: `cqr_043546076f41664d77c0`.

Future-facing pass: applied at the release-verification boundary by replacing the weaker venv-only PyPI check with the actual DSH target-install contract. No broader refactor is needed for this narrow compatibility fix.

Merge decision: self-merge is authorized by the owner. Admin bypass is limited to the repository review/time gate; correctness, DCO, privacy, and exact-head validation remain satisfied.

PR #3903, opened 2026-09-03
fix(change-window): allow read-only SSH transport

huangruiteng:
## Summary

- classify Git SSH services before repository-provider health checks
- allow read-only `git-upload-pack` and `git-upload-archive` even when the local change-window provider has drifted
- keep `git-receive-pack` fail-closed behind provider health and the configured time policy

## Why

The change-window capability governs public repository writes. A drifted local provider was rejecting every SSH transport before it determined whether the service was a write, so a harmless `git fetch` failed with the same message as a blocked push. That made recovery harder and caused agents to misreport an open nighttime window as closed.

## Validation

- focused read-only/provider-drift regression: passed
- live hook invocation against the drifted checkout: returned `reason=read_only_ssh_transport`, exit 0
- Ruff, Python compile, and `git diff --check`: passed

The broader change-window test file still has pre-existing failures caused by the machine-wide Git hook layout leaking into its temporary-repository fixtures; the two SSH-focused cases pass and the unrelated fixture issue is not changed here.

### DIFF
--- a/loopx/capabilities/repository_change_window/git_hook.py
+++ b/loopx/capabilities/repository_change_window/git_hook.py
@@ -1047,6 +1047,19 @@ def run_git_hook_provider(
     hook_stdin: bytes = b"",
     now: datetime | None = None,
 ) -> dict[str, Any]:
+    if event == SSH_TRANSPORT_EVENT and not _ssh_transport_is_push(hook_args):
+        return {
+            "ok": True,
+            "schema_version": "repository_change_window_hook_result_v1",
+            "status": "allowed",
+            "event": event,
+            "exit_code": 0,
+            "guarded_change": False,
+            "previous_hook_invoked": False,
+            "policy_evaluated": False,
+            "reason": "read_only_ssh_transport",
+        }
+
     context = resolve_repository_context(repo_path)
     state = _read_state(_state_path(context.common_dir))
     enforcement = _state_enforcement_level(state)
@@ -1078,8 +1091,6 @@ def run_git_hook_provider(
             hook_args=hook_args,
             hook_stdin=hook_stdin,
         )
-    elif event == SSH_TRANSPORT_EVENT:
-        guarded_change = _ssh_transport_is_push(hook_args)
     else:
         guarded_change = True
     if guarded_change and not decision["allowed"]:
--- a/tests/capabilities/test_repository_change_window.py
+++ b/tests/capabilities/test_repository_change_window.py
@@ -484,6 +484,8 @@ def test_reference_guard_uses_fake_clock_and_shared_linked_worktree_state(
     assert allowed_ssh["status"] == "allowed"
     assert allowed_ssh["guarded_change"] is False
     assert allowed_ssh["previous_hook_invoked"] is False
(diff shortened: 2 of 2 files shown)


huangruiteng:
No blocking findings on exact head `e6f244f55326bafe3915ab0ffcc5f5299217bc99`.

Product/architecture judgment: the change-window capability owns public repository writes, not read transport. The prior ordering checked provider health before classifying the SSH service, so provider drift incorrectly denied `git-upload-pack` and made a harmless fetch look like a closed write window. This patch moves the existing typed service classification to the outer boundary: upload/archive return a typed no-policy result immediately, while receive-pack still fails closed through provider-health and time-window checks. That is the smallest reusable fix and does not weaken commit or push enforcement.

Validation:

- focused pytest: 2 passed, including provider drift allowing upload-pack/upload-archive while receive-pack remains `provider_drift`
- live invocation against the affected checkout: exit 0, `reason=read_only_ssh_transport`
- Ruff, py_compile, diff hygiene, and public-boundary scan: passed
- change-quality receipt: `cqr_ea0b58b9249dc7db6a63`, exact scope valid
- `loopx canary premerge --from-git-diff --goal-id loopx-meta`: passed, standard tier, 2 Python files, one catalog canary, no failures, no manual holds, `self_merge_allowed=true`
- full test-file run is not used as a blocker because unrelated existing temporary-repository fixtures inherit this machine's external hook layout; the two changed semantic paths pass independently

Merge decision: self-merge after required GitHub checks settle. Future-facing pass applied: the redundant post-classification SSH branch was removed; no broader abstraction is needed.

huangruiteng:
Final qualification after rebasing onto the current main: no blocking findings on exact head `4c24eac9c42ab8349cc19ecdae829f5f9800f389`.

Root cause and design judgment remain unchanged: repository change-window authority governs writes. The former ordering inspected provider health before classifying the SSH service, so provider drift could deny `git-upload-pack` and misreport a harmless fetch as a closed write window. The final patch classifies the service first, immediately allows upload/archive with `policy_evaluated=false`, and keeps `git-receive-pack` fail-closed through provider-health and time-window policy.

Validation on the rebased exact diff:

- focused real-Git pytest: 2 passed, covering linked-worktree write enforcement and provider drift allowing upload-pack/upload-archive while receive-pack remains guarded
- Python 3.13 compile, diff hygiene, and public-boundary scan: passed
- change-quality fingerprint: `aa07d1bfca53158e361860aa3380fa599929c16cc69fb3c95a69222ecff0a4f2`
- receipt: `cqr_aa07d1bfca53158e3618`, exact scope valid; no additional safe-fix pass
- standard canary: passed, one selected catalog canary, no failures or manual holds, `self_merge_allowed=true`
- the prior GitHub pytest run was cancelled while running the unchanged fast-test suite, not failed; pushing this rebased head started a fresh run

Future-facing pass: the existing typed SSH-service classifier remains the single protocol decoder, and the early return removes read traffic from the policy/provider-state boundary without creating a second configuration or transition authority.

Merge decision: self-merge with the maintainer-authorized admin path.

PR #3937, opened 2026-09-04
fix(authority): remove status reverse import

songoow:
## Summary

- replace the authority-shadow adapter reverse import from top-level `loopx.status` with the existing goals-owned event-log candidate primitive;
- inject the same goal-local path resolver explicitly, preserving event-log discovery semantics;
- restore the existing Todo read-model bounded-context smoke on the exact current `main`.

## Regression evidence

On exact official `main` `bf5dda59c`, `examples/control_plane/todo-readmodel-boundary-smoke.py` fails with:

`loopx/control_plane/coordination/local_authority_shadow_adapter.py:690 from ...status import`

The smoke passes on this branch. The change keeps the adapter inside lower-level control-plane owners and does not alter authority, writer, outbox, or projection behavior.

## Validation

- `python examples/control_plane/todo-readmodel-boundary-smoke.py` — passed
- authority-shadow drain and runtime-shadow writer capture suites — 18 passed
- focused Ruff and `git diff --check` — passed
- `loopx check` public/private boundary scan — clean
- `loopx canary premerge --from-git-diff --git-diff-base official/main` — 12 selected checks passed, 0 failures, 0 manual holds

## Future-facing scope pass

Applied: reused the existing goals-owned primitive and dependency-injection boundary instead of adding another event-log resolver or compatibility wrapper. No broader module move is needed for this regression.

### DIFF
--- a/loopx/control_plane/coordination/local_authority_shadow_adapter.py
+++ b/loopx/control_plane/coordination/local_authority_shadow_adapter.py
@@ -686,12 +686,17 @@ def _read_state_text(path: Path) -> str:
 
 
 def _event_present(sources: _GoalSources, event_id: str) -> bool:
+    from ..goals.active_state_event_projection import state_event_log_candidates
+    from ..goals.path_resolution import resolve_goal_local_path
     from ...event_sourced_state import AppendOnlyStateEventStore
-    from ...status import state_event_log_candidates
 
     if sources.goal is None:
         return False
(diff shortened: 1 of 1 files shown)


songoow:
Superseded by upstream PR #3938, now merged in official `main@e26928f24`. The merged implementation removes the same `status` reverse import, injects `resolve_goal_local_path`, and passes the boundary smoke; the remaining tree difference from this branch is import ordering only. Closing this duplicate so maintainers have a single source of truth.

PR #3965, opened 2026-09-05
fix(dashboard): unify machine and Goal capability discovery and settings views

huangruiteng:
## Summary

- Machine and Goal settings now discover the same configuration capability catalog. Previously the machine API returned only registered machine namespaces, so sharing navigation widgets still left most capabilities invisible.
- Reuse the typed detail summary, provenance display, editor status and scope-aware editability predicate across both settings pages.
- Show Goal-only capabilities with an explicit bilingual explanation. Do not turn catalog discovery into machine write authority: unsupported capabilities have no machine form, JSON fallback or mutation actions. Machine namespace persistence and Goal override transactions remain separate.
- Include the production chat bundle and regression coverage. No machine policy, Goal state, defaults, or delivery authority is changed by installation.

## Validation

- 79 focused Python tests, including the real Chat HTTP handler against disposable storage, namespace rejection and existing preview/apply/rollback contracts.
- TypeScript checking and production chat build.
- Shared workspace source-contract test.
- Packaged browser E2E: complete catalog, Goal-only negative write boundary, periodic-report preview/apply, English/Chinese and narrow-screen layouts.
- Diff whitespace and added-line public/private scans.

The owner reviewed the synthetic-data UI preview and requested this PR. Screenshots, local runtime data and credentials are not included. This fixes discovery and UI reuse; it does not add machine-level runtime inheritance to capabilities that currently support Goal configuration only.


huangruiteng:
No blocking findings after refinement. The owner approved the updated preview and explicitly authorized self-merge and local installation.

The machine API now discovers the same configuration capabilities as Goal settings without granting unsupported machine writes. Both scopes reuse detail/provenance and editability rules; their persistence transactions remain separate. Readable fields and actions precede collapsed raw JSON. Normal editable-state contract banners are removed; read-only explanations remain. The existing preview/revision fence is unchanged.

Validation: 79 focused Python tests (including real Chat HTTP against disposable storage), TypeScript checking/production build, shared workspace contract test, and packaged English/Chinese desktop/narrow-screen E2E including keyboard disclosure and negative machine-write checks passed. Public/private and diff checks passed. Initial premerge runs exposed a child-process Python selection issue (also reproduced on the baseline) and an import-formatting module-line budget regression; the former passes with the project Python explicitly selected, and the latter was corrected without widening the budget. Final premerge was rerun on the refined head.

The future-facing refinement is bounded shared discovery and presentation, not a second configuration authority. This does not add machine-default runtime support to Goal-only capabilities or change existing policies. Local preview files and intermediate bundles are excluded. Merge decision: approved for owner-authorized admin self-merge after the final gate passes.

PR #3966, opened 2026-09-05
fix(dashboard): enable persistent direct Goal drag ordering

huangruiteng:
## Summary

- Restore direct mouse dragging of active Goal titles in the personal workspace sidebar, with an insertion marker and click suppression. Escape cancels the gesture.
- Persist display order per status source in browser storage, retaining absent/stopped identities across partial projections. This is not scheduler priority and grants no lifecycle authority.
- Keep move-up/down controls hidden behind a small reorder toggle as a touch and keyboard alternative, with accessible announcements. Existing selection, stop, resume and stopped-Goal grouping remain intact.
- Ship the rebuilt Chat assets. The owner approved the default and expanded first-screen previews before commit.

## Design and risk

The current sidebar had no reorder wiring. A small typed ordering module owns reconciliation; a local hook owns pointer gestures and persistence. No backend, capability, provider, Goal configuration, or scheduling contract changes. New Goals retain source order until the user reorders them. Browser-local preferences do not synchronize across devices; touch uses the explicit controls and retains scrolling.

## Validation

- TypeScript check and production Chat build passed.
- Goal-order invariants and personal-workspace contract smoke passed.
- Packaged personal-workspace browser smoke passed, including real mouse dragging, reload persistence, Escape cancellation, no accidental selection, keyboard moves/boundaries, lifecycle regressions, English/Chinese and narrow-screen coverage.
- Packaged status-source-switch browser smoke passed.
- Final risk-based premerge gate passed: zero direct/run failures and zero manual holds. Public/private boundary and diff hygiene passed.
- Build emits the existing large-chunk advisory; no validation failures. Native desktop interaction was not separately automated; packaged Chromium exercised the production assets.

Only product source, durable regression tests and final referenced bundle assets are included. Screenshots and local runtime data are excluded. Owner-authorized self-merge after final review.

PR #3967, opened 2026-09-05
test(dashboard): keep Goal order smoke compatible with Node 20

huangruiteng:
## Summary

Fix the Node 20 Frontstage Pages test regression introduced by #3966: compile the typed order helper with the repository TypeScript compiler, then run its JavaScript invariants. Do not require Node's newer experimental type-stripping flag.

Only the test runner and its module import change. Product code, shipped assets, lockfile, and runtime behavior are unchanged. No private artifacts included.

## Validation

- Order invariants pass on local Node and explicitly on Node 20.
- Full packaged Personal Workspace smoke is also run under Node 20 before merge.
- Diff/public-boundary and risk-based premerge checks pass.

This is a bounded compatibility correction, not a weakened assertion or increased CI version requirement. It retains the same typed product module and all behavioral assertions.


huangruiteng:
Final review: no blockers. The two-line test-only correction removes the unsupported Node flag without dropping any assertions or changing production assets. The full packaged workspace browser smoke and order invariants now pass under Node 20, matching the Frontstage Pages runtime. Final premerge, diff hygiene and public-boundary checks passed with no failures or holds. DCO/dependency checks pass; hosted build jobs remain pending, not claimed as passed. Self-merging as the owner-authorized compatibility follow-up to #3966.

PR #3991, opened 2026-09-06
fix(update): retry bootstrap downloads with safe diagnostics

huangruiteng:
Archive updates stopped on a transient installer-download error with only a curl message. They now retry selected HTTP/network failures up to three times, download into a private temporary file, and execute only a complete successful response. JSON and text reports expose the attempt HTTP status and curl return code without recording response bodies, headers, URLs or raw curl errors.

The download has a 60-second budget capped by the command timeout; installer execution shares the remaining command budget and is never retried. Plan previews use the managed apply command. This affects archive bootstrap downloads only; pip/pipx, read-only checks and later archive transfers retain their existing behavior.

Validation: 46 focused update/download/activation/install-freshness/release tests passed; one Windows-only test skipped on macOS. The update smoke, new-module Ruff checks, changed-file syntax lint and public-boundary scan passed. Coverage includes transient 403 recovery, permanent HTTP failures, partial-body rejection, timeout budgets, redaction and no installer replay. No credentials or private runtime evidence are included.

### DIFF
--- a/docs/guides/installing-loopx.md
+++ b/docs/guides/installing-loopx.md
@@ -140,6 +140,23 @@ loopx update plan        # read-only command, validation, and rollback plan
 loopx update apply       # explicit local-environment mutation
 ```
 
+For archive installs, `update apply` downloads the bootstrap installer to a
+private temporary file before executing it. Downloading is limited to three
+attempts, a 60-second total download budget (or the smaller command timeout),
+and 20 seconds per transfer, with 1- and 2-second retry delays. Transient
+403/408/429/500/502/503/504 responses and selected connection/transfer failures
+are retried; 401/404 and certificate-verification failures stop immediately.
+Installer execution is never retried, and partial downloads are never executed.
+The command timeout covers downloading and installer execution together.
+
+JSON `execution.installer_download` and the text execution report show the
+stage and each attempt's HTTP status and curl exit code. HTTP `0` means no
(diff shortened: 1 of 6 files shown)


huangruiteng:
Self-review of b1839b8728df09678c46a4d862dd586b240df0ee: no blocking findings.

The changed surfaces are archive bootstrap execution, its plan/diagnostic rendering, focused regression coverage and operator documentation. `run_archive_installer` truncates the private temporary file between attempts, admits execution only after a successful nonempty 2xx transfer, shares the command budget, and never retries the installer. Failure diagnostics contain only stage, attempt, HTTP status and curl exit code. `_command_for_source` reuses the existing action-command builder so previewed commands take the same path. Package-manager and read-only paths retain their existing call contracts.

Risk-based premerge validation: 46 update/download/activation/install-freshness/release tests passed, update smoke passed, new files passed Ruff, touched files passed syntax lint, and the public-boundary scan passed. The single skipped test is Windows-only on macOS; Windows CI remains part of the PR checks. Existing whole-file lint findings outside this change were not broadened into cleanup. A real GitHub Pages bootstrap download returned HTTP 200 and a complete shell script; execution was intercepted, so that probe performed no installation. No manual holds; merge remains contingent on inspection of CI results.

Scope limit: retries apply only to the initial bootstrap download, not the bootstrap script's subsequent archive/API requests. This is sufficient for the reproduced failure without introducing a new generic network client or retrying partially applied installations.

PR #4004, opened 2026-09-06
fix(lark): verify bot mentions and preserve default member-read scope

huangruiteng:
## Summary

- Verify native mention targets with `im +chat-members-list --member-types user,bot`, preserving the declared ID type, bot identity, exact chat and pagination. The legacy member endpoint does not establish bot absence.
- Accept bot app-ID readback only through an unambiguous mapping from the same chat's verified bot-member bucket. Unknown IDs, human-bucket aliases, ambiguous mappings and denied member reads still fail closed.
- Document that `im:chat.members:read` is already in the canonical core/recommended bundle, and lock both the bundle and generated approval URL with a regression test. Configuration is not provider authorization; existing apps still require approval.

## Commit grouping and placement

Runtime + behavior tests/smoke, followed by default-scope contract test + docs. Owned by the existing bundled Lark extension's outbound inbox boundary; no new capability or global permissions added. Future-facing pass retained the existing shared readback matcher with an optional verified alias input; no speculative abstraction.

## Validation

- 122 focused tests passed (inbox reactions/delivery, bot scopes, group history, goal channel).
- Event inbox smoke passed.
- Ruff passed for runtime/tests; smoke passes with its existing E402 bootstrap exception (unfiltered Ruff reports two pre-existing import-order errors).
- git diff --check passed.
- LoopX check: six public files clean, zero errors; three unrelated local registry/history warnings.
- Authorized live provider preflight passed. One message was sent; independent readback confirmed its native mention and complete body. Live mget uses app_id while chat history uses open_id; the repaired matcher verified the saved intent using a freshly read same-chat bot mapping, without resending.

## Review hold / boundaries

Requires separate review: changes an identity-verification path. Not self-merged or globally installed. No permission grants, recipient-name fallback, production operations, or repeated outbound messages. Private identifiers, content, logs, local state and credentials are excluded. Existing runtime remains untouched; qualified worktree used for the bounded validation.

### DIFF
--- a/examples/lark-event-inbox-smoke.py
+++ b/examples/lark-event-inbox-smoke.py
@@ -347,11 +347,14 @@ def main() -> None:
                     return successful_runner(args)
                 if args[3:6] == ["im", "chats", "get"]:
                     return {"returncode": 0, "stdout": "{}", "stderr": ""}
-                if args[3:6] == ["im", "chat.members", "get"]:
+                if args[3:5] == ["im", "+chat-members-list"]:
                     return {
                         "returncode": 0,
                         "stdout": json.dumps(
-                            {"items": [{"member_id": "ou_reviewer_fixture"}]}
+                            {
+                                "users": [],
+                                "bots": [{"member_id": "ou_reviewer_fixture"}],
+                            }
                         ),
                         "stderr": "",
                     }
--- a/examples/lark-event-inbox-smoke.py
+++ b/examples/lark-event-inbox-smoke.py
@@ -352,8 +352,16 @@ def main() -> None:
                         "returncode": 0,
                         "stdout": json.dumps(
                             {
-                                "users": [],
-                                "bots": [{"member_id": "ou_reviewer_fixture"}],
+                                "ok": True,
+                                "data": {
+                                    "users": [],
+                                    "bots": [
+                                        {
+                                            "app_id": "cli_reviewer_fixture",
+                                            "member_id": "ou_reviewer_fixture",
+                                        }
+                                    ],
+                                },
                             }
                         ),
                         "stderr": "",
@@ -383,10 +391,7 @@ def main() -> None:
                                         "mentions": [
                                             {
                                                 "key": "@_user_1",
-                                                "id": {
-                                                    "open_id": readback_open_id,
-                                                    "user_id": "fixture-user-id",
-                                                },
+                                                "id": readback_open_id,
                                                 "name": "Reviewer",
                                             }
                                         ],
@@ -401,14 +406,14 @@ def main() -> None:
 
             return run
 
-        mention_reply = '<at user_id="ou_reviewer_fixture">Reviewer</at> 已记录并修正。'
+        mention_reply = '<at open_id="ou_reviewer_fixture">Reviewer</at> 已记录并修正。'
         normalized_mention = reply_lark_event_inbox(
             project=project,
             config_path=config,
             message_id="om_review_2",
             text=mention_reply,
             execute=True,
-            runner=mention_runner(readback_open_id="ou_reviewer_fixture"),
+            runner=mention_runner(readback_open_id="cli_reviewer_fixture"),
         )
         assert normalized_mention["status"] == "sent_verified", normalized_mention
 
--- a/loopx/extensions/lark/README.md
+++ b/loopx/extensions/lark/README.md
@@ -247,6 +247,14 @@ permission or silently enables an external write.
   `im:message.p2p_msg:readonly`
 - 交互/文档 sink：`cardkit:card:read/write`、`docs:document.comment:read/create/delete`
 
+`im:chat.members:read` 是原生 @ 身份校验的核心权限，也适用于 @ 机器人。
+发送前使用 `im +chat-members-list --member-types user,bot --page-all` 查询
+精确成员身份；旧的 `chat.members get` 不能作为机器人不在群内的证据。
+默认权限清单不代表已有应用已获授权：旧应用仍需在开发者后台申请并完成审核，
+LoopX 不会自动授予权限，也不会在成员读取失败时绕过 @ 校验。
+若消息回读将机器人标为 `app_id`，仅接受当前群机器人列表中已验证且无歧义的
+`member_id` ↔ `app_id` 映射；名称相同不构成身份验证。
+
 拿到 App ID（`cli_xxx`）后，可在开发者后台一键批量申请：
 
 ```text
--- a/loopx/extensions/lark/inbox_reply.py
+++ b/loopx/extensions/lark/inbox_reply.py
@@ -275,6 +275,7 @@ def _deliver_lark_inbox_outbound(
         )
 
     expected_mentions = expected_lark_mention_identities(reply_text)
+    bot_alias_candidates: dict[str, set[str]] = {}
     if expected_mentions:
         member_identity_sets: dict[str, set[str]] = {}
         membership_failed = False
@@ -284,10 +285,11 @@ def _deliver_lark_inbox_outbound(
                 base
                 + [
                     "im",
-                    "chat.members",
-                    "get",
+                    "+chat-members-list",
                     "--chat-id",
                     chat_id,
+                    "--member-types",
+                    "user,bot",
                     "--member-id-type",
                     identity_kind,
                     "--page-all",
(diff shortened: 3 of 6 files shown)


huangruiteng (commented):
## Request changes conclusion (author-owned PR; GitHub blocks formal self-review)

Exact head reviewed: `c7900d89882db41e0d43c3fe4d70b7e7525103fc`

### Findings

- **[P1] 混合 identity kind 会让已验证的 bot app-ID 回读被错误判成歧义。** 触发条件是同一条消息同时包含不同 `member-id-type` 的 mention，例如 bot 使用 `open_id`、真人使用 `user_id`。[`_deliver_lark_inbox_outbound`](loopx/extensions/lark/inbox_reply.py) 会分别查询两种 ID；同一个 bot 因而在两次 `bots[]` 回包中产生 `ou_bot` 和 `u_bot`，但 278/316 行把它们合并到仅按 `app_id` 分组的一个 set。486-489 行又要求整个 set 长度为 1，于是不给 `lark_readback_matches_outbound` 传任何 alias。若 mget 按本 PR 要修复的真实场景返回 bot `app_id`，消息已经发送，却落成 `sent_unverified`，收件箱 reaction 不会完成，后续恢复路径会把一次成功发送当失败处理。最小修复是让候选映射保留 `identity_kind`，或先只保留本次 outbound 中实际期待的 member ID，再判断 app-ID 是否唯一；并增加“bot open_id + human user_id、同一 bot 在两次成员查询均出现、mget 返回 app_id”的回归测试。我的独立探针稳定得到：`status=sent_unverified, ok=false, reply_verified=false`。

### 动机

旧实现用 legacy `im chat.members get` 验证 mention，但它不能证明 bot 不在群内；同时发送后的 mget 可能把 bot mention 回读成 `app_id`，而 outbound 使用的是 `open_id/user_id/union_id`，导致真实、正确的发送无法完成强回读。受影响的是 CLI 的 inbox reply/send 与 Goal Topic runtime 调用链。这个问题会把外部写已成功的场景留在未验证状态，因此修复方向是必要且与 v1.0 可靠性直接相关。非目标也清楚：不按显示名猜身份、不自动授权 scope、不改变普通文本消息。

### 改动思路

入口仍是 `reply_lark_event_inbox` / `send_lark_inbox_message`，决策与副作用仍由 `_deliver_lark_inbox_outbound` 持有：先验证 bot 与目标群，再按 outbound 中声明的 ID 类型查询当前群的 `users[]/bots[]`，随后 preview、send、mget，最终由 `lark_readback_matches_outbound` 对正文与 mention 做精确匹配，成功后才清理 reaction。正向路径把同群 `bots[]` 中的 `app_id ↔ member_id` 作为受约束 alias；未知 ID、真人 bucket alias、读取拒绝和多对一映射都 fail closed。这个所有权边界是合理的，未把 provider 身份猜测扩散到通用 matcher 之外。

### 具体改动

- 生产代码 33+/6-：`inbox_reply.py` 将 legacy 成员读取改成 `im +chat-members-list --member-types user,bot --page-all`，保留 outbound 声明的 ID 类型，并收集 bot alias；`outbound.py` 给回读 matcher 增加可选的 `verified_bot_aliases`，只在调用者提供验证结果时扩展 mention 身份集合。
- 测试/fixture 142+/13-：覆盖 user/bot bucket、读权限拒绝、未知 app ID、歧义 mapping、reply/chat-root 两个入口，并锁定命令参数。
- 文档 8+：说明 `im:chat.members:read` 已属于 canonical core/recommended scope，配置不等于 provider 授权。
- 活跃调用点已核对：CLI `loopx/cli_commands/lark_inbox.py` 与 `goal_topic_runtime.py` 都进入上述 delivery 边界；通用 reviewer notification 因新参数有默认值而保持原行为。没有 generated 或机械搬移。

### 对主干的风险

当前最强回归就是上面的 mixed-ID false negative；它不是未发送，而是“已发送后无法证明”，爆炸半径覆盖包含混合 mention 的 inbox/Goal Topic outbound。修复后应由同一个测试同时断言一次 send、app-ID 回读通过、reaction cleanup 完成。其余负向路径总体保守：成员读取失败、身份缺失和歧义仍在发送前或回读阶段失败关闭。

验证矩阵：122 个相关 Lark tests 通过；`lark-event-inbox-smoke.py` 通过；Ruff 与 `git diff --check` 通过；本地 mixed-ID 探针失败，证明 finding。required CI 的 shard 2 / aggregate pytest 仍红；其中两个失败测试在 PR exact head 与 base `8f72c21944b9d5ce9a130a7e962939e80c93c43e` 都可复现，因此不是这 6 个文件引入，但发布/合并门禁仍需先恢复。本 PR 不宣称 default-off，相关隔离项不适用；public authority 命名与实际 bot/member 语义一致；没有新增通用控制面 obligation 或领域化状态分类。残余风险是 provider 截断会造成保守拒绝，不会造成错误身份被接受。

### 我的整体评价

方向正向、范围基本成比例：以 33 行生产增量把群成员来源、bot bucket 与 readback alias 串成同一条可审计链，价值明显高于维护成本；文档和负向测试也扎实。但 exact head 目前仍遗漏一个由其自身“按 ID 类型逐次查询”产生的常见组合路径，且 required checks 为红，所以不能给出无阻塞结论。请先按类型保留 alias 关系并补 mixed-ID 回归；随后在新 exact head 重跑 focused tests、smoke 与 required CI，我再复审。

English verdict: **Request changes** — the approach is sound, but mixed mention ID kinds collapse one bot's typed member IDs into a false ambiguity after the message has already been sent; required CI is also still red (the current CI failures reproduce on the base).

now-ing:
Reviewed at head `c7900d89`. Scope: the Lark outbound mention identity verification switch and the bot app-ID readback alias mapping.

The direction here is right on every load-bearing point: the legacy `im chat.members get` endpoint genuinely could not establish bot absence, so moving verification to `im +chat-members-list --member-types user,bot --page-all` is the correct evidence upgrade, not a scope-grab — the new permission was already in `CORE_BOT_SCOPES` on `main`, and this PR only locks that fact with a regression test plus README clarification that the default bundle is not provider authorization. The fail-closed posture is preserved end to end: denied member reads return `gate_required` without any send call, app-IDs outside the verified bot bucket (or aliased from the `users` bucket) land in `sent_unverified`, and the alias is only accepted when it maps unambiguously to an identity the outbound text actually mentions. The 36-case test matrix (placement × denied × bucket × readback ID) plus the ambiguity test pins the contract well. Locally I ran `tests/test_lark_inbox_reactions.py` + `tests/test_lark_bot_scopes.py`: 60 passed.

Two findings before merge:

## 1. Confirming the author's self-review P1, with a scope note

I reproduced the reasoning in the author's own review and agree it is real: when one message mentions identities of two different `member-id-type`s (e.g. a bot via `open_id` and a human via `user_id`), the loop in `inbox_reply.py:285-317` issues two queries, and the same bot legitimately appears in both `bots[]` responses with kind-encoded IDs (`ou_…` and `u_…`). Both land in the single `bot_alias_candidates[app_id]` set, so the `len(identities) == 1` gate at `inbox_reply.py:485-490` discards the alias entirely — an already-delivered message then reads back as `sent_unverified`, and the recovery path treats a successful send as a failure. One addition to the author's analysis: the blast radius is bounded — messages whose mentions all declare the same identity kind (the overwhelmingly common case) take exactly one query and cannot hit the ambiguity, so this is a targeted fix, not a redesign. Since `expected_mentions` already knows each mentioned identity's kind, narrowing the alias source to the query for the declared kind (the author's option B) is the minimal correct fix, with the mixed-kind regression test the author proposed.

## 2. The two member-list fixtures disagree on the response shape, and the smoke never exercises the alias path (small)

The unit-test `ReplyRunner` serves `+chat-members-list` as `{"ok": true, "data": {users, bots}}` (`tests/extensions/test_lark_inbox_reactions.py:371-385`) while the smoke fixture serves a top-level `{"users": [], "bots": [...]}` with no envelope (`examples/lark-event-inbox-smoke.py:350-356`). `inbox_reply.py:302-304` accommodates both via `.get("data", member_payload)`, but the real CLI emits only one shape — so at least one fixture is not pinning reality, and nothing in CI can tell which. Related: the smoke's bot member carries no `app_id` field, which means the PR's headline behavior (readback reporting an app-ID, verified through the alias mapping) has zero end-to-end coverage — only the unit runner exercises it. Suggested follow-up inside this PR if cheap: align both fixtures to the live-probed shape, give the smoke bot an `app_id`, and make the smoke readback use it.

Suggested resolution: finding 1 is a genuine correctness gap in the new feature and worth fixing before merge; finding 2 is inexpensive hardening. Neither changes my overall read — the verification upgrade itself is sound and the failure modes all fail closed.

huangruiteng:
Resolved the mixed-identity P1 on `04cbd064c`.

The root cause was that bot aliases returned by separate `open_id` and `user_id` member queries were collapsed into one `app_id -> member_ids` set. A bot mentioned by `open_id` could therefore become falsely ambiguous when the same chat also contained a human `user_id` mention. Alias candidates are now admitted only when the returned bot member ID matches an expected mention declared for that exact identity kind. Genuine same-kind ambiguity still fails closed.

I also aligned the durable inbox smoke with the live `chat-members-list` response envelope and made it verify the `app_id` readback path end to end. The new regression covers a bot `open_id` plus a human `user_id` in one outbound message, including the provider returning both identity forms for the bot.

Validation:

- 174 focused Lark/periodic-report delivery tests passed.
- `lark-event-inbox-smoke.py` and `lark-goal-topic-connection-smoke.py` passed.
- Ruff passed on changed source/tests; the inbox smoke passed with its intentional `E402` bootstrap exception.
- repository-configured mypy and scoped mypy for the changed Lark modules passed.
- public/private boundary scan passed.
- exact-scope change-quality receipt `cqr_ea611f570d542200beb5` is valid for fingerprint `ea611f570d542200beb51e5797cedba47fb0fff46cd220bd581e4485efb0e90c`.
- `loopx canary premerge --from-git-diff --goal-id loopx-meta` passed 14/14 with zero failures, skips, or manual holds.

The prior CI shard failures were the upstream CLI extraction regressions fixed on current `main`; the new push will exercise this head against the current merge base. No permissions, scopes, configuration defaults, or provider write authority were broadened. The bounded future-facing review found no additional refactor worth coupling to this identity-boundary fix.

Issue #4012, opened 2026-09-06
[Bug]: Loopx control plane app is not working on macos 26.5/ M5


xiaods:
### Preflight

- [x] I searched existing issues and discussions for this behavior.
- [x] This report is not an unpatched security vulnerability.
- [x] The report and attachments contain no credentials or private/internal material.

### Issue origin

Observed or reproduced in a real environment

### LoopX version or commit

Loopx 0.5.4 and control plane v1.0

### Host or runtime surface

Other

### LoopX area

Control plane (goals, todos, quota, scheduler, registry, runtime)

### Problem

<img width="628" height="749" alt="Image" src="https://github.com/user-attachments/assets/33711a6f-b6d7-4bd9-a2ed-b3d44c37a966" />
the control plan app is loading and not working .

but pi host run loopx doctor show the loopx env is good.
<img width="1227" height="564" alt="Image" src="https://github.com/user-attachments/assets/ac2d6db2-ca56-4632-a624-6ecd43382737" />

### Minimal reproduction

1. install loopx app v1.0
2. run loopx app and the ui is always loading 
3. 

### Expected behavior

the control plan app working like a charm.

### Actual behavior

the loopx app is not working correctly

### Sanitized diagnostics

```shell

```

xiaods:
LGTM

PR #4014, opened 2026-09-06
fix(desktop): expose actionable recovery diagnostics

huangruiteng:
## Problem and change

Desktop recovery previously reduced several startup/installation failures to a generic message, leaving a healthy terminal CLI insufficient to diagnose App startup (refs #4012).

The existing recovery section now shows actionable bundle, installer, timeout and identity failure messages plus selectable/copyable diagnostics. Native status retains the last failure across subsequent update checks within the App process. The export includes only App version, failure category, installer exit code and identity availability/match booleans; raw logs, environment variables, paths and Goal content are excluded. Clipboard denial falls back to selecting the text. Default collapsed startup layout is unchanged.

Ownership stays in Desktop maintenance and its existing IPC; installation, version matching, permissions and the decision kernel are unchanged. The related simplification pass uses bounded structured evidence rather than a raw-log redaction framework.

## Validation

- Rust library tests: 25 passed; existing opt-in real installer and signed backup qualification tests remain excluded. No live installation or signed release was performed.
- Clippy (all targets, warnings denied) and Rust formatting passed. First Clippy attempt exhausted local disk; passed after cleaning regenerable Desktop package build artifacts.
- Two Node regression tests passed (privacy allowlist, retained failure, installer exit message, clipboard fallback).
- Production static recovery page inspected in a real browser, including unavailable native IPC diagnostics.
- Public boundary scan and diff whitespace check passed.

This improves diagnosis; it does not claim to resolve the reporter's installation failure. Existing installed Apps need a newly built Desktop artifact to receive this native change.

### DIFF
--- a/apps/desktop/loopx-control-plane/README.md
+++ b/apps/desktop/loopx-control-plane/README.md
@@ -54,6 +54,15 @@ matching runtime too. Goal state is neither deleted nor migrated backwards by
 this action, so data-schema compatibility still governs rollback suitability.
 Older backup directories are retained for manual recovery and can consume disk.
 
+The embedded Recovery & updates section includes selectable, copyable diagnostics
+with the App version, last failure category, installer exit code when available,
+and runtime identity availability/match results. The last failure survives a
+subsequent update check within the same App process. Copying never includes raw
+installer output, environment variables, local paths, or Goal content. Installation
+failures, missing/corrupt bundles, and runtime mismatch show distinct recovery
+instructions. A terminal `loopx doctor` checks the terminal-selected runtime;
+it does not prove that Desktop selected the same installation or matching revision.
+
 The updater accepts only fixed official HTTPS channels, not browser-provided
 commands, paths or download URLs. Its signing private key is confined to the
 release secret; the App embeds the public key. PR validation has no signing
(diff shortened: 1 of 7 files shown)


huangruiteng:
Author review of aedf491be962d492f9f37ac892f02e5aa32535b8: no actionable blockers found in the bounded diagnostics change.

The native maintenance owner retains a last failure without adding installation authority. The existing IPC gains an additive field; the recovery export selects known codes/booleans and validates the version rather than copying arbitrary diagnostics. Missing native IPC yields a useful diagnostic; clipboard failure leaves selectable text. Related simplification: reuse existing status/panel and avoid raw-log collection. Startup default viewport is unchanged.

Validation: 25 native Rust tests passed; Clippy all-targets with warnings denied, formatting, two Node regressions, production static browser inspection, public boundary scan and premerge 18/18 passed. Exact receipt cqr_e2b224d1162a167ab1d2 is valid. Premerge has no manual holds; native validation was run separately because its automatic surface selection is broad docs/public-boundary coverage. The two existing opt-in signed install/backup tests were not run; this change only reports installer exit status and does not modify installation or backup operations. Remote full Python shards and cross-platform Desktop builds may still be running; no signed release or user installation has been performed.

Owner explicitly authorized self-merge for this PR. This is a diagnostics improvement for #4012, not evidence that the reported machine's original failure is fixed.

PR #4074, opened 2026-09-07
test: align public smokes with shipped contracts

huangruiteng:
## Summary

- align the dashboard marker with the localized `personalRunEvidence(payload, row, t)` call
- replace stale pre-1.0 README assertions with the shipped bilingual LoopX 1.0 statements
- cover the typed `todo_replan_cadence` catalog contract, including default/current and preview/apply boundaries
- align the host-integration smoke with atomic hard-lease claim promotion while retaining quota, capability, write-scope, and workspace guards

This changes validation only; it does not change runtime behavior.

## Validation

- focused Python 3.11 execution of all four changed smokes: passed
- full-public shards at offsets 220, 330, and 440: 292/292 passed
- Ruff on all changed Python files: passed
- `git diff --check`: passed
- `loopx canary premerge --from-git-diff --goal-id loopx-meta`: 13/13 passed, no manual holds
- exact-scope change-quality receipt: `cqr_0bc07cd8405dca6a6c4f` (valid)

During validation, an initial concurrent multi-shard run exposed three environment-sensitive failures: a timing-budget overrun under contention, an install timeout, and the Apple Git hook-path behavior in the issue-fix fixture. The affected checks were rerun without cross-shard contention; the Git fixture used the bundled Git 2.53 runtime, and all three complete affected shards then passed. No checks are skipped.

## Boundary and scope

- no private state, credentials, raw logs, local paths, or generated evidence are included
- no first-screen product content changes; the README changes are assertions against already-shipped copy
- future-facing pass: the configuration smoke now covers the typed cadence contract rather than only admitting its feature id; no broader helper or abstraction is warranted for four separately owned contracts


huangruiteng:
No blocking findings.

Product/architecture judgment: this PR repairs four stale validation contracts against already-shipped behavior; it does not change runtime authority or presentation. The assertions now follow the real owners: the typed cadence catalog, atomic hard-lease claim promotion, the localized dashboard call signature, and current 1.0 README copy. Keeping the repair in those existing smokes is preferable to adding another compatibility layer.

Validation reviewed: all four focused smokes passed; the affected full-public offsets passed 292/292 without skips; Ruff, diff hygiene, public/private boundary, exact-scope change-quality (`cqr_0bc07cd8405dca6a6c4f`), premerge 13/13, both pytest shards, Windows, and Stage 2C installed/mutant/e2e checks are green. The current v1.0.1 release full-public run independently reproduced exactly these four stale assertions, so the fix directly clears a release blocker.

Merge decision: self-merge is appropriate under the narrow public-smoke cleanup policy. Scope is four test assertions, single-purpose, reversible, and carries no runtime, permission, private-state, benchmark-scoring, or evidence-policy change.

PR #4090, opened 2026-09-08
Fix direct homepage anchor links after React rendering

huangruiteng:
Opening the homepage directly at `?lang=zh#explore` left the viewport at the hero because the static shell did not contain the section when the browser resolved the fragment. React added it later without replaying fragment navigation; even `:target` remained unresolved on cold entry.

Resolve the URL fragment after the homepage mounts, align the matching element immediately, and keep its Reveal wrapper stationary and visible. Missing or malformed fragments leave normal page entry intact. Ordinary in-page links and history remain browser-owned.

The browser regression now starts from cold fragment URLs with delayed JavaScript, asserts the actual section/heading viewport before any scroll or focus helper, and covers reloads, both languages, desktop/mobile, motion preferences, encoded IDs, and invalid/missing fragments. The old test's manual scrolling no longer masks navigation failures. This requirement is also recorded in the presentation interaction baseline.

Validation: the new cold-entry assertion failed against the pre-fix build; all cases pass with the fix. Production homepage/dashboard export, strict documentation build, complete-site browser smoke, public share-bundle smoke, design boundary check, final changed-path boundary scan, and diff hygiene passed. Live reproduction before the fix recorded scrollY=0 with the section over 2,800px below the viewport; the fixed production preview lands at its top.

Scope/future-facing review: a bounded fix in the existing homepage and Reveal CSS, with no new abstraction, backend behavior, permission change, or scrolling listeners. The risk-based premerge set covers rendering, fragment timing, motion, URL input handling, and the public publication path; no manual holds remain.

### DIFF
--- a/apps/presentation/site/src/App.tsx
+++ b/apps/presentation/site/src/App.tsx
@@ -776,6 +776,21 @@ export function App() {
     window.history.replaceState({}, "", url);
   }, [language]);
 
+  useEffect(() => {
+    // The static shell has no section IDs. Replay initial fragment navigation
+    // once React has committed them; ordinary links/history stay browser-owned.
+    let id: string;
+    try {
+      id = decodeURIComponent(window.location.hash.slice(1));
+    } catch {
+      return; // Malformed fragments should leave the normal page entry intact.
+    }
+    const target = document.getElementById(id);
+    // :target may also be unresolved when the browser parsed an empty shell.
+    target?.closest(".reveal-block")?.setAttribute("data-anchor-entry", "");
+    target?.scrollIntoView({ behavior: "instant" });
+  }, []);
+
   async function copySetup(option: "agent" | "shell") {
     const text = option === "agent" ? setupPrompts[language] : shellSetupCommand;
     try {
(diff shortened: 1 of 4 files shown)


huangruiteng:
Reviewed 8897ef6d953ad6e20fc23bd43e4a887546d972a6: no blocking findings.

The fix belongs to initial homepage rendering, not the router or a new scrolling service. It replays a valid fragment once after React commits the section, keeps the destination stationary during Reveal effects, and safely ignores malformed or missing IDs. Native subsequent navigation is preserved. There are no backend, capability, permission, or optional-feature changes.

Risk-based premerge coverage passed: production export/typecheck, complete-site browser navigation with cold/delayed-JS entry and reloads across languages/viewports/motion settings, encoded and invalid fragments, ordinary navigation without test-forced scrolling, public share-bundle checks, strict docs build, design boundary and changed-path privacy scans, and diff hygiene. The regression failed on the base build before the fix. No required checks were skipped; no manual holds. This coverage exercises the actual entrypoint and the rendering/timing boundary affected by this small change.

The bounded future-facing pass avoids a new hook/module and records the deep-link testing requirement in the existing interaction baseline. Approved for owner-authorized self-merge after Pages validation.

PR #4170, opened 2026-09-10
fix(manager): use global Core evidence and independent model defaults

huangruiteng:
The manager previously inherited the first selected Goal's project context and had no authoritative Goal inventory, so asking which Goals it could see produced a project-only answer. It now has a canonical global identity and neutral workspace, reads fresh scoped Core evidence every conversational turn, and migrates legacy managed sessions without crossing Codex homes.

The read-only Goal Portfolio provider preserves recorded deliveries across quota noise and reports discovered, verified, stale, conflicting, unreadable and omitted sources. Full owner inventory collection shares one Core snapshot. External reads use an injected live authorization resolver: Lark resolves the current active connection, audience, session, executor and connector identity. The previous session Goal is never an authorization grant. Disabled, missing or ambiguous connections expose no Goal evidence, and scope changes during collection discard the snapshot.

The manager now also reads current canonical-first Todo details for every authorized Goal, preserving concrete titles, owner gate/action distinctions and affected-task relations. This read has its own provenance and coverage, separate from stale progress receipts. The neutral workspace receives default managed AGENTS.md instructions to explain decisions and prioritization from evidence; custom instructions are preserved. Owner continuation notes remain outside external-audience projections. Focused regression and HTTP smoke passed, including stale progress with fresh task records, missing-source handling and default-instruction delivery.

Frontend creation and automatic restoration no longer use the first Goal as an anchor. Normal manager questions no longer silently substitute a limited client projection. Codex manager sessions explicitly default to `gpt-6-astra` / `medium` at thread start, resume and turn start, with dedicated service overrides; worker settings are unchanged. Targetless Todo proposals cannot silently apply to the first visible Goal.

Validation: 108 focused Python, connection, Git fast-path and maintainability tests passed, plus HTTP smoke and lint. Earlier runtime smoke, TypeScript/build and packaged browser smoke passed; frontend behavior is unchanged by the authorization follow-up. Regression coverage includes a 45-Goal inventory, live authorization/revocation, scope-before-read and recheck-after-read, per-turn evidence receipts, legacy migration, and explicit model RPC parameters. Local App/CLI/services were paired and a real Codex turn verified the neutral workspace and effective model/effort. Read-only qualification of an existing Lark connection confirmed its evidence scope; no external test messages were sent.

CI: the stale manager-wording assertion is fixed. The Git `test_real_fetch_pull_quiet_and_new_commit_still_denied` failure also occurs on base f28b13cf3 ([base job](https://github.com/huangruiteng/loopx/actions/runs/34460441673/job/102817239282)); it passes locally. The test now retains captured Git diagnostics for CI investigation. This does not relax enforcement or claim a green merge gate.

Cross-entry request deduplication, optional Decision Context integration and live Lark reply/ACK qualification remain separate work. The scope resolver is a reusable control-plane boundary, not a second progress store or a grant of global visibility to every group.

After rebasing onto current `main`, the exact head passed a real Codex app-server manager Turn through an isolated synthetic registry: the global `loopx-manager` session completed, used fresh owner-global Core context, and produced an answer without mutating Goal state. No external Lark message was sent. The refine also binds external manager continuity to the exact authorized Goal set: missing authorization fails before a model call, and a changed scope starts a fresh upstream thread without prior model history.


huangruiteng:
Self-review/refine completed on exact head `aab9cb5cffef9944b379955bff8f5f9107cc4e69` after rebasing onto `7845a213ac9884c1fa95f6804e2702dce8285021`.

### Review finding fixed

The external authorization check originally bounded only the fresh evidence payload. If a manager binding were revoked or changed, the same upstream model thread could still retain evidence from the former Goal scope. The refine now:

- rejects empty/missing external authorization before any model call;
- persists an opaque identity for the exact authorized Goal set;
- starts a fresh upstream thread with no previous model history when that set changes;
- keeps the same upstream thread for later Turns in the unchanged scope.

Focused negative and retry tests cover revocation, scope change, and stable-scope continuation.

### Conflict resolution

- Rebased all five original commits onto current `main`.
- Preserved the mainline empty Git reference-transaction coverage while retaining the PR's subprocess diagnostics.
- Rebuilt the packaged Chat bundle from the merged dashboard source, preserving #4175's conversation readability behavior.

### Validation

- 119 focused Python tests passed.
- Ruff passed on all changed Python paths.
- Repository mypy configuration passed (21 source files); the changed manager context/store modules also passed scoped checking.
- Control-plane TypeScript typecheck passed.
- Dashboard Chat build, packaged personal-workspace browser smoke, Chat HTTP smoke, and PWA bundle smoke passed.
- A real Codex app-server Turn passed through an isolated synthetic global-manager entrypoint; no Goal state was mutated and no external Lark message was sent.
- Public/private boundary scan passed for all 23 changed files.
- `loopx canary premerge --from-git-diff --goal-id loopx-meta`: 18/18 passed, zero failures, zero manual holds.
- Change-quality receipt: `cqr_3cd8a024baa273db1c13`; exact diff fingerprint `3cd8a024baa273db1c13605cbb4b13b8455b68add1c3a2be8595437d9b24da8d`.

No unresolved code blocker remains in this self-review. I am not self-merging because this PR changes runtime behavior and an external authorization boundary, which the repository policy reserves for independent review. Requesting the relevant frontend/Lark CODEOWNER review; CI is also still pending on the rewritten head.

Future-facing pass: applied only the adjacent authorization-continuity repair above. Broader manager/request-ledger work remains outside this PR.

PR #4182, opened 2026-09-10
fix(blog): provide bilingual large-image social cards

huangruiteng:
Blog shares currently declare a summary card without an image. Add language-specific 1200 × 630 PNGs and large-image card metadata to the English and Chinese long-horizon-control article pages so social crawlers can render an illustrated article link.

The patch changes only static sharing metadata and two public assets. Article text, navigation, runtime behavior, and authorization are unchanged. The existing static-site public asset owner is sufficient; no capability or provider is introduced. A broader abstraction is unnecessary for these two pages.

Validation: production site build; metadata URL/dimension checks; decoded image dimensions; unchanged article-body comparison; visual inspection of both cards; diff and public/private boundary checks. Live crawler caches remain provider-controlled and require verification after deployment.


huangruiteng:
Self-review: no blocking findings in the four-file diff.

Changed surfaces: two static article head sections and two public PNG assets. The article body is byte-equivalent after parsing; no visible site navigation or runtime contract changes. Both cards were visually inspected at their native 1200 × 630 size.

Validation passed: clean npm install and production site build (TypeScript + Vite), both declared image URLs mapped to shipped assets, PNG dimensions decoded, matching Open Graph/Twitter image references, large-image card declaration, unchanged article bodies, git diff whitespace and public/private boundary scans. No local validation failures or skips. CI and production deployment are reported separately.

This risk-based set covers the static metadata and asset delivery without exercising unrelated runtime/benchmark suites. The existing public/blog asset boundary is sufficient; no companion refactor is needed. Residual risk is external crawler caching and X selecting another URL when several links are pasted, so the final composer must display the intended article card before handoff. Owner authorized self-merge.

PR #4205, opened 2026-09-10
fix(lark): surface unexpected event-source disconnects

huangruiteng:
A misconfigured Feishu/Lark profile can authenticate and read message history while its event WebSocket fails. The local bus registers the consumer, then shuts it down with `reason: signal` and exit code zero. LoopX previously treated that sequence as a normal stream ending, hiding the distinction between a planned restart and a disconnected source.

Treat an exit as planned only for LoopX's own stop or the requested timeout/limit. Preserve a content-free `lark_event_source_disconnected` diagnostic through the listener health projection and show actionable bilingual guidance in Lark settings. Correct the registration example that prescribed international Lark for a Feishu setup.

Validation: 121 focused Lark runtime, routing, and HTTP contract tests; Chat server HTTP smoke; dashboard TypeScript check and both production builds; Ruff and diff checks. Tests cover zero-code signal, unknown/missing reason, nonzero exit, timeout/limit, and owner-requested shutdown. Build emits the existing large-chunk advisory.

Risk and limits: provider versions omitting exit markers are conservatively reported as disconnected. The local ready marker still proves consumer registration only. This does not automatically change profiles, grant scopes, or replay historical messages. A real affected installation was repaired and its exact pending message recovered with the original idempotency key and provider reply readback; private configuration and message evidence are excluded from this branch.


huangruiteng:
Self-review: no blocking findings in this bounded change.

The provider owns interpretation of CLI exit markers; core Goal/Todo state and authorization are unchanged. The UI consumes a content-free error code. This fixes misleading restart classification and the platform-selection example; the affected local profile was repaired separately. It does not claim that a local ready marker proves the upstream connection or that historical replay is automatic.

Risk-based premerge coverage: 121 Lark stream/routing/HTTP contract tests passed; Chat server HTTP smoke passed; dashboard typecheck and both production builds passed; Ruff and diff checks passed; public/private scan passed. Fixtures cover planned bounds, explicit shutdown, zero-code signal, absent/unknown exit reason, and nonzero exit. The bundled Chat asset was rebuilt from the reviewed source. Full repository CI is still running; no reported failing check at review time. No new permissions or message replay policy is introduced.

Owner explicitly authorized self-merging the manager/Lark improvements. Merge only this reviewed head after checking the current PR comments and checks again.

PR #4213, opened 2026-09-11
fix(manager): retain concrete Core findings in progress reports

huangruiteng:
Manager reports could list delivery identities and follow-up plans while omitting concrete findings already stored in Core. Supply bounded checkpoint explanations, observed reality, result/probe fields and opaque evidence lineage from the existing authorized run-index read; update the default manager instructions to explain those findings with their verification limits.

The change stays in the built-in manager projection. It adds no artifact reader, filesystem access, new capability, or authority store. Existing scope/revocation and snapshot boundaries remain in place. Missing, invalid and truncated fields remain explicit; recorded claims never imply independent artifact verification. Existing managed sessions migrate to the updated instructions.

Validation: 64 focused tests covering manager context/history, Todo details, portfolio, delegation, Lark API and Codex-home isolation; Chat HTTP and runtime/restart smokes; Ruff and diff checks; four-path public-boundary scan. Read-only integration against three registered Goals confirmed concrete findings are present while artifacts remain marked unread. The contract check passed with three pre-existing registry/history warnings outside this diff.

This is a risk-based premerge set for the changed presentation, session migration and scope surfaces. No trading or state-transition semantics change. Live installed-model and fresh external-entry qualification remain rollout checks. The adjacent refactor is intentionally small: one private projection helper reuses the existing scan rather than adding a parallel receipt store or reader.

### DIFF
--- a/docs/reference/protocols/manager-evidence-and-continuity-v0.md
+++ b/docs/reference/protocols/manager-evidence-and-continuity-v0.md
@@ -92,6 +92,18 @@ unknown. Collection is capped at 128 Goals and eight agents per Goal, with
 explicit omissions. Recorded evidence references identify Core receipts, not
 independent artifact verification.
 
+Each dated delivery now includes bounded `recorded_details` from the same Core
+run index: the Agent's checkpoint explanation, observed reality, path outcome,
+result class, probe kind and surface identity. Missing, malformed and truncated
+fields are explicit. Evidence identifiers are hashed into stable lineage refs,
+with included/omitted counts; they are not artifact access capabilities.
+The manager should explain these concrete recorded findings and counterevidence,
+joined to the task title, rather than return only receipt IDs and future plans.
+These facts retain `recorded_claim_not_independent_verification` and
+`artifact_read_status=not_read`. No repository, arbitrary path, URL or transcript
+reader is added. Existing owner/external scope checks and snapshot identity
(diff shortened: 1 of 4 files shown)


huangruiteng:
Reviewed head `7f08a9c313b763dc078f24f4564b97966c63c39e`: no blocking findings.

The defect is at the manager presentation boundary: concrete findings were already present in the authorized Core index but discarded before model invocation. This patch retains bounded typed fields and updates the shipped manager instructions. It does not add another authority store, artifact reader, new capability, or implicit access grant. The small companion refactor is a private projection helper at the existing read boundary; a general filesystem/receipt service would be unnecessary here.

Validation: 64 focused tests, Chat HTTP and runtime/restart smokes, Ruff, diff check and the four-file public-boundary scan passed. A read-only real-source integration confirms the fields reach three registered Goal projections without claiming artifact verification. Scope/revocation, dated-history filtering, accounting exclusion, session migration and Codex-home isolation are included in the test set. No failures in this set; the contract scan has three unrelated pre-existing registry/history warnings. DCO, dependency review, package build and both Node compatibility checks have passed. Broader CI jobs are still pending, not claimed as passed.

Under the owner's standing authorization for manager fixes, self-merge using this risk-based validation coverage. No manual hold on the code change; installed-model and fresh external-entry acceptance remain explicit rollout work and are not proven by CI.

PR #4220, opened 2026-09-11
fix(lark): deliver full manager reports within provider byte limits

huangruiteng:
A manager can finish a report while its Lark source remains unanswered: the runtime accepted up to 6,000 characters, but the shared sender rejected replies above 1,200 characters and the exception escaped the terminal-notice path.

Source-bound replies now retain the complete answer and use the documented 150 KB text-request budget, measured as UTF-8 JSON bytes including the provider-preview envelope. Standalone notifications retain their existing compact default. Local format/size failures produce a verified diagnostic reply; source ACK still requires delivery readback, and no model request is replayed.

Validation: 192 focused Lark runtime, connection, reaction and byte-budget tests; inbox smoke; Ruff and diff checks. Provider dry-run preserved 6,001 and 40,000 CJK characters. An actual report above the former limit was sent and read back exactly; large dry-runs are not server acceptance tests. Standard premerge results will be recorded before merge.

The existing bundled Lark outbound adapter owns this change; no new capability, authorization grant or state store. Beyond the provider limit, the persisted full answer remains available in Chat and Lark receives a diagnostic instead of truncated text.

### DIFF
--- a/loopx/extensions/lark/goal_topic_runtime.py
+++ b/loopx/extensions/lark/goal_topic_runtime.py
@@ -33,6 +33,7 @@ from .goal_channel_contracts import LarkTopicEventDecisionReason, bindings_for_g
 from .goal_channel_targets import goal_channel_target_for_name
 from .goal_topic_connections import decide_lark_topic_event
 from .inbox_reply import CommandRunner, reply_lark_event_inbox
+from .outbound import LarkOutboundTextError
 from .inbox_reactions import (
     _create_reaction, _delete_reaction, ensure_lark_event_inbox_received_reaction,
 )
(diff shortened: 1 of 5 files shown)


huangruiteng:
No blocking findings in the reviewed head 6625cb38e69f95f7f48c0a7f833bab83cf302bb2.

The runtime previously accepted a report that the sender could never deliver. This change removes the conflicting reply character limits and preserves the report, while retaining the existing identity, membership, preview, readback and source-ACK gates. Format rejection is a delivery failure with a verified diagnostic, not a successful answer. The typed ValueError subclass preserves CLI compatibility and keeps raw exception details out of the group.

Validation: 192 focused tests, a further 50-test final runtime/byte-boundary pass, inbox smoke, Ruff, and standard premerge all passed. Standard premerge covered the maintainability canary and eight extension/Lark risk-profile smokes plus compile/diff checks. Provider previews retained long CJK content; an actual over-1,200-character answer was delivered and read back intact. The near-limit payload probes were dry-runs only. Required Sign-off, build, dependency review and Node compatibility checks passed. Other hosted suites remain in progress; release publishing jobs are intentionally skipped for PRs.

Placement/future-facing review: the bundled Lark outbound module remains the shared owner; a typed format exception now connects its failures to the existing manager receipt path. No new capability or parallel state store, permission widening, model retry, or finance-specific behavior. Standalone notifications retain their existing limit. Content above the actual transport budget is preserved in Chat with an explicit group diagnostic; multipart delivery is outside this fix.

Merge decision: self-merge under existing owner authorization for manager/Lark repairs and today's gate bypass, after the above local validation. Remaining hosted checks are recorded as pending, not represented as passed.

PR #4227, opened 2026-09-11
Return delegated outcomes automatically to the original manager conversation

huangruiteng:
A manager delegation previously stopped at delivery or adoption, leaving the user to ask a second question to discover the outcome. This change makes the original exchange include an automatic receiver-authored conclusion.

- Persist the exact originating conversation and retain the worker's inbox item after adoption until it publishes a conclusion. Workers keep ownership of relevance, planning and priority; optional intermediate updates remain separate from private decision reasons and Core progress.
- Deliver queued conclusions through the existing Chat service and Lark inbox transport, with original-source routing, live authorization checks, deterministic deduplication, retry/readback and explicit ambiguous-write handling. No model polling or new automation is introduced.
- Receive follow-up transcript messages automatically in an open manager frontend. Update the packaged bundle and built-in operating contract; older requests recover a return route only from an exact existing Chat receipt.

Validation: 90 focused Python tests; dashboard typecheck/build; full personal-workspace browser smoke in development and packaged modes, including automatic conclusion arrival, no new model turn, deduplication, desktop and mobile views. Risk-based premerge validation is recorded in the review comment before merge.

The receiver still needs a normal work turn to process the request. A planning decision can conclude a direction-setting request; an implementation request requires a concrete result or explicit blocker. Ambiguous external writes are held for verification rather than blindly resent. Runtime records, live provider data and local validation artifacts are excluded.


huangruiteng:
No blocking findings remain in the delegated-return path after the implementation review.

Product/architecture: this closes the original request instead of adding another user-driven status query. Request/decision truth stays in manager-context and work truth stays in Core; the new module owns only exact return routing, audience-ready replies and transport receipts. The worker retains planning autonomy. The existing Chat service, Lark inbox send/readback path and frontend transcript are reused. A separate business automation or copied task-progress store would be less coherent.

Reviewed risks: original-source targeting, sender/recipient grant revocation before sends, source ACK ordering, private decision text versus audience-ready text, restart/concurrent deduplication, transport outages and ambiguous writes. Terminal replies are immutable; longer work may publish one meaningful intermediate update. Existing legacy data is not bulk-replayed.

Validation: 90 targeted Python tests; actual Chat server smoke; typecheck/build; full development and packaged personal-workspace browser smokes, including automatic arrival without a second prompt/model turn and mobile/desktop rendering. The first development browser attempt timed out at startup on a shared port; isolated development and packaged runs both completed successfully. Privacy and diff scans passed.

Premerge: the initial maintainability finding was fixed by moving service composition out of chat_server into the return transport module. The rerun's remaining hot-path failure is reproduced identically on untouched base b57c61502: heartbeat_prompt_json is 3825 characters against 3600 after the separately merged heartbeat change. No budget was weakened here. The final premerge run selected 18 checks: 17 passed and this one base-reproduced failure remained; direct diff/compile checks passed and no tracked side effects occurred. A final transport receipt field-name correction was independently verified with all 9 roundtrip/Lark return tests.

Future-facing refactor pass: no new model polling, scheduler, parallel progress authority, arbitrary recipient selector or persistent permission escalation. Public changes cover runtime/API, frontend bundle, built-in contract and durable regression tests; personal runtime state and provider data remain excluded. Existing owner authorization covers manager/Lark self-merge and deployment.

Merge decision: self-merge under the owner’s explicit manager/Lark authorization, with the independently reproduced upstream heartbeat budget failure disclosed rather than presented as green.

PR #4228, opened 2026-09-11
fix(update): harden pinned installs and separate prompt readiness

huangruiteng:
## Summary

- Pinned full-SHA archive installs bypass unnecessary GitHub commit resolution. Symbolic refs retain exact resolution, with bounded authenticated `gh` fallback and no alternate-ref fallback.
- After successful binary installation and core doctor, reconcile managed prompts even if an optional extension doctor fails. Preserve the extension failure and original repair action.
- Report `upgrade_complete` separately from the existing runtime `ok` result; retain blocked-entry reasons and require review for custom prompts or alternate loader bindings.
- Document first-upgrade limitations for older CLIs without the migration hook. No fuzzy prompt adoption, scheduler mutation, session migration, or new capability/provider.

## Validation

- Tested revision: 934b2db44
- Run state: finished
- Input classes: synthetic

| Check kind | Result | Public-safe evidence / limitation |
| --- | --- | --- |
| `unit` / `integration` | `passed` | 85 passed: `tests/test_self_update_runtime_activation.py`, `tests/test_self_update_download.py`, `tests/test_archive_installer_commit_response.py`, `tests/control_plane/test_automation_prompt_upgrade.py`. One native-Windows-only test skipped on macOS. |
| `real_entrypoint` | `passed` | Shipped shell installer executed with real archive extraction and synthetic transport/installation target; new-runtime Python CLI exercised against isolated real SQLite/TOML stores. |
| `regression_parity` | `passed` | Before the fix, full-SHA installs still queried the failing API, and non-hex 40-character responses were accepted (4 new regression cases failed). After the fix, pinned installs make only the archive request; symbolic auth fallback and failure cases are covered. |
| `static` | `passed` | Ruff on changed Python files, `bash -n scripts/install-from-github.sh`, diff whitespace and public/private scans. |

Coverage: exact SHA, public/authenticated/unavailable lookup, malformed SHA, temporary-file cleanup, core-install/doctor failures, optional-extension failure, custom/canary prompt non-adoption, replay, and SQLite/TOML preservation. No live model calls: this change does not alter inner prompt or Todo/quota semantics. Full hosted CI and post-merge installation are reported separately; tests do not simulate the whole GitHub service or prove long-horizon model uplift.

## Design / risk

The existing installer owns transport; the existing heartbeat upgrade boundary owns prompt readiness. No new abstraction or second migration engine. The default change is deliberate: optional extension failure no longer suppresses otherwise safe prompt reconciliation, but still prevents a fully healthy upgrade result. Exact-template ownership, same-home checks, compare-and-swap, and schedule/thread boundaries remain unchanged.

## Boundary checklist

- [x] No private prompts, session data, local paths, credentials or raw logs in this diff/body.
- [x] Dedicated branch based on current main; one cohesive fix with focused regressions.
- [x] DCO sign-off.
- [x] Owner authorized self-merge and a task-scoped local Git-window bypass after validation.

Shared-authority RFC fixture impact: N/A; no authority-store, Todo, projection-routing or promotion change.

### DIFF
--- a/docs/reference/automation-prompt-upgrades.md
+++ b/docs/reference/automation-prompt-upgrades.md
@@ -35,14 +35,32 @@ not Todo, quota or scheduler authority.
 
 ## Automatic upgrade and manual adoption
 
-`loopx update --apply` now captures an owner-only snapshot **before** replacing
+`loopx update apply` now captures an owner-only snapshot **before** replacing
 the runtime and invokes `automation-prompts sync-installed` in the **new**
 runtime afterward. Exact recognized managed bootstraps and byte-identical
 prompts reproduced by the old installed generator may migrate automatically.
 An automation name, matching prose or Goal id alone never authorizes adoption.
 Custom instructions remain `review_required`; a canary executable, different
 registry/home or changed preview is not silently retargeted. Binary-install
-success and prompt-migration success are reported separately.
+success and prompt-migration success are reported separately. `upgrade_complete`
+is true only when runtime qualification and prompt reconciliation both succeed;
+the existing `ok` field retains its runtime-result meaning. A successful install
+and core doctor still run prompt reconciliation when an optional extension
+fails its doctor, without clearing that extension failure or its repair action.
+Failed installation or core doctor never starts prompt writes.
+
+Upgrading from a CLI that predates this hook cannot retroactively capture its
+old template evidence. After installation, run the new `automation-prompts plan`
+and review/adopt the selected tasks through the App. Never interpret absence of
+a migration report as proof of migration, or loosen exact-template ownership
+to make a historical custom prompt appear automatically eligible.
+
+Archive updates pinned with `--ref <full-commit-SHA>` download that exact archive
+without a GitHub commit-API lookup. Symbolic refs still require resolution to a
+full hexadecimal SHA. If the public API fails, the installer can reuse existing
+`gh` authentication for the same GitHub repository/ref under a bounded timeout;
+it neither logs in nor selects another ref. If both routes fail, retry with an
+independently verified full SHA. Downloads have bounded timeouts and retries.
 
 **Exact managed v1 wrappers upgrade to v2 automatically through this path;
 they do not require per-task approval.** `automation-prompts plan` is only a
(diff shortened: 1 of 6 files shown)


huangruiteng:
Self-review of 934b2db44: this is a bounded installer/prompt-reconciliation fix, not a prompt-policy or scheduler change. Full hexadecimal SHAs bypass mutable-ref lookup; symbolic refs retain bounded resolution and may reuse existing GitHub CLI authentication for the same repository/ref. No credentials are requested or printed, and no alternate ref is selected. Prompt reconciliation now depends on successful installation and core doctor, while optional extension failures remain visible and keep aggregate upgrade_complete=false. Custom prompt/loader adoption still requires review; schedules, models, bindings and notification preferences are untouched.

Validation on the final head: 85 focused tests passed, 1 native-Windows-only case skipped on macOS; Ruff, bash syntax and diff whitespace checks passed. Coverage includes the real shell installer entrypoint with synthetic transports, archive extraction/install boundaries, update activation/download regressions, and new-runtime CLI reconciliation against isolated SQLite/TOML fixtures. Four new archive regressions were observed failing before the shell fix. Public/private scans are clean. PostgreSQL and model-behavior qualification are not applicable to these changed surfaces; this does not claim long-horizon model uplift.

The related refactor separates installation/core readiness from optional extension qualification without adding another prompt authority. Hosted CI currently has no failures but some jobs are still running. Real local/SSH installation and App/runtime identity readback will follow merge; existing unrelated extension failures will be reported rather than hidden.

huangruiteng:
Post-merge installation qualification at 081b1142df89: both macOS and Linux/SSH installed the exact same GitHub archive (matching archive digest), with install=0 and core doctor=0. The installer itself was pinned to this reviewed commit while Pages deployment was still propagating. Both installer downloads returned HTTP 200 on the first attempt, and full-SHA installation did not perform commit-API resolution. The macOS App embedded source, CLI release manifest and live chat-service runtime identity agree; the unchanged native shell was locally repackaged, not published as a new signed release.

Prompt readback: eligible managed entries remain current, unrelated unmanaged entries remain untouched, and an inconsistent paused binding remains explicitly blocked. On Linux, pre-existing optional extension failures preserve the aggregate unhealthy result, but prompt discovery now completes independently and reports not_installed for that host instead of skipped_runtime_update_failed. No schedule, notification, model or session rebinding was performed by this fix.

Hosted checks completed so far pass (including installed qualification, real CLI E2E, mutation checks and Windows); the two full Python shards are still running. This installation evidence is not a claim about model long-horizon uplift.

PR #4230, opened 2026-09-11
Read remote manager progress from scoped SSH Core evidence

huangruiteng:
Remote progress questions previously inspected only the local registry, then substituted local tasks mentioning SSH when Goal host metadata was empty. The manager can now select a real SSH evidence source and read its Goal portfolio, current Todos and recorded outcomes through the same Core providers used locally.

Configured SSH aliases reuse the frontend host catalog. Owner-local reads are available on demand; group reads require persistent exact host/Goal summary grants and the existing live connection authorization. Grants are checked around reads and included in upstream context identity. SSH executes only a fixed bounded CLI projection against the remote global registry; the model cannot choose commands or paths. Failures and older unsupported runtimes remain explicit unknowns. Source host qualifies Goal identity independently of optional declared execution-host metadata.

Delivery reads support an explicit 1–90 day lookback so a progress report can explain older recorded outcomes alongside freshly read Todos. Dates and artifact-verification limits remain visible. The built-in manager skill and context version are updated; no new automation or parallel progress store is introduced.

Validation: 102 focused tests including real CLI export, unauthorized-source rejection, revocation during reads, source provenance, old/outage behavior and historical dates; actual Chat server smoke. Premerge results and live remote acceptance will be recorded separately. Public diff contains only runtime, CLI, built-in instructions and durable tests; host identities, audience grants and live reports remain private.

### DIFF
--- a/loopx/capabilities/manager_context/README.md
+++ b/loopx/capabilities/manager_context/README.md
@@ -91,12 +91,41 @@ recent delivery window is still yesterday through now; arbitrary artifact paths
 and external links are not fetched. Existing non-Codex adapters retain their
 context projection until they implement an equivalent tool contract.
 
-Manager context version 8 starts a fresh upstream session for older manager
+Manager context version 9 starts a fresh upstream session for older manager
 contexts. The logical Chat session and its receipts remain intact. Runtime support
 uses the Codex app-server dynamic tool protocol; explicit upstream terminal
 errors remain errors and are not retried as part of inspection. The version
 change refreshes the operating contract on existing installations;
 resuming an old upstream thread would retain its previous instructions.
+
+### Remote evidence sources
+
+The manager discovers SSH aliases through the same host catalog as the frontend
+source switcher. `loopx_manager_read view=sources` lists eligible sources; select
+`source_id=ssh:<alias>` for portfolio, Todo or delivery reads. Reads execute a
+fixed, bounded CLI projection on the selected host, using its global registry,
+not local tasks whose titles mention SSH. Source host and Goal ID jointly identify
+the evidence; a missing declared execution `host_id` does not erase source provenance.
+
+Owner-local conversations may inspect configured hosts on demand. External
+conversations require a persistent, exact host/Goal read grant from the local
+operator, in addition to their live connection authorization:
+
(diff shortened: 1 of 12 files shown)


huangruiteng:
No blocking findings after reviewing the complete branch diff and permission boundaries.

The product defect is missing remote evidence, not insufficient prose: local SSH-related tasks cannot establish remote Goal progress. This change reuses the configured SSH catalog and existing Core projections, with source-qualified identity and a bounded read protocol. Exact audience grants are checked before and after reads and included in model context identity. It adds neither an arbitrary shell tool nor a second progress store. Older remote installations explicitly require upgrade. Live model/report acceptance remains a deployment check and is not claimed by unit tests.

Validation: 102 focused tests, actual Chat server smoke, and 18/18 selected premerge checks passed. The final source-pagination/protocol tightening was followed by all 102 focused tests, Ruff and diff hygiene. Rebase onto the installer/prompt-readiness change in current main had no overlapping implementation paths; 32 source/inspection/context tests passed again afterward. Public/private diff scan is clean. No skipped selected check or manual hold remains.

Self-merge under the owner's explicit authorization for manager/Lark fixes. Then deploy both ends and validate the original remote-report question against actual remote records.

huangruiteng:
Live follow-through for #4230 and #4231 is complete. The installed manager selected the configured SSH source itself, paginated the remote portfolio, read current tasks and dated delivery records, and exposed unavailable records instead of claiming full health. CLI, running Chat and desktop bundled runtime were upgraded alongside the remote CLI; installation checks passed.

The first real answer exposed a separate error: a fresh task read was presented as proof that an old external approval was still needed. That answer was not accepted as a successful report. #4231 corrected the default reasoning contract; a focused real-manager follow-up now explicitly withholds unsupported owner actions and calls for Agent reconciliation. An independently checked correction and useful progress report were automatically returned through the original delegation, with transport readback and exactly one Chat return verified.

These checks establish source access, the corrected recommendation behavior and automatic return for the tested exchange. They do not claim that stale remote tasks were reconciled, missing records recovered, or all referenced artifacts independently verified. Live identities, records and receipts remain private.

PR #4231, opened 2026-09-11
Verify external waiting claims before recommending owner action

huangruiteng:
The manager could read a current Core Todo correctly yet tell the owner to approve an external change that had already completed. A fresh read verifies the stored task, not the current PR, deployment or access condition described inside it.

The default manager instructions now require current authoritative evidence before converting an old external waiting claim into an owner action. Missing or conflicting evidence produces an explicit reconciliation task for the Agent. Context version 10 refreshes existing manager sessions so the new rule takes effect. This preserves Core task truth without inventing live external state or rewriting task priorities.

Validation: existing manager context, detail, report and SSH evidence tests; Ruff and diff hygiene. A focused real-manager acceptance follows deployment. No new tool or permission is introduced.

### DIFF
--- a/loopx/capabilities/manager_context/README.md
+++ b/loopx/capabilities/manager_context/README.md
@@ -91,7 +91,7 @@ recent delivery window is still yesterday through now; arbitrary artifact paths
 and external links are not fetched. Existing non-Codex adapters retain their
 context projection until they implement an equivalent tool contract.
 
-Manager context version 9 starts a fresh upstream session for older manager
+Manager context version 10 starts a fresh upstream session for older manager
 contexts. The logical Chat session and its receipts remain intact. Runtime support
 uses the Codex app-server dynamic tool protocol; explicit upstream terminal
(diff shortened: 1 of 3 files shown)


huangruiteng:
No blocking findings in the three-file diff. The live manager test exposed a semantic freshness error: it treated an open task about a past external decision as proof the owner still needed to act. Core remains authoritative for its stored task state, while current external state needs separate evidence.

The fix belongs in the default manager reasoning contract and bundled skill; no new state store, external tool, access grant or task mutation is needed. Existing sessions refresh through the context-version mechanism. Unknown dependencies are explicitly left for Agent reconciliation instead of being silently marked complete or surfaced as verified owner obligations.

Risk-based validation for this instruction-only follow-up: 31 context/detail/report/SSH tests, Ruff and diff hygiene passed. No executable provider behavior changed; the preceding source integration already passed its broader premerge selection. A focused live manager check after deployment will verify that an old external waiting task no longer causes an unsupported approval request. Self-merge is owner-authorized.

PR #4232, opened 2026-09-11
fix(ui): manage remote SSH source lifecycle

huangruiteng:
## Summary

- make configured remote SSH status sources explicitly pausable from the personal workspace
- keep tunnel ownership with a dedicated, user-private OpenSSH control socket so pause targets only the tunnel LoopX started
- persist the validated SSH Host alias and safely upgrade legacy configured-source rows when their labels exactly match a configured alias
- keep manual or pre-existing unmanaged tunnels fail-closed instead of using process-name matching
- keep SSH route registration inside the SSH API mixin rather than growing the Chat server composition module

## Root cause

The previous ensure flow launched `ssh -f -N -L` and immediately lost the process lifecycle identity. The browser catalog also persisted only a display label and status URL. As a result, the UI could delete a catalog row but had no safe way to stop or resume the exact tunnel.

This change carries the validated Host alias through the catalog, starts a dedicated OpenSSH control master, and pauses only through its exact control socket. On macOS the socket uses a short private `/tmp/loopx-ssh-<uid>` directory so OpenSSH's temporary suffix stays within the Unix socket path limit.

## Validation

- `ruff check` on changed Python runtime, API, and tests
- scoped `mypy --follow-imports=skip`
- 25 focused pytest cases for tunnel lifecycle, loopback API, route ownership, and CORS
- Dashboard TypeScript check plus desktop and packaged Chat production builds
- typed status-source catalog smoke, including legacy alias migration
- development and packaged browser smokes for pause, local fallback, catalog retention, stale-request fencing, and resume
- disposable-port real SSH lifecycle: ensure/read 200, pause/closed, resume/read 200, pause again
- control-plane maintainability ratchet
- public/private boundary scan
- LoopX premerge: 18/18 selected checks passed, no manual holds
- exact-scope change-quality receipt: `cqr_b3a67cfc7efeb85c5adc`

## Future-facing pass

Applied narrowly: SSH lifecycle routes are now registered by their owning mixin, and legacy catalog binding reads the latest catalog revision so async discovery cannot overwrite a concurrently added source. No new capability, provider, daemon, or generic lifecycle framework was added.


huangruiteng:
Self-review completed on `fa773193892610cf0f4fe5a08374f25097853e95`.

No actionable findings remain.

Reviewed boundaries:

- lifecycle ownership: pause uses one LoopX-created OpenSSH control socket; no PID scan, substring match, or broad process kill
- authority: the API remains loopback/origin guarded and the tunneled control plane remains read-only
- compatibility: legacy configured rows are upgraded only when the stored label exactly matches a currently configured safe Host alias; manual/unowned listeners fail closed
- concurrency: async Host discovery binds against the latest catalog revision and cannot overwrite a concurrently added source
- composition: SSH API routes are owned by the SSH mixin, avoiding new `chat_server.py` maintainability debt
- UX: pause returns to local without deleting the reusable source; selecting it again resumes the managed tunnel

Evidence: 25 focused pytest cases; Ruff; scoped mypy; catalog smoke; development and packaged browser smokes; full Dashboard production build; disposable-port real SSH pause/resume cycle; maintainability ratchet; public/private scan; LoopX premerge 18/18. Exact-scope receipt: `cqr_b3a67cfc7efeb85c5adc`.

PR #4235, opened 2026-09-11
fix(lark): keep manager received ACK visible after its answer

huangruiteng:
Manager receipt reactions disappeared when the answer arrived because reply cleanup deleted both received and processing reactions. Manager routes now retain their received ACK, so users can still see that their message entered the processing chain after the answer.

The Lark extension owns an explicit `received_reaction_policy`: generated manager routes select `retain`, while ordinary Inboxes and bound Goal routes preserve the `transient` default. The existing private receipt ledger provides restart/replay idempotency. Completion still clears processing indicators; a retained ACK does not claim task completion or grant authority. Existing historical reactions are not recreated automatically.

Validation: 143 tests across manager routing, reaction lifecycle, turn-start sync and reply sizing passed; Lark event Inbox smoke and Ruff passed. Regression coverage checks retention through processing/completion/restart, replay without provider calls, generic-default cleanup and invalid policy rejection. Live provider create/readback also verified the received reaction using the configured bot identity. Installed completion/replay acceptance follows deployment.

### DIFF
--- a/loopx/extensions/lark/docs/lark-event-inbox.md
+++ b/loopx/extensions/lark/docs/lark-event-inbox.md
@@ -255,9 +255,10 @@ messages; use `configured_chat_all` for complete collaboration threads:
 For every reply-enabled Inbox, a missing `reply.received_reaction_emoji`
 defaults to `Get`. Set it explicitly to the empty string to disable this
 provider write. The reaction belongs to the same explicit sender boundary as
-source-thread replies, but only the Agent's turn-start hook may create it:
-realtime collection persists events without reacting, and the hook writes the
-reaction only after it has read and confirmed a still-pending human message.
+source-thread replies. The Agent's turn-start hook creates it after reading and
+confirming a still-pending human message; the synchronous manager route creates
+it immediately before invoking the manager. Realtime collection alone persists
+events without reacting.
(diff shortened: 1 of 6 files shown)


huangruiteng:
No blocking findings after reviewing the full diff and regression results.

Product/architecture: the defect conflated a durable received acknowledgement with transient processing state. The fix stays inside the existing Lark extension, using its config normalization and reaction ledger; no new control-plane authority, scheduler, or provider framework is needed. The enum governs both processing and completion transitions. Manager retention is an intentional documented default change; generic Inbox and bound Goal behavior remain transient. Only this bot's recorded processing reaction is deleted, and existing sender checks and replay fences remain intact. The adjacent-boundary refactor pass found no additional extraction warranted for these two small transition changes.

Changed surfaces: normalized Inbox config, generated manager routing, received/processing completion lifecycle and operator documentation. Equivalent risk-based validation: 143 focused tests (manager routing, reactions, restart/replay, turn-start sync and reply size), the Lark event Inbox smoke, Ruff, diff hygiene and public/private diff scan all passed. Live reaction creation and readback were verified under the configured bot profile. No failures or manual holds; installed lifecycle verification is scheduled after deployment. No permission, model, or task-state changes are included.

Merge decision: self-merge under the owner's explicit authorization for manager/Lark fixes.

PR #4236, opened 2026-09-11
fix(lark): preserve manager ACK on delayed worker conclusions

huangruiteng:
Worker conclusions rebuild the original manager Inbox config before returning to Lark. That path omitted the manager conversation kind, silently restoring the transient reaction policy from before #4235 and removing the original received ACK after a delayed conclusion.

Preserve the already-validated manager kind when rebuilding the config. The extended real reply-path regression verifies a received receipt survives conclusion delivery and replay, while original-source acknowledgement and authority-revocation checks still apply.

Validation: 145 tests across manager returns, routing, reaction lifecycle, turn-start sync and reply size passed; Ruff and diff hygiene passed. The new regression failed on the attempted received-reaction deletion before the fix. The patch contains only the route metadata and synthetic regression assertions.

### DIFF
--- a/loopx/extensions/lark/manager_returns.py
+++ b/loopx/extensions/lark/manager_returns.py
@@ -58,6 +58,7 @@ def send_return(
             raise ValueError("manager return target disabled")
         routing = {
             "target_ref": binding["target_ref"],
+            "conversation_kind": "manager",
             "app_ref": (target_config.get("identity") or {}).get("sender_profile")
             or "default",
             "topic_root_message_id": (binding.get("topic") or {}).get("root_message_id")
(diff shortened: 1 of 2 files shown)


huangruiteng:
No blocking findings. This closes the second active config producer for manager replies: initial answers and delayed worker conclusions must both preserve the same received-ACK lifecycle. The route kind comes from the manager binding already verified by the return path, so this adds no inferred authority or permission. The existing Lark extension remains the correct owner; the adjacent-boundary pass found no new abstraction necessary.

Equivalent risk-based validation covers the two config producers plus received/processing transitions, replay, sender and revocation boundaries: 145 related tests passed, with a demonstrated pre-fix regression failure, and Ruff/diff hygiene passed. No private state or new policy defaults are introduced beyond #4235. Installed original-message return/readback remains the post-deployment acceptance step. Self-merge under the owner's explicit authorization for manager/Lark fixes.

PR #4301, opened 2026-09-12
fix(subagents): keep coordinators investigating alongside children

huangruiteng:
Enabled sub-agent guidance described coordinator work mostly as validation and integration, making it easy to delegate the investigation and then wait. The existing three lifecycle phases now guide the coordinator to retain a useful independent evidence question, work on it while children run, and revisit uncovered questions after reconciliation.

This remains guidance only: no minimum worker count, forced busywork, new fields, scheduler changes, or disabled-capability activation. Provider revision v2 identifies the changed content. The bounded provider is reused; no additional abstraction is needed.

Validation: 9 TypeScript context tests and 22 Python context/model-configuration tests passed, including disabled parity, managed/native transport, signed envelopes and the maximum-length model identifier within the unchanged context budget. Diff and the three-file public-boundary scan passed. The full operator registry check additionally reported pre-existing unrelated Goal metadata errors; it is not claimed as passing.

User entry points: managed Turn and native agent-context receive the same provider. Existing frontend settings already control enablement, capacity and model preference; this change introduces no setting or UI-owned state, and the existing configuration/roundtrip tests pass. No Lark mutation flow changes.

This PR only improves participation guidance; canonical configuration routing is a separate fix.

中文：通过现有规划、委派、回收三个阶段，引导主 Agent 在子任务运行期间推进独立证据问题，而不只负责汇总验收。没有新增强制并发、完成门禁或配置开关，关闭能力时保持不注入；配置写回来源问题单独修复。

### DIFF
--- a/docs/integrations/codex-subagent-orchestration.md
+++ b/docs/integrations/codex-subagent-orchestration.md
@@ -40,6 +40,24 @@ Parallelize when it reduces uncertainty or latency:
 - run an independent review or validation pass;
 - inspect separate adapter, evidence, or boundary questions.
 
+The enabled capability also guides **coordinator participation** at its existing
+context phases. Before planning, reserve a distinct decision-relevant evidence
+question for the coordinator when useful independent work exists. Before
+launch, identify that question and its dependencies alongside the child briefs;
+continue it while children work. After results return, validate decisive sources,
+reconcile findings and revisit uncovered questions. Reviewing children is still
+necessary, but does not replace the coordinator's own parallel investigation.
+Wait when the remaining useful work depends on child results; do not invent
+work, require a minimum child count, or duplicate evidence to fill capacity.
+These are guidance-only instructions, not a new scheduler or completion gate.
+
+启用能力后，现有三个上下文阶段也引导主 Agent 参与研究：规划前保留一个
+独立、影响决策的证据问题；委派前说明自己的问题、依赖和子任务边界，并在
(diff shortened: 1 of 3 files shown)


huangruiteng:
Reviewed exact head 3edab2945b6f9cfe474be6b2cfd61b8899b06965. No blocker found in the three-file scope.

Product/architecture: coordinator participation belongs in the existing multi_subagent provider. This replaces vague integration-only guidance without introducing a scheduler, new authority, minimum worker count, or busywork. Disabled capability parity and signed/native/managed transport remain covered. Configuration-source persistence is deliberately a separate repair.

Validation: 9 focused TS tests, 22 Python context/model tests with the project runtime, typecheck and diff hygiene pass. Independent reviewer measured all three contributions below 1,300 bytes, within the unchanged 2,048-byte bound. Its broad TS run passed 1,236 tests and encountered 14 unrelated tests invoking an incompatible system Python 3.9; that broad run is not claimed green. The three-file public scan passes; unrelated operator-registry metadata errors are outside this change. Hosted CI has no observed failure at review time but still has running jobs.

User entry points: same context provider reaches native CLI and managed Turn. No configuration field or frontend state is added; existing configuration tests pass. Future-facing pass: reuse the bounded lifecycle provider, no new abstraction. Owner explicitly authorized self-merge for this repair; using scoped local validation and independent review, rather than claiming all hosted checks completed.

PR #4335, opened 2026-09-13
fix(quota): preserve action and monitor receipt identities

huangruiteng:
## Why

After a healthy identity-less heartbeat guard, a legitimate same-Turn action selection can be preempted by a newly due hard lane or autonomous replan. The current CLI collapses that unbound preflight outcome into heartbeat_receipt_identity_conflict and then projects a false write_failed receipt. That makes a safe refusal look like corrupt settlement identity and gives the agent no deterministic recovery path.

Closes #4327.

## Behavior

- Preserve the TypeScript action-selection reducer as the single eligibility/arbitration owner and expose its existing action_selection_qualification_v0 result in the full quota payload.
- Return quota_action_selection_deferred or quota_action_selection_rejected for unbound choices, with the exact typed reason and a same-Turn refresh/re-entry instruction.
- Replay an existing identity-less receipt without mutation; for a first-call rejection, report status=not_committed and append no receipt event.
- Keep an already bound Todo/replan mismatch as the existing hard heartbeat_receipt_identity_conflict.
- Fail closed if the TypeScript qualification is absent or contradicts the selected projection.

## Product entry points

The managed/Codex CLI selection journey changes directly. Frontend and Lark need no separate component or source change: they consume the same interaction/receipt projection, and this patch changes only the second-guard failure/readback packet rather than adding a setting or alternate UI authority. The protocol reference documents the new states and retry behavior.

## Validation

- tests/control_plane/test_quota_settlement_cli.py: 47 passed (full file before the non-overlapping latest-main fast-forward).
- Post-fast-forward exact regression: first-call rejection, autonomous-replan preemption, and true bound-receipt conflict: 3 passed.
- Action portfolio/boundary/effect-turn Python suites: 19 passed.
- TypeScript action-portfolio reducer: 11 passed.
- Quota CLI projection: 9 passed.
- Ruff, git diff --check, and docs governance smoke passed.

## Scope and follow-up

This is a bounded quota/receipt correction. The future-facing pass kept policy in the existing TypeScript reducer and avoided a second Python eligibility rule or a new receipt store. Broader manager handoff work remains under RFC #4330 and its existing milestones.

PR #4333, opened 2026-09-13
fix(heartbeat): reserve readable Reward Memory prompt headroom

huangruiteng:
## Summary

- Keep the Reward Memory outcome contract readable in compact/brief/thin automation prompts instead of compressing away its validator, readback, and private-boundary guidance.
- Apply a fixed `+640` prompt headroom only when Reward Memory automatic outcome ingestion is enabled; feature-off prompts keep their existing budgets and omit the contract.
- Wire the same policy through CLI heartbeat generation and upgrade planning, with regression coverage across all heartbeat modes.

## Validation

- `pytest -q tests/control_plane/test_heartbeat_prompt_support.py tests/control_plane/test_cli_output_differential.py tests/control_plane/test_host_bootstrap_lifecycle.py tests/control_plane/test_automation_prompt_upgrade.py tests/control_plane/test_cli_output_budget.py` — 169 passed.
- Targeted Ruff, `compileall`, and `git diff --check` — passed.
- `loopx canary premerge --from-git-diff --git-diff-base origin/main` — direct checks, Python compile, public-boundary scan, and 15/18 selected smokes passed. Three canary entries failed because the runner selected the machine's Python 3.9 / installed-worktree environment; the same affected canary command that is runnable under the repository Python 3.13 environment passes where independently reproducible. No manual holds.

## Scope / review

Runtime surfaces are heartbeat prompt budgeting/rendering, CLI support control, and upgrade planning; docs and focused tests are included. No frontend, scheduler authority, provider calls, scoring, or private evidence changes. Feature-off parity is asserted for full, compact, brief, and thin prompts.

自审结论：改动保持 Reward Memory 默认关闭时的原有预算和行为；开启 automatic ingestion 时仅为对应自动化提示增加固定容忍，并保留可理解的安全边界说明。

PR #4302, opened 2026-09-12
fix(heartbeat): recover unsettled host turns

huangruiteng:
## Summary

- persist `closeout_required=true` on selected Todo-bound, delivery-enabled must-attempt heartbeat guards
- on the next fresh heartbeat, compare the latest flagged guard with its exact writeback/spend settlement and typed Todo lifecycle
- preempt ordinary selection with `unsettled_host_turn_recovery_v0` when no legal closeout exists
- require a typed external wait plus independent successor, rerun the same current Turn, and continue work without spending quota for the recovery transition
- keep the rule inside the shared live quota/interaction-contract projection used by CLI and managed Turn

## Root-cause boundary

- the host agent remains responsible for obeying `must_attempt_work` and for not treating `DONT_NOTIFY` as an execution gate
- this change closes the LoopX amplification gap: a later heartbeat now detects the missing closeout instead of recommending the same waiting Todo again
- heartbeat cadence is not changed

## Product entry points

- CLI: the recovery packet includes exact typed Todo-update and same-Turn quota-rerun actions
- managed Turn: the shared live decision projects the same recovery mode and no-spend contract
- heartbeat prompts: full, compact, brief, and thin prompts preserve the continuation rule within the existing output budgets
- frontend/Lark: no new setting or separate state source is introduced; both already consume the shared interaction contract, and the recovery user channel remains quiet because it is an agent-owned lifecycle repair

## Validation

- `63 passed` — full quota settlement CLI suite plus managed-Turn and prompt-behavior tests after final rebase
- `11 passed` — maintainability ratchet and the three prior autonomous-replan regressions after rebase
- CLI base/head output-budget regression smoke passed after rebase
- `40 passed` — heartbeat receipt plus actual default-model behavior portfolio
- `141 passed` — heartbeat prompt, lifecycle, activation, dispatch, and host-behavior suites
- Ruff and `git diff --check` passed
- public outbound scan passed with zero findings
- commit carries a valid DCO sign-off and was rebased onto current `origin/main`

Fixes #4300

PR #4296, opened 2026-09-12
fix(loopx): reconcile skill discovery and persistent heartbeat startup

huangruiteng:
LoopX workflows disappeared outside connected projects, while new App heartbeats could still persist stale execution text after an upgrade. This consolidates #4284 and #4288, incorporates the shared renderer from #4297, and fixes the complete discovery/startup/migration path.

- Keep the six reusable packaged workflows globally discoverable. Material and change-quality remain project-default; an explicit project copy may also use a globally scoped canonical source. Visibility does not grant execution, activation or goal authority.
- Reconcile managed duplicate discovery roots, preserve user-owned files, retain one canonical Codex command spelling, and materialize the core `$loopx` entry independently. Ship scope metadata in wheels and resolve sources from the active source/wheel/frozen installation.
- Generate the same heartbeat bootstrap v2 from App startup and automation upgrades. Preserve native Goal scheduler semantics and explicit owner inputs. Recognize authentic legacy loaders strictly, reject malformed loaders, and actually migrate legacy App wrappers to v2.
- Never report live-host migration complete after direct SQLite/TOML writes. Running Apps receive a complete native `automation_update` request; the offline adapter requires the App closed. Update the core entry skill, workflow guidance and activation readback instructions accordingly.

The owning boundaries remain workflow/project skill delivery and heartbeat lifecycle. The bounded refactor reuses source resolution and the bootstrap renderer; no new capability, provider, scheduler, authority store, or configuration UI is introduced. The affected user entrypoints are CLI installation, Codex skill discovery, `$loopx` startup and existing automation upgrades. The existing App automation editor consumes the generated prompt; no frontend implementation change is needed.

Validation: focused skill/packaging/scope, real CLI heartbeat/onboarding and SQLite/TOML migration tests; source installer, wheel install/readback/uninstall, packaged Codex installation and slash-command smokes; risk-selected premerge checks; Ruff and public/private boundary scans. Final counts and exact-head outcomes are recorded in the qualification comment. Windows-specific tests are skipped locally and covered by hosted Windows CI. No benchmark jobs, scoring rules or private goal/session data are changed.

Default behavior changes are intentional: global discovery is retained for general workflows; persistent App prompts become v2 loaders; live disk-only automation migration defers to the App API. Existing schedules, task bindings, explicit policy, modified user files, and non-heartbeat host semantics are preserved.

PR #4359, opened 2026-09-14
fix(smokes): restore the public smoke fleet and the capability roster regression

songoow:
## Behavior

`Full Public Smokes` has been red on `main` for two days (no successful run in the last ~100), with 20+ scripts failing per shard sweep. This restores the fleet by fixing one production regression and realigning the smoke contracts that drifted behind merged behavior changes. Locally 9 of the 12 first-failing scripts pass after this PR; the remaining three are listed under Boundaries.

## Production fix

`resolve_agent_capabilities` re-read the goal registry file to obtain the registered-agent roster before the capability-memory observation. A relative or unreadable registry path silently emptied that roster, the typed handler rejected the read-only observation as an unregistered agent, and `quota should-run` failed with `EffectRuntimeInternalError` instead of returning a projection. The resolved identity already carries the authoritative roster, so it is passed through; roster entries keep agent-id normalization (they are not capability tokens).

## Contract realignments

- **#4302 closeout wording**: `install-local`, `blocker-push` and `heartbeat-prompt` asserted the pre-#4302 `guard receipt; 2 stalls->replan` line.
- **#4333 Reward Memory headroom**: the heartbeat fixtures measured prompts against the bare `INTERFACE_BUDGET_CHARS` ceiling; production budgets add the 640-char outcome-contract headroom when the marker is present. The fixtures now mirror `build_interface_budget`, and in-process comparisons against CLI output opt out of the Reward Memory contract explicitly, because the CLI derives it from the goal registry (absent means off).
- **#4254 typed capability owner**: `capability-gate-projection` imported three private helpers that moved into `capability_gate.ts`; it now reads the public gate. The requirement-ordering case now correctly expects no gate when no candidate declares a requirement.
- **#4283 presentation ordering**: `todo-user-gate-readmodel` pinned a display order that native records without a v0 wire coordinate no longer have; the lane membership assertion is order-independent.
- **#4335 monitor identity**: the rejection report names the actually selected monitor and its declared material-change flag instead of an empty identity.
- **#4296 docs/skill wording** and the thin agent-input projection: assertions follow the current two-layer prompt documentation and read the embedded guard command from the task body.

## Module budgets and classification

`agent-capabilities` and `handoff` were registered without a manual/help-only classification; both are command-specific surfaces and join the help-only set. Three cohesive owners are extracted to return their modules under the 1000-line budget: `quota_failure_report.py` (honest failure/validation payloads, moved verbatim), `support_control_chat.py` (chat + dashboard registration), `project_lifecycle_reward.py` (reward overlay registration). `quota.py` 1118→928, `support_control.py` 1015→889, `project_lifecycle.py` 1014→963.

## Validation

- Focused suites: `test_quota_settlement_cli` 57 passed, heartbeat prompt suites 44 passed, capability-memory CLI tests 3 passed (run via importlib against the repo checkout), shadow/lease suites unaffected.
- Full TypeScript control-plane suite: 1,490 passed (the PostgreSQL integration test requires a local service and fails identically on an unmodified checkout).
- Ruff on all touched files; the extractions also remove the imports that only the moved code used (33→17 pre-existing warnings on those modules; the remainder are pre-existing broad-exception handlers).

## Boundaries

- `turn.py` (1,114 lines) exceeds the budget on `main` before this PR; it is a single large handler and pin-vs-extract is a maintainer decision this PR does not take.
- `configure-goal-smoke` needs a complete `reward_memory_experiment_config_v1` fixture since the enablement preflight landed; that is a corpus-contract decision, not an assertion drift.
- `monitor-poll-writeback`'s capability-declaration parity case fails identically on unmodified `main` and is not addressed here.


songoow:
Closing as superseded: #4355 landed the same fleet restoration while this PR was in flight, covering the smoke contract realignments, the `agent-capabilities`/`handoff` classification, and the capability-memory failure on `quota should-run`. I verified on `c5e43a161` that the previously failing scripts here now pass.

Two notes for the record:

- #4355 resolves the capability-memory failure by failing open (`except Exception: state = {}`), which also drops remembered observations whenever the status packet's registry path is unreadable. This PR instead passed the already-resolved roster from the agent identity through to the observation, keeping remembered capabilities live in that case. If that distinction matters, I can reopen that one change as a small follow-up on top of the fail-open guard.
- #4355 pins `quota.py` (1118) and `support_control.py` (1015) at their baselines "while each cohesive extraction lands". This branch contains two such extractions (the honest quota failure/validation payload owner, and the chat/dashboard registration pair) that retire both pins; I will resubmit them as a focused follow-up against the new baseline.

PR #4409, opened 2026-09-15
feat(turn): bind the default Turn host to the operator credential

huangruiteng:
## Summary

`loopx turn run-once` defaulted to `generic-cli` and `loopx turn plan` hard-coded
`codex-cli`, so an operator who had configured their own model credential still
had to pass `--host dsh` explicitly to keep governed Turns off an individual CLI
subscription.

Default host resolution is now one typed rule in
`loopx/control_plane/turn_driver/host_binding.py`:

| Operator configuration | Default `--host` |
| --- | --- |
| `DEEPSEEK_API_KEY` set to a non-empty value | `dsh` |
| no configured credential | `codex-cli` |

`DEEPSEEK_BASE_URL` selects the endpoint the credential is used against but does
not switch hosts on its own; whitespace-only values count as unconfigured; an
explicit `--host` always wins, and the resolved value is what `loopx turn plan`
reports back. `turn plan` also accepts the `dsh` host it can already route.

## Default behavior change

- `loopx turn run-once`: `generic-cli` -> credential-resolved (`dsh` with a
  configured `DEEPSEEK_API_KEY`, otherwise `codex-cli`).
- `loopx turn plan`: `codex-cli` -> same credential-resolved default; `dsh` added
  to its explicit `--host` choices (`claude-code` planning choice unchanged).
- Affected lanes: codex-cli, dsh, generic-cli Turn hosts.

Tests and one smoke that encoded the former `generic-cli` default now pass
`--host generic-cli` explicitly, so they keep testing the host they named rather
than the ambient default.

## Validation

```
.venv/bin/python -m pytest \
  tests/test_turn_default_host_binding.py tests/test_loopx_turn_driver.py \
  tests/test_loopx_turn_executor.py tests/test_dsh_goal_mode.py \
  tests/test_default_public_scan_root.py tests/control_plane/test_cli_output_budget.py -q
# 226 passed (2:09)
```

`tests/test_turn_default_host_binding.py` covers the resolution matrix
(configured, empty, whitespace-only, endpoint-without-credential), both CLI
defaults with and without the credential, and explicit `--host` precedence.

No live provider call is made by this change or its tests. Public/private scan of
the touched paths is clean (no local paths, credentials, or private context).

## Refactor pass

The resolution rule lives in one typed module with a single call site per
command default, so the next host surface (the manager executor and manager
model defaults) can consume the same rule instead of re-deriving it.

## Stack

This stack now carries the dsh release pin: #4420 (`deepseek-harness-sdk==0.1.5rc1`, upstream released channel) -> #4409 -> #4416 -> #4417 -> #4419. Each branch was rebased on the pin; the managed dsh default therefore runs on the current released runtime rather than the previous `0.1.2a3` pin.


huangruiteng:
Superseded by #4443 (`feat(turn): select the managed Turn host explicitly`).

The value here is kept: the shipped managed default, the explicit `LOOPX_TURN_HOST` selector, the per-command `--host` override, the migrated tests, and the disclosed default-behavior change. What changed is the rule this PR encoded: the Turn host is now **selected, never inferred**. A configured `DEEPSEEK_API_KEY` only authenticates the host that was already selected, instead of resolving the default host and the billing boundary for the operator.

#4443 is main-based rather than stacked, so it needs one review and carries no layer-refresh cost. Closing this to compress the delivery chain.

PR #4412, opened 2026-09-15
fix(doctor): bound offline source registry reads

### LINKED ISSUE #4407: doctor: bound reads of offline source registries
## Problem / 问题

`loopx doctor`, `loopx update check`, and `loopx update plan` can block indefinitely when the global registry contains a `source_registry` path whose metadata exists but whose content read stalls (for example, an offline/unhydrated file-provider placeholder).

The hang occurs in:

`collect_runtime_projection_route_diagnostics -> _source_routes_for_registry -> load_registry -> read_json -> Path.open`

The current code catches `OSError`, `ValueError`, and JSON errors only after the read returns; it has no bounded-read deadline. One stale project registry therefore prevents diagnosis and even blocks the supported update-readiness path for every healthy Goal.

全局 registry 中只要有一个“路径存在、但读取内容会阻塞”的 `source_registry`（例如离线或未 hydrate 的文件提供商占位文件），`doctor`、`update check` 与 `update plan` 就会无限等待。现有异常处理只有在读取返回后才生效，导致单个历史项目阻塞所有健康 Goal 的诊断和升级。

## Reproduction evidence / 复现证据

- The installed runtime was upgraded successfully to an immutable current-main snapshot.
- Both pre-upgrade and post-upgrade `loopx --format json doctor` blocked for more than 60 seconds at the same `Path.open` call.
- A bounded per-source probe identified exactly one existing regular-file registry whose first-byte read timed out; all other registered source registries returned immediately.
- No registry mutation was required to reproduce.

No local absolute paths are included here because the bug is provider/state-class generic.

## Expected behavior / 预期行为

Runtime projection diagnostics should treat one stalled source registry as bounded, typed diagnostic evidence and continue:

- enforce a short per-source and/or aggregate read deadline;
- report a public-safe `source_registry_timeout` or equivalent status without exposing local paths;
- keep `doctor` and update check/plan responsive;
- do not silently drop the affected Goal or mutate/remove its registry binding;
- preserve deterministic readback and allow a later retry after the file becomes available.

诊断应对单个 source registry 使用有界读取，超时后输出不含本地路径的 typed 状态并继续；不得静默删除 Goal、修改 registry 或阻塞整个 doctor/update 流程。

## Acceptance / 验收

1. A blocking/offline source-registry fixture cannot hold `doctor` or update check/plan beyond the documented deadline.
2. The affected route is reported as unavailable/timeout while healthy routes remain present.
3. Normal local files retain current behavior.
4. Retry after hydration succeeds without registry repair or state loss.
5. Tests cover both the runtime projection diagnostic and update-readiness caller.

huangruiteng:
Closes #4407

## Summary
- bound each unique runtime-projection source-registry diagnostic read to a one-second deadline
- report a typed, path-free unavailable / source_registry_timeout item while continuing to inspect healthy Goals
- preserve missing and unreadable classifications, leave registry bindings untouched, and retry from source on every later diagnostic call
- cover doctor diagnostics plus update check and update plan callers

## Product delivery
Affected entry points are the doctor CLI and the update check/plan CLI, which all consume the existing runtime-projection diagnostic owner. No frontend or Lark control changed: neither surface owns this read or exposes a separate source-registry action, and the existing status projection continues to consume the same healthy boolean.

## Validation
- python -m pytest -q tests/test_doctor_installation_scope.py tests/test_self_update_runtime_activation.py: 21 passed, 1 skipped
- ruff check on all changed Python files: passed
- py_compile on all changed Python files: passed
- live-checkout loopx doctor, update plan, and update check: returned successfully
- premerge canary: 14 of 15 selected smokes passed; install-local-smoke failed on the existing packaged skill-id assertion and the same failure was reproduced unchanged on clean origin/main

## Boundary and review
The public diagnostic contains Goal id, typed status/reason, and the configured deadline only; it does not expose source paths. No registry state is written. The helper stays in the existing Python filesystem-diagnostic adapter and does not duplicate TypeScript state authority. This changes runtime behavior, so it is intentionally left for separate review rather than self-merged.


huangruiteng:
CI triage for exact head `c54081e7d8fcafd30612cb1ab5b9c1322d449ea6`:

- `test-shard (3)` failed only the fixed-date `resume_at:2026-09-15T00:00:00Z` case in `tests/control_plane/test_delivery_response.py`. I reproduced the identical failure both on this PR head and on a clean detached current `origin/main` (`a1a91dc61`), so it is not caused by the source-registry timeout diff. The failed shard rerun is in progress.
- PostgreSQL qualification completed its service checks and failed only while finalizing the evidence artifact with HTTP 403; failed-job rerun passed completely.
- All changed-path checks and the remaining test lanes passed.

No PR patch is warranted for either failure. The delivery-response date-bound regression should be repaired independently on main.

PR #4420, opened 2026-09-15
chore(dsh): pin the managed dsh host to the latest released SDK

huangruiteng:
## What

Pins the managed dsh host to the newest **released** dsh channel:

- PyPI `deepseek-harness-sdk==0.1.5rc1`, which pulls its bundled
  `deepseek-harness-runtime-bin==0.1.5rc1`.
- npm `@deepseek-ai/dsh` publishes the same release train as `latest`
  (`0.1.5-rc.1`); newer `next` / `alpha` tags exist upstream but are not the
  released channel.

Previously pinned: `0.1.2a3` (2026-09-01).

## Why

LoopX now binds the default Turn and steward host to dsh when an operator
credential is configured (#4409, #4416). The runtime behind that default should
be the current released dsh, not a two-week-old pin.

## Surface diff

The SDK's Python client is byte-identical between `0.1.2a3` and `0.1.5rc1`: a
wheel diff shows only `dist-info` metadata changed. What moves is the bundled
self-contained dsh runtime. LoopX passes no `profile`, so it runs the SDK's
default `sdk` profile unless the operator supplies `--cordis`; the upstream
`sdk-minimal` profile change in this release therefore does not affect the
managed host path.

## Validation

Hermetic runs against a local mock OpenAI-compatible endpoint: no provider
call, no API key, no Codex subscription use.

| check | 0.1.2a3 (control) | 0.1.5rc1 |
| --- | --- | --- |
| `examples/loopx-turn-dsh-real-e2e-smoke.py --host generic-cli` | passed | passed |
| `examples/loopx-turn-dsh-real-e2e-smoke.py --host dsh` | passed | passed |
| `tests/test_dsh_goal_mode.py` | — | 55 passed |
| `examples/dsh-turn-host-adapter-smoke.py` | — | 11 checks passed |
| `examples/loopx-turn-dsh-e2e-smoke.py` | — | passed |
| `examples/loopx-turn-dsh-builtin-host-e2e-smoke.py` | — | passed |

The candidate release was installed into a throwaway target directory with
`uv pip install --target ... --no-deps` and injected on `PYTHONPATH` (the shared
dev environment already provides `pydantic 2.13.4`, above the release's
`>=2.12` floor), so the previous pin stayed untouched during the comparison.

## Upstream gap carried forward

The stock `headless` profile still cannot boot from the bundled snapshot at
`0.1.5rc1`: `@deepseek-ai/dsh-session-title-first-prompt-llm` imports the
[...]

PR #4419, opened 2026-09-15
feat(chat): show the steward execution binding in the manager header

huangruiteng:
## Motivation

The steward channel already **selects** its executor endpoint and model, and the Chat
capabilities payload already projects that selection (`manager.channel_binding`:
endpoint, endpoint source, executor kind, resolved model and source, credential
variable name, `available`, `unavailable_reason`). The packaged frontends rendered none
of it, so an operator could not read which executor answers for the steward, which
credential pays for it, which model follows it, or that a selected managed host cannot
serve the channel at all. The contract doc already promised "a frontend can show which
executor and model the steward channel resolved", with no surface doing it.

## What changed

One compact chip in the manager channel header — `executor · executor kind · model` —
plus the transport gate inline when the projection proves the selected endpoint cannot
serve this channel.

- `apps/presentation/dashboard/src/data/chat.ts` lifts the payload shape into a named,
  reused `managerChannelBindingSchema` / `ManagerChannelBinding` and declares the
  optional `manager.channel_binding` field, so the projection is no longer silently
  dropped by the capabilities parse.
- `channel-header.tsx` renders the chip from the projected fields only. The kind keeps
  the `individual` / `managed` vocabulary of the governed Turn surface, and a kind this
  build does not recognize renders as an unclaimed registered endpoint instead of being
  labelled as one the operator can act on.
- `dashboard-page.tsx` / `personal-workspace-page.tsx` thread the projection through one
  prop; the value is never derived locally.
- An unavailable selection is carried by the chip itself (`is-unavailable`) and the
  narrow header keeps this row, so a wrapped-out reason cannot leave the state looking
  healthy.

This is display-only: no change to selection, credential resolution, model resolution,
Chat admission, or the managed-host transport gate. An older service that projects no
binding keeps the previous header.

## Evidence

- New browser scenario `execution-chip` (registered in `personal-workspace-browser-smoke.mjs`)
  covers the shipped binding, the managed-host binding, an unrecognized kind, the
  no-binding case, and the 390px layout where the reason must stay readable.
- `npm run build:chat` (tsc + vite), `npm run smoke:goal-order`, the workspace theme test,
  and `loopx canary premerge --from-git-diff` (19 selected, 0 failures) pass locally.
- Browser smoke passes in both modes: `scenarios=navigation-sorting,chat-recovery,typed-actions,execution-chip`.
- Packaged surface: `loopx/web/chat/index.html` references `assets/index-BWxHtIJu.js` and
[...]

PR #4416, opened 2026-09-15
feat(turn): report the managed executor and fail closed when it cannot launch

huangruiteng:
## Summary

A planned managed Turn now reports which executor it would run on and fails
closed when that executor provably cannot launch here. Stacked on the
credential-resolved default Turn host change.

## Changes

- `managed_executor_binding` projects the executor, its kind (`managed`,
  `individual`, `generic`), the credential env var name (never its value), a
  launchability verdict, and a typed reason when unavailable.
- `loopx turn plan` and `loopx turn run-once` attach `managed_executor` to their
  payload.
- The Turn executor stops before any effect when the planned executor is
  unlaunchable: status `unavailable`, typed reason, no journal write, no quota
  spend.
- `turn plan` default execution mode follows the resolved host, so a managed
  default plans `isolated-headless` instead of an execution mode the outer
  controller rejects.

## Validation

- `tests/test_turn_managed_executor_binding.py`, executor fail-closed and
  plan-preview tests, `tests/test_turn_default_host_binding.py` CLI default-mode
  test; 93 focused tests then 290 across the turn driver/executor suites, green.
- `examples/loopx-turn-managed-executor-binding-smoke.py` proves the four-step
  CLI behavior, including an import-blocked dsh runtime.
- Live CLI readback (no provider calls): no credential -> `codex-cli` /
  `individual`; credential -> `dsh` / `managed` / `available: true`; blocked
  runtime -> `dsh` / `available: false` / `dsh_runtime_unavailable`.

## Boundaries

- No credential values are read or projected; only env var names appear.
- No provider calls are made by tests or the smoke.

## Review notes

- Interface budget: `loopx turn plan` gains the semantic field
  `managed_executor`, attributed as `managed_executor_binding_v0`. The
  agent-facing CLI base/head differential grants one bounded, one-time growth
  allowance (512 chars / 512 bytes / 12 lines / 448 compact chars) bound to the
  declared none-to-v0 binding transition on `loopx_turn_plan`,
  `loopx_turn_plan_transaction_detail` and `loopx_turn_run_once_preview` only.
  Quota, status and every other agent-facing surface keep their ordinary budget,
  and a v0-to-v0 change receives no allowance.
- Module budget: the fail-closed decision and the payload projection moved to
  `host_binding.py`, the binding owner, so the already-large
  `turn_driver/executor.py` does not grow (1507 -> 1496 lines; limit 1500).
- Validation: `tests/canary/test_maintainability_ratchet.py` 8 passed;
[...]

PR #4400, opened 2026-09-14
docs(rfc): record the shipped managed selection surfaces and promotion gates

huangruiteng:
## Summary

A docs-only refresh of the managed-runtime selection RFC so it records the rule the repository **selects**, not the one it used to infer. This replaces the earlier revision of this PR, which recorded the credential-resolved host rule.

- The Turn host is **selected, never inferred**: shipped default `dsh`, `LOOPX_TURN_HOST` re-points it, an explicit `--host` wins over both. A configured `DEEPSEEK_API_KEY` only authenticates the host that was already selected (PR #4443).
- The steward channel is a **different** surface with its own default: `codex`, because that is the only transport that can hold an interactive steward session today. `LOOPX_MANAGER_ENDPOINT` re-points it, and its model follows the executor it selected, so a configured credential can no longer swap the model while the endpoint stays on the CLI (PR #4446).
- The role table gains a steward channel executor row, and the `codex-cli` row now states that it is the `individual` executor kind billed to one person's CLI login rather than a managed lane.
- The steward-channel section records the real promotion gate for the managed host: an interactive Chat transport, not a credential appearing. Selecting the managed host today reports `available: false` / `managed_host_chat_transport_unsupported` and fails a session request as the same typed host-tool gate.
- The milestone table records the steward channel's M1 contract as *selected* -- endpoint, model, source, and executor kind -- and keeps the session-identity gap explicit: a channel answer still cannot prove which session served it.
- Evidence and pin rows are re-pointed at the replacement PRs (#4443, #4446) and the dsh pin PR #4420. The branch is rebased onto current `main`.

## What this is not

No runtime is promoted, no scoring or performance number is added, and no default behavior changes here. Session ownership (`managed_runtime` vs `attached_host`) and the M1-M4 integration milestones stay owned by the agent session execution modes RFC.

## Validation

- `examples/docs-governance-smoke.py` -> ok
- `git diff --check` -> clean
- English and Chinese mirrors updated together and diffed against each other for the changed rows.

PR #4427, opened 2026-09-15
feat(turn): qualify the credential-resolved default managed flow

huangruiteng:
## What this adds

The **credential-resolved default managed flow** was shipped across the stack but only the *explicit-host* path was ever qualified end to end. This PR closes that gap and fixes the one readback claim that flow exposes.

**1. `operator_credential_bound` in the managed executor readback** (`loopx/control_plane/turn_driver/host_binding.py`)

`managed_executor_binding("dsh")` reported `executor_kind: managed` even with no operator credential configured. That is the field an operator and the manager header read to know where a Turn's model work is billed and bounded, so it must not be inferred from the host id alone. The binding now also states the narrower claim:

| selection | `executor_kind` | `operator_credential_bound` |
| --- | --- | --- |
| credential configured, default host | `managed` | `true` |
| explicit `--host dsh`, no credential, no runner | `managed` | `false` |
| explicit `--runner`, no credential | `managed` | `true` |
| `codex-cli` / `claude-code` / `generic-cli` | `individual` / `generic` | `false` |

Launch decisions are unchanged: `available` still decides the fail-closed refusal, and no host moves on its own.

**2. Public smoke for the default flow** (`examples/loopx-turn-managed-default-flow-smoke.py`)

Hermetic (local mock OpenAI-compatible SSE endpoint; no operator key, no individual CLI subscription) and driven only through the public CLI:

- no credential: default host `codex-cli`, executor `individual`;
- credential: default host `dsh`, executor `managed`, `available: true`, mode `isolated-headless`;
- credential: `turn run-once` **without `--host`** starts the real dsh runtime, commits one validated Turn (`status: committed`, `result_kind: validated_progress`, independent validation `passed`, one quota slot), and reports mode/executor/status;
- credential but runtime hidden: typed `dsh_runtime_unavailable` refusal with no host invocation, no state write, no journal, no spend;
- explicit `--host dsh` with no credential: readback reports `operator_credential_bound: false`.

## Validation

- `pytest -q tests/test_loopx_turn_executor.py tests/test_turn_managed_executor_binding.py tests/test_turn_default_host_binding.py` → **84 passed**
- `examples/loopx-turn-managed-default-flow-smoke.py` → **passed**
- `examples/loopx-turn-managed-executor-binding-smoke.py` → **passed**
- `examples/loopx-turn-dsh-real-e2e-smoke.py --host dsh` and `--host generic-cli` → **passed**
- `examples/docs-governance-smoke.py` → **ok**
- `tests/test_turn_managed_executor_binding.py tests/test_turn_default_host_binding.py` → 27 passed

All runs used the shared repo venv with `PYTHONPATH` pointed at this worktree; no provider call, no Codex subscription use.

## Delivery boundary

Partial by design: the minimal managed slice still owes the **second-executor** guard (a Turn must fail closed when another executor is already live on the same agent lane). No such guard exists today — the Turn CLI has no lane lease or writer fence, and the Turn session binding carries only `goal_id`/`agent_id`/`todo_id`. That is tracked as a successor item with the evidence rather than guessed at here, so this PR stays reviewable and reversible.

## Stack

Stacked on `codex/steward-execution-chip-20260915` (#4419). Merge order stays #4420 → #4409 → #4416 → #4417 → #4419 → this PR; #4400 stays standalone.

🤖 Generated with [Codex](https://openai.com/codex)


huangruiteng:
Superseded by #4443 (`feat(turn): select the managed Turn host explicitly`).

The value here is kept, and the qualification is stronger: `examples/loopx-turn-managed-default-flow-smoke.py` still runs the **real** dsh runtime against a local mock OpenAI-compatible SSE endpoint (no operator key, no individual subscription) and commits one validated Turn. It now qualifies the selected managed default instead of a credential-resolved one.

#4443 also carries the second-executor protection question this PR flagged as open, so the flow evidence and the rule it qualifies land together. Closing to compress the delivery chain.

PR #4396, opened 2026-09-14
fix(runtime): pin UTF-8 decoding for every remaining text-mode subprocess read

BigDataDZ:
Fixes #4155

## What

`subprocess.run(..., text=True)` without an explicit encoding decodes the child's stdout with `locale.getpreferredencoding(False)` — gbk on a zh-CN Windows host — while `gh`, `git`, and the wrapped LoopX CLI all emit UTF-8. A single non-ASCII byte killed the pipe reader thread, left `result.stdout = None` with `returncode 0`, and surfaced as a misleading secondary error. The PyInstaller-frozen sidecar overrides `PYTHONUTF8`, so an explicit codec is the only robust fix.

The already-merged #4158 / #4160 / #4161 / #4338 slices fixed their own paths; this closes the remaining runtime surface:

- **Mechanical sweep**: the 116 remaining call sites across 68 files in the shipped `loopx/` package now pin `encoding="utf-8", errors="replace"` (tests/, examples/, scripts/ are out of scope per the issue).
- **`change_quality/scope.py`**: its git calls funnel through `_git(..., text=...)`, so the codec is pinned only when the caller requested text mode — a non-None encoding would silently flip bytes-mode callers into text mode.
- **Static contract test**: walks the shipped package and fails when any text-mode subprocess call omits an explicit codec, so this class of bug cannot silently return.
- **Behavioral regressions**: the goal-mode CLI JSON probe under a simulated cp936 host (real child emitting UTF-8 bytes), and `_git` in both modes with non-ASCII content.

## Verification

- New + pre-existing encoding suites: 54 passed.
- Suites for heavily patched modules (canary, lark contract, goal-mode MCP, kunluncode, self-update download) pass; the only failures reproduce identically on unmodified `main` (host environment lacks Node ≥ 22.18).
- `ruff check` clean on the touched paths.

PR #4448, opened 2026-09-15
fix(frontstage): classify exported page links as pages or shipped assets

huangruiteng:
## Problem

`main` is red. The `Frontstage Pages` → `build` check has failed on every commit since the architecture diagrams shipped in the editorial blog pages:

```
Error: Missing expected file: /tmp/loopx-frontstage-share-bundle-smoke/site/blog/images/long-horizon-control/position-en.svg/index.html
    at assertExists (examples/frontstage-share-bundle-smoke.mjs:34:11)
    at examples/frontstage-share-bundle-smoke.mjs:131:7
```

The blog pages link their diagrams with a relative anchor:

```html
<a href="../images/long-horizon-control/position-en.svg" target="_blank" rel="noopener">
```

The smoke walked every relative `<a href>` and assumed it was another exported **page directory**, asserting `<target>/index.html`. For an asset link that can never exist, so the check fails on correct content and blocks unrelated PRs.

## Change

Relative references are now classified by form:

- **asset** reference (known file extension) → the asset file itself must ship;
- **page** reference → `<target>/index.html` must exist, as before.

The scheme skip now recognizes any `scheme:`/protocol-relative/`#` reference instead of only `http`/`https`, so a legitimate `mailto:` or `data:` anchor cannot produce the same class of false failure.

Local `` diagram sources are checked with the same helper. The diagrams are the primary content of these pages, and before this change an `` pointing at a diagram that never shipped would pass silently.

## Evidence

Exported bundle, then ran the smoke on this branch:

```
$ node examples/frontstage-share-bundle-smoke.mjs
frontstage-share-bundle-smoke: ok
```

Negative control — moved `position-en.svg` out of the site public tree, re-ran, then restored:

```
Error: Missing expected file: .../site/blog/images/long-horizon-control/position-en.svg
    at assertRelativeReferenceExists (examples/frontstage-share-bundle-smoke.mjs:48:5)
```

The new assertion is live: it fails on the missing diagram and names the asset. Before the change the same input could not pass at all.

Companion check: `python examples/frontstage-pages-workflow-smoke.py` → `frontstage-pages-workflow-smoke: ok`.

## Boundary

No presentation, asset, or workflow change — the smoke script is the only file touched. No required coverage is removed: page references keep their `index.html` assertion, and asset coverage is added where there was none.

Local `python3` is 3.9 and the export's fixture helper needs 3.13, so the run above used a PATH shim onto the 3.13 interpreter; CI uses its own 3.11+ setup.


huangruiteng (commented):

**Exact head reviewed**: `50fbf2e63c5d0ff865fdd7b53cad3e1e5d244d95`

**Verdict**: APPROVE. Single-purpose, 1 file, +25/-2, no runtime, asset, workflow, or first-screen presentation change.

**What I verified**

- Root cause reproduced from CI, not inferred: run 34961535228 job 104356022184 fails with `Missing expected file: .../site/blog/images/long-horizon-control/position-en.svg/index.html`, and the same job fails identically on `main` (2f84af990). Main has been red since the diagrams shipped; this is not a #4396 regression.
- The page really does ship that anchor. `apps/presentation/site/public/blog/from-one-shot-agents-to-long-horizon-control/index.html` links `<a href="../images/long-horizon-control/position-en.svg">` for each of the three diagrams, in both the EN and ZH pages. The old loop required every relative `<a href>` to be a page directory, so correct content could never pass.
- Positive: exported the bundle locally and the smoke passes (`frontstage-share-bundle-smoke: ok`).
- Negative control: moved `position-en.svg` out of the site public tree, re-ran, got `Missing expected file: .../position-en.svg` reported from `assertRelativeReferenceExists`. Restored the file and re-ran to `ok`. The new asset assertion is live, not dead code.
- Companion: `python examples/frontstage-pages-workflow-smoke.py` → ok.
- CI: 19/19 non-skipped checks pass, including `build` (the check this fixes) at 5m50s.

**Gate direction**: no coverage is removed. Page references keep the `<target>/index.html` assertion; asset references now assert the asset file; local `` diagram sources are checked with the same helper and were previously unchecked. The scheme skip broadened from `http`/`https` to any `scheme:`/protocol-relative/`#` reference, which cannot mask a bundle-relative path.

**Pre-merge canary**: `loopx canary premerge --from-git-diff` → `merge_gate_passed: true`, `self_merge_allowed: true`, 3 catalog checks + public-boundary scan passed.

**Residual risk**: local `python3` is 3.9 and the export fixture helper needs 3.13, so the local run used a PATH shim onto 3.13; the CI Python setup is the authoritative environment and it passed.

PR #4443, opened 2026-09-15
feat(turn): select the managed Turn host explicitly

huangruiteng:
## Summary

The Turn host is now **selected, never inferred**, and the operator credential
only *authenticates* the host that was already selected. This replaces the
credential-resolved default rule, which let a configured `DEEPSEEK_API_KEY`
silently re-point `loopx turn plan` and `loopx turn run-once` from the
individual Codex CLI host to the managed dsh host.

Discovering a credential is not a decision to change where a Turn runs. Under
the old rule an operator could not tell a chosen configuration from an
incidental one, and the affected surfaces reported a billing boundary (`managed`
vs `individual`) that the environment had picked for them.

## What changed

| surface | value |
| --- | --- |
| shipped default host | `dsh` (managed executor) |
| explicit default selector | `LOOPX_TURN_HOST` |
| per-command override | `--host codex-cli\|claude-code\|dsh\|generic-cli` (plan), `codex-cli\|dsh\|generic-cli` (run-once) |
| authenticating credential | `DEEPSEEK_API_KEY`, optional endpoint `DEEPSEEK_BASE_URL` |

- `resolve_default_turn_host` / `selected_turn_host` in
  `loopx/control_plane/turn_driver/host_binding.py` are now pure functions of
  configuration: the product default applies, `LOOPX_TURN_HOST` re-points it,
  and no credential value or presence participates in selection.
- `loopx/control_plane/operator_credential.py` reports credential *facts* only
  and documents that selection is owned by each surface, not by this module.
- The planned execution mode follows the **selected** host: the managed default
  plans `isolated-headless` rather than an execution mode the outer controller
  rejects.

## The managed host authenticates or it refuses

`managed_executor_binding` projects the executor, its kind (`managed`,
`individual`, `generic`), the credential env var **name** (never its value), the
endpoint env var name, whether the executor is operator-credential-bound, and
whether it can launch here. When it cannot, `available` is `false` and
`unavailable_reason` names the missing fact:

| `unavailable_reason` | meaning | remediation |
| --- | --- | --- |
| `dsh_runtime_unavailable` | the DeepSeek Harness runtime is not importable and no explicit runner hook was supplied | install the released runtime, pass its runner hook, or select `--host codex-cli` |
| `operator_credential_unconfigured` | the managed host is selected but nothing would authenticate it | set `DEEPSEEK_API_KEY`, or select `--host codex-cli` explicitly |

`run-once --execute` fails closed on that verdict: status `unavailable`, no host
[...]

PR #4450, opened 2026-09-15
refactor(turn): bind the advisory primary through the shared decision owner

huangruiteng:
## Problem

`main` is red on `full-public-smokes`:

```
AssertionError: turn.py has 1123 lines, above budget 1114; extract a cohesive command owner before adding more code
```

The managed Turn host-selection wiring took `loopx/cli_commands/turn.py` from 1094 to 1123 lines against a frozen baseline of 1114, and the ratchet smoke (`examples/cli-command-module-size-ownership-command-modularization-smoke.py`) correctly rejects that.

While finding a cohesive extraction, the underlying duplication was the right thing to fix: `turn.py` reimplemented the controller advisory-primary rebinding inline, and `loopx/cli_commands/turn_decision.py::apply_controller_advisory_primary` already owns that exact rule for `managed-step`. Its own docstring states the intent:

> `run-once` and `managed-step` must agree on what "the current governing decision" means: read the same live status, apply the same controller advisory primary, and sign the same `loopx_turn_envelope_v0`. Duplicating that chain would let the two drift, so both owners build through this module.

The two copies had to stay step-identical by inspection: resolve the advisory primary, re-resolve the decision with it requested, fail closed when it is no longer eligible, record `selected_by`, and keep the advisory portfolio on the decision.

## Change

`turn.py` calls the shared owner instead of repeating the rule:

```python
decision = apply_controller_advisory_primary(build_turn_decision)
```

No behavior change. `turn.py` is back to 1111 lines, under its 1114 baseline, and the fail-closed rule now has exactly one home so the two Turn subcommands cannot drift.

## Evidence

- `python examples/cli-command-module-size-ownership-command-modularization-smoke.py` → `ok` (this is the red check).
- `pytest tests/test_loopx_turn_driver.py tests/test_loopx_turn_managed_step.py tests/test_turn_default_host_binding.py -q` → 109 passed. This includes `test_turn_cli_binds_advisory_primary_without_hiding_portfolio`, which drives the CLI end-to-end and asserts both `selected_by == "turn_controller_advisory_primary"` and that the advisory portfolio stays on the envelope — i.e. the behavior this refactor moves.
- `pytest tests/test_loopx_turn_driver.py -k advisory_primary -q` → 1 passed.
- `ruff check loopx/cli_commands/turn.py` → clean.
- `loopx canary premerge --from-git-diff` → `merge_gate_passed: true`, `self_merge_allowed: true`; 4 catalog canaries including the maintainability ratchet and CLI output-budget smokes passed.

## Boundary

`loopx/cli_commands/turn.py` only, +5/-17. No new module, no new surface, no budget number was raised or lowered.


huangruiteng (commented):

**Exact head reviewed**: `a1213edce`

**Verdict**: APPROVE. One file, +5/-17, no user-visible surface, no new module, no behavior change.

**What I verified**

- The red check is real and reproduced: `python examples/cli-command-module-size-ownership-command-modularization-smoke.py` fails on latest `main` with `turn.py has 1123 lines, above budget 1114`. The regression is mine — `3f916ff38` (#4443) took `turn.py` from 1094 to 1123 against a frozen baseline of 1114 that was already in place at `3f916ff38^`.
- The extraction is not invented for the gate. `turn_decision.apply_controller_advisory_primary` already owns this exact rule and `turn.py` reimplemented it inline, step for step: resolve the advisory primary, re-resolve with it requested, fail closed when it is no longer eligible, set `selected_by`, keep the advisory portfolio. `turn_decision.py`'s own docstring says both Turn subcommands must build through it so they cannot drift. This is duplicate knowledge being collapsed, which is the refactor the size gate is asking for.
- Call counts are unchanged: both the removed inline version and the shared function call the builder once, and a second time only when an advisory primary was returned.
- Behavior is pinned by an existing end-to-end test, not by inspection: `tests/test_loopx_turn_driver.py::test_turn_cli_binds_advisory_primary_without_hiding_portfolio` drives the CLI and asserts both `selected_by == "turn_controller_advisory_primary"` and that the advisory portfolio survives onto the envelope. It passes.
- `pytest tests/test_loopx_turn_driver.py tests/test_loopx_turn_managed_step.py tests/test_turn_default_host_binding.py -q` → 109 passed.
- `ruff check loopx/cli_commands/turn.py` → clean. `py_compile` → clean.
- The ratchet smoke is green at the new head (1111 ≤ 1114).
- `loopx canary premerge --from-git-diff` → `merge_gate_passed: true`, `self_merge_allowed: true`; 4 catalog canaries passed including the maintainability ratchet and CLI output-budget smokes.
- `docs/reference/protocols/turn-envelope-v0.md` documents `selected_by=turn_controller_advisory_primary`; that value is unchanged.

**Boundary**: `loopx/cli_commands/turn.py` only. No budget number was raised or lowered, so the ratchet keeps its existing meaning. No frontend, Lark, or CLI entry point changes, so there is no user-visible surface to read back.

**Residual risk**: none identified for this diff. `full-public-smokes` shard 0 on `main` has two further, unrelated failures that pre-date this branch — `cli-help-manpage-smoke` (`unclassified: ['goal-actions']`) and `canary-promotion-readiness-smoke` (nested `install-local-smoke.py` exits 1 in CI, while it passes locally on a clean latest-`main` worktree). Neither is touched here and neither is caused by this change.

PR #4454, opened 2026-09-15
feat(turn): resolve the default host from the operator credential

huangruiteng:
## What changed

The bounded Turn default host is now resolved from the operator's own credential
facts instead of being fixed to the managed host:

- operator credential configured (`DEEPSEEK_API_KEY`): the default stays `dsh`
  (managed executor), authenticated by that credential;
- no operator credential configured: the default resolves to `codex-cli`
  (individual executor), because a managed host nothing can authenticate would
  otherwise refuse to run.

An explicit `--host` or `LOOPX_TURN_HOST` still wins over either default, so a
credential resolves the shipped default only -- it never re-points a host the
operator already selected. The readback now carries the source
(`explicit_config`, `operator_credential`, `no_operator_credential`).

## Changed surfaces

Runtime/API: `loopx/control_plane/turn_driver/host_binding.py` (default
resolution and readback source), `loopx/control_plane/operator_credential.py`
(module contract), `loopx/cli_commands/turn_registration.py` (comment only),
`loopx/semantics/inventory_v0.json` (regenerated).

CLI/plan/run-once: `loopx turn plan` and `loopx turn run-once` without `--host`.
No frontend change is needed: the resolved host and the `managed_executor`
readback are already what the Chat/Dashboard surfaces project, and this change
does not alter that projection.

Validation: both public smokes and both host-binding test modules.

Docs: Turn protocol reference, DeepSeek Harness connector guide, and the
DSH/Pi harness RFC (English + Chinese).

## Default behavior change disclosure

Both `plan` and `run-once` previously defaulted to `dsh` regardless of the
credential, which made a lane without a credential fail closed on
`operator_credential_unconfigured`. A lane without an operator credential now
runs the individual CLI host by default.

`--host dsh` remains the explicit managed path and still fails closed with the
same typed reason when nothing can authenticate it. `--host generic-cli` remains
the compatibility path, and a machine that wants one fixed host should set
`LOOPX_TURN_HOST` once instead of relying on the ambient environment.

## Validation

- `python -m pytest tests/test_turn_default_host_binding.py tests/test_turn_managed_executor_binding.py tests/test_loopx_turn_executor.py tests/test_host_mode_planner.py tests/test_host_parity_smoke.py` -> 134 passed
- `python -m pytest tests/test_loopx_turn_driver.py tests/test_loopx_turn_managed_step.py tests/test_dsh_goal_mode.py tests/test_default_public_scan_root.py tests/control_plane/test_cli_output_budget.py` -> 182 passed
[...]

PR #4476, opened 2026-09-15
fix(host-mode): state the shipped Turn host default in the plan contract

huangruiteng:
## What changed

The host-mode planner previews the *resolved* host default for the headless Turn modes instead of pinning one, which is the right behavior. Three surfaces explained that preview with semantics the governed Turn does not ship:

- the `RESOLVED_DEFAULT_TURN_HOST_MODES` comment in `loopx/host_mode_planner.py`;
- the `host` field description in `docs/reference/protocols/host-mode-plan-v0.md`;
- the inline comment in `examples/host-mode-plan-smoke.py`.

All three said the headless default is resolved **from the operator credential**. `control_plane.turn_driver.host_binding.selected_turn_host()` is environment-independent: it ships one explicit product default that `LOOPX_TURN_HOST` or an explicit `--host` re-points, and a configured credential authenticates that host instead of selecting it. The preview behavior is unchanged; only the stated reason was wrong — and a stated reason is what a later editor acts on, so the comment now names the owner that decides the default instead of restating a rule that does not hold.

The smoke pins the reason as well: the headless mapping is built with and without `DEEPSEEK_API_KEY` and must stay identical and unpinned. Wiring credential resolution back into the planner now fails a public check instead of passing review.

## Scope

Public-safe: no private context, credentials, host names, or local paths. No runtime behavior change, no new CLI surface, no permission or evidence-policy change.

## Validation

- `python3 examples/host-mode-plan-smoke.py` → `host-mode-planner-smoke ok` (includes the new credential-independence check)
- `python3 examples/project/host-mode-plan-cli-smoke.py` → ok
- `python3 -m pytest -q tests/test_host_mode_planner.py` → 2 passed
- `python3 -m pytest -q tests/test_host_parity_smoke.py` → 38 passed
- `python3 scripts/generate_semantic_inventory.py --check` → up to date

## Context

The credential-resolved Turn default this wording came from was proposed in #4454 and deliberately not adopted: that PR is closed, and `selected_turn_host` keeps selection independent of the environment.


huangruiteng:
## Self-review validation comment

Changed surfaces: `loopx/host_mode_planner.py` (comment only), `docs/reference/protocols/host-mode-plan-v0.md` (contract prose), `examples/host-mode-plan-smoke.py` (comment + one new check).

Checks run, all at head `b4faf001d`: `examples/host-mode-plan-smoke.py` → ok; `examples/project/host-mode-plan-cli-smoke.py` → ok; `tests/test_host_mode_planner.py` → 2 passed; `tests/test_host_parity_smoke.py` → 38 passed; `scripts/generate_semantic_inventory.py --check` → up to date.

Failures/skips: none. Manual holds: none.

Why this coverage is enough: the only executable change is a comment inside the planner, so behavior is unchanged by construction; the contract prose has no other consumer than the smoke's doc-wiring check, which passes; and the new assertion is the regression guard for the sentence being fixed (the headless mapping must not depend on `DEEPSEEK_API_KEY`).

PR #4545, opened 2026-09-16
fix(quota): make an unreconcilable action selection self-diagnosing

huangruiteng:
## Problem

When a guard is bound with `--todo-id` and that selection cannot be reconciled with the current projection, the preflight raised a bare `RuntimeError`. The CLI turned it into:

```json
{"error_code": "quota_unexpected_collection_error",
 "reason": "heartbeat receipt was not committed because quota or stall writeback did not complete: quota collection failed",
 "recommended_action": "retry quota should-run with the same --turn-instance-id after repairing heartbeat receipt writeback"}
```

That is wrong twice over: the cause is not a collection failure, and the recommended action points at receipt writeback. I hit exactly this earlier today while closing out an unsettled turn — the real message (`requested action selection qualification conflicts with its projection`) was only visible by instrumenting the CLI in-process.

## Fix

The preflight now raises a typed `QuotaActionSelectionConflictError` instead of a bare `RuntimeError`, with two kinds:

- **`unqualified`** — the projection carries no typed action-selection qualification at all;
- **`conflict`** — the requested Todo is neither the projection's current selection nor deferred/rejected by it.

The error carries its own `error_code` (`quota_action_selection_conflict`) and `recommended_action`, following the existing `QuotaIdentityPreconditionError` pattern. `quota_error_code` maps it, and `quota_failure_payload` publishes:

- `status: quota_action_selection_conflict` (instead of `quota_collection_failed`);
- `reason` = the real message, naming the requested Todo, the projection's current selection and the qualification state;
- `recommended_action` = a real next read ("rerun without `--todo-id` to read the current selection, then bind that Todo, a deferred Todo, or the Todo the recovery obligation must settle");
- a typed `action_selection_conflict` block (`kind`, `requested_todo_id`, `selected_todo_id`, `qualification_state`).

## What Is Deliberately Unchanged

- A Todo that the delivery frontier **defers or rejects** keeps its existing typed payload (`quota_action_selection_deferred` / `quota_action_selection_rejected`).
- A **qualified** selection for the requested Todo still passes the preflight and never raises (this is the case that was broken two PRs ago and is fixed in `04e035864`; the new test pins it).
- No behavior change to collection, spend, receipts or defaults — only the failure report for an already-failing path.

## Validation

[...]

PR #4546, opened 2026-09-16
feat(dashboard): let the owner confirm a validated steward team plan

huangruiteng:
## What Was Broken

The steward's team intake produces one validated multi-lane preview, and the Chat action surface already owns `team.plan` with a preview and an apply that re-validates at `PRE_SETTLEMENT`. The dashboard could not render it: `typedActionKindSchema` in `src/data/chat.ts` is a closed zod enum without `team.plan`, so a validated preview failed to parse on the client instead of becoming a card an owner could confirm. The one-sentence team flow therefore had **no confirmation surface at all** — the last missing piece of this lane's P1.

## What This Changes

- **Transport:** `typedActionKindSchema` accepts `team.plan`; `WorkspaceActionPreview["actionKind"]` matches.
- **Presentation:** new presentation-only reducer `src/features/personal-workspace/team-plan-preview.ts` renders what an owner must see before confirming:
  - the Goal the plan staffs, and its objective;
  - one row per lane: the Agent that runs it, that lane's **first bounded Todo** (`priority · action_kind · text`), and its **acceptance signal**;
  - a **staffing gap** lane reads `unstaffed · <reason_code> · <declined work>` and never shows invented work;
  - the gap summary, the **quota envelope**, and the **stop condition**.
  It owns no authority, performs no effects, and never renders a lane as already created.
- **Card wiring:** the proposal card uses that reducer, with a localized summary, an impact statement that says exactly what confirming does (ready lanes get their first bounded Todo through the canonical owner; a gap lane creates nothing; no lane exists before the apply receipt returns; a Goal/Agent change makes the proposal stale), and a confirmation label.
- **i18n:** the new strings in English and Chinese.

The generic review-plan compiler already handles a `preview_ready` non-lifecycle proposal as applyable, so no authority logic was added or changed.

## Verification

- `tsc --noEmit -p tsconfig.json` — clean.
- `npm run smoke:team-plan-proposal` (new) — `team plan proposal smoke ok`. It asserts the transport accepts the kind (the exact thing that was broken), that the kind stays exact, that a ready lane shows its first Todo, priority and acceptance, that a gap lane is reported as unstaffed with its reason and declined work, and that the quota envelope and stop condition render.
- `npm run smoke:chat-route` — `chat-route-smoke: ok`; `npm run smoke:action-review-plan` — PASS. The kind addition does not disturb the other proposal surfaces.
- `loopx canary premerge --from-git-diff` — passed (diff hygiene, 8 risk-profile smokes, catalog canaries; one advisory: the known baseline maintainability ratchet).

## Entry Points And Scope

- **Changed:** the dashboard chat/workspace proposal card.
- **Unchanged:** the Lark goal channel, the CLI, and the Chat backend action surface (they already carried `team.plan`).
- **First screen:** the card only exists when a validated team plan is present, so the dashboard's first viewport, hero, primary CTA and opening navigation are unchanged; no preview gate applies.
- Packaged frontend: the dashboard bundle is built at install time, so I will verify the shipped chat bundle contains the new kind after promoting the local install, and report it in the PR comment.

## Residual Gaps

- There is no browser-level end-to-end acceptance for confirming a team plan from the dashboard (it would need a seeded steward preview plus a running dev server); the smoke covers the transport and the presentation reducer, not the click path.
- The card renders the preview it is given. If an apply makes the plan stale mid-flight, the existing stale/`regenerate` path handles it; nothing new was added for that flow.


huangruiteng:
Closed: this branch still carried the already-merged #4545 commits, so the PR re-showed those files. Reopening from a clean branch based on the current main with only the dashboard change.

